PR #100, opened 2015-01-31
Implemented Print and Print Now

rprego:
I implemented the on_actionPrint_triggered function which was previously commented out, and added a on_actionPrint_Now_triggered function to handle the Print Now action


danieleds:
Thank you!
Printing has still some problems, for example a row could be rendered in between two pages, so I don't think I'll enable printing in the next release unless it gets fixed.
I'm going to merge your changes, anyway.

By the way, is there any reason why you put QPrinter and QPrintDialog on the heap, instead of on the stack?

rprego:
Now that you mention it, no particular reason. I suppose it would be better
off for them both to be on the stack since there no real need otherwise.
I'll make that change.

PR #108, opened 2015-02-23
add a real --help output and some library and tool checks

darealshinji:


danieleds:
Great!
I see you check for git and pkg-config... shouldn't they be added as deb package dependencies too?

(actually, git can be left off: the deb package already contains the complete source code with all the submodules checked out, so it's not necessary).

darealshinji:
I don't know if the qt5 development tools even rely on pkg-config. I've basically added this as a platform independent way to check for the required libraries. That's intended to be a help to check if all the libraries are installed. I didn't want the script to error exit if a library is missing in case someone might be using his own versions of libraries.
Adding pkg-config to the Debian build dependencies might be a good idea though.

Some notes:
I'm using qtchooser to check for qt5 qmake because I have some trouble on my system with the qmake command. And /usr/bin/qmake might not even be the default command on all distributions. However the current qtchooser command does work only if the $QMAKE variable is used without quotes. But that shouldn't be a problem on most systems.
I guess I could also let the script fail if the make command isn't in PATH. And I could add a variable, in case people want to use a different make tool or GNU make is prefixed (gmake).
I think I could also add a small check to see if the c++ compiler is working.

danieleds:
Ok! Let's keep the Debian package as it is now.
Merging

PR #114, opened 2015-03-06
don't install editor files into pointless subdirectory

darealshinji:
I'm not sure if line 161 in src/ui/ui.pro is now required anymore.


danieleds:
Not pointless anymore (see extensions branch)

PR #119, opened 2015-03-08
no "--qmake=COMMAND" option, hint in qmake's errorExit warning

darealshinji:
Removed the "--qmake=COMMAND" option because I always got the warning "shift: can't shift that many".

### DIFF
--- a/configure
+++ b/configure
@@ -29,9 +29,7 @@ Defaults for the options are specified in brackets.
 
   --fail-on-missing       return exit 1 if non-optional components are missing
 
-  --qmake COMMAND,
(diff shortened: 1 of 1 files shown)


danieleds:
If I remember correctly, --qmake=COMMAND was there because for the deb package, dh_auto_configure uses that syntax.

danieleds:
Probably, the problem is that here https://github.com/notepadqq/notepadqq/blob/90e1ccc37fa6985bcb16957731da4d3e3fc329bc/configure#L84
you shouldn't shift

darealshinji:
Yes, that's it. It works without the "shift".

PR #157, opened 2015-11-21
Add German translation

darealshinji:


darealshinji:
I've added the QTranslator code to load the translation and modified the .pro file to generate and install it.

nqq_de

danieleds:
Awesome!
We still need to figure out a way to expose translations to the codemirror part, but this is a good start :)

PR #171, opened 2016-02-08
Create notepadqq_hu.ts

SandorPetrovics:
Add Hungarian language to User Interface of Notepadqq


danieleds:
Thank you, could you add the file to `src/ui/ui.pro`?
Search for `notepadqq_de` to see where you should add what. You should find 3 results.

SandorPetrovics:
I added, but it would require a review because I'm not quite sure. that is correct

danieleds:
Thank you, merged!

PR #176, opened 2016-02-20
added Shift+Insert paste functionality

creoflux:
These changes let the user use Shift+Insert shortcut to paste.  I have been adding support every time I get a new version of notepadqq since this the shortcut I am in the habit of using.  I am not attached to my specific implementation since I am not a Qt developer and there is probably ample room for improvement.


danieleds:
Nice addition! Thank you!

PR #181, opened 2016-03-05
Add file extensions for newer Fortran standards

vpop:
I noticed a user asking for this feature on the [site](http://notepadqq.altervista.org/wp/download/#comment-27456) and hopefully there are other Fortran archaeologists out there benefiting from this.


danieleds:
Thank you!

PR #184, opened 2016-03-07
Feature/keyboard tab switch

vpop:
Added shortcuts for:

1.   Navigating to a specific tab by pressing <kbd>Alt</kbd>+<kbd>1 .. 9 digit</kbd>. This way you can quickly jump to any of the nine left-most tabs in Notepadqq. For example, <kbd>Alt</kbd>+<kbd>3</kbd> will change focus to the third tab from the left or do nothing if you do not have three documents opened yet.
2.   Easily move to previous (left) tab or next (right) tab by using the <kbd>Ctrl</kbd>+<kbd>PageUp</kbd> or <kbd>Ctrl</kbd>+<kbd>PageDown</kbd> shortcuts. This will wrap around if you are on the first tab and try to move to the left, or if on last and moving to the right.

The choice of keys used for the combinations above matches other tab-based applications such as Chrome, Firefox, Pidgin, etc. These should help power users rely less on their mouse.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -480,6 +480,9 @@ void MainWindow::keyPressEvent(QKeyEvent *ev)
         } else {
             toggleOverwrite();
         }
+    } else if (ev->key() >= Qt::Key_1 && ev->key() <= Qt::Key_9 &&
+               QApplication::keyboardModifiers().testFlag(Qt::AltModifier)) {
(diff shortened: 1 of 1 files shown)


danieleds:
Thank you! Merged as 04427002d786b2ff6dc0fb4c1eeaf1a9111ccefb

danieleds:
Feel free to add your name to https://github.com/notepadqq/notepadqq/blob/master/CONTRIBUTORS.md

PR #186, opened 2016-03-11
Make copyright year dynamic on About Notepadqq form

vpop:
Hopefully this reduces the headache of keeping it up to date, especially as the number of translations increases.


danieleds:
I like the parametric year in the translation files a lot, but I'd avoid assigning the current year to it.
Could you change it to something like this?

```c++
int toYear = 2016;
return QObject::trUtf8("Copyright © 2010-%1, Daniele Di Sarli").arg(toYear);
```

danieleds:
Thanks! Merged: d06729ec27772e7144c99d6fd635c76a89df3f2d

Issue #183, opened 2016-03-06
Shift lines up/down


superscript18:
Notepad++ has a shortcut to move a line up or down in the file.  Does this feature exist in NotepadQQ?

PR #187, opened 2016-03-13
Adding ability to Shift lines up/down ref issue #183

creoflux:
Since I didn't see a way to this after reading issue #183, this pull request adds the ability to Move a Line Up (using Ctrl+Shift+Up) or Move a Line Down (using Ctrl+Shift+Down).  The way I have implemented it only works on single lines (similar to how delete line and duplicate line behave).  From what I recall this may be slightly different from notepad++ which I think allows the shifting of a selected region of text also.

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -398,6 +398,42 @@ UiDriver.registerEventHandler("C_CMD_DUPLICATE_LINE", function(msg, data, prevRe
     editor.replaceRange('\n' + line, pos);
 });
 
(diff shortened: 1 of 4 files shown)


danieleds:
Very nice code!
Merged as 681d2c61a53660f5e65435cd69fa24561139a8d9
Feel free to add your name to https://github.com/notepadqq/notepadqq/blob/master/CONTRIBUTORS.md

creoflux:
Thanks :)

Issue #155, opened 2015-11-13
Save all tabs when closing program


grenzor:
Please add feature like ++ where you can close the program with open tabs (saved/unsaved) and when you open it back it restores them.

danieleds:
Planned

Teklad:
Implemented in PR #249

PR #188, opened 2016-03-13
Add localization switch functionality within Preferences

vpop:
Added a combo box inside the Preferences/General form in order to allow the user to switch the localization. This will translate the application the next time it runs to the language the user selected.

notepadqq_localization_setting

The choice gets saved in a QSettings property. Every time Notepadqq runs, it will perform the translation based on the language found in the property, or default to English if no property is found.

The list of translations in the combo box gets populated by reading the .qm files and extracting the language from them.

Upon the user making a language selection in the Localization combo box, a popup will notify the user of the requirement of a Notepadqq restart for the changes to take effect.

notepadqq_localization_setting_popup

As soon as #155 is implemented, we could do a restart upon accepting the preferences or add a "Restart Now" button on the popup.

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -398,6 +398,42 @@ UiDriver.registerEventHandler("C_CMD_DUPLICATE_LINE", function(msg, data, prevRe
     editor.replaceRange('\n' + line, pos);
 });
 
+UiDriver.registerEventHandler("C_CMD_MOVE_LINE_UP", function(msg, data, prevReturn) {
+    
+    var cur = editor.getCursor();
+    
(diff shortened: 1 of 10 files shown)


danieleds:
Perfect! :)
Just one thing: could you remove the "don't show again" checkbox on the popup? I believe it adds unnecessary complexity, and we'll have to remove it anyway when we're going to add the "Restart Now" button.

Thanks!

danieleds:
Thank you, merged! Here it is: 4e5bbbc25bf354e59488f80335a1d7803dc183c6

PR #207, opened 2016-04-07
Add full path, dir path and filename to clipboard options in tab context menu

vpop:
I have enabled the following existing functionality to the tab context menu:
- Current Full File Path to Clipboard
- Current Filename to Clipboard
- Current Directory Path to Clipboard

Having it available there matches other editors/IDEs such as Notepad++, IntelliJ IDEA, Visual Studio.

notepadqq_tab_context_menu_paths

Although the word "current" is kind of redundant in a tab context menu, to remove it would mean three new actions defined in the _mainwindow.ui_. Let me know if you would prefer that.

The three options are currently available under **Edit -> Copy to Clipboard**.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -60,6 +60,8 @@ MainWindow::MainWindow(const QString &workingDirectory, const QStringList &argum
     m_tabContextMenu = new QMenu(this);
     QAction *separator = new QAction(this);
     separator->setSeparator(true);
+    QAction *separatorBottom = new QAction(this);
+    separatorBottom->setSeparator(true);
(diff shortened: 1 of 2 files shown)


danieleds:
Nice!
About the redundant naming: now that you make me notice it, it's actually redundant in the Edit menu too... it's pretty obvious that the path is relative to the current file.
Maybe the labels of the actions should be totally revisited, for example in the following way:

 * Copy Full Path to Clipboard
 * Copy Filename to Clipboard
 * Copy Directory to Clipboard

what do you think?

vpop:
Yes, those sound good. I will update the existing actions and use them for the tab context menu too.

danieleds:
Thank you!

Issue #206, opened 2016-04-06
Indentation messages not localized


darealshinji:
I can't localize the indentation warning message.
Adding `tr(` to line 18 and 31 in `src/ui/EditorNS/bannerindentationdetected.cpp` doesn't work. And the parts like `addButton(tr("Use tabs"))` that are already localized are still always displayed in English.

darealshinji:
Duplicate of https://github.com/notepadqq/notepadqq/issues/173

PR #210, opened 2016-04-08
Update bannerindentationdetected.cpp

SandorPetrovics:
Make two strings translatable


danieleds:
Merging for consistency, but take a look at #206 ... we still have to find out where the problem is
Thank you

Issue #191, opened 2016-03-15
Regex capturing groups and backreferences don't seem to work


Swyter:
I tend to use a lot of regular expressions for quickly editing text so I noticed that opening the _Search and replace_ dialog with Ctrl + H, switching to _Search with regular expressions_ and using capture groups with references doesn't seem to work at all.

In Notepad you can use something like `(a-z)+` as search string and then use Perl-style back-references (`\1` in older versions and `$1` since v6.0) in the replacement input box to get the capture result.

In Notepadqq the regular expressions seem to work fine, but they get replaced by literals (like `$1`), so they are useless.

---
- http://www.rexegg.com/regex-capture.html
- https://notepad-plus-plus.org/news/notepad-6.0-release.html
- http://stackoverflow.com/questions/11970405/
- http://stackoverflow.com/questions/2890700/
- http://stackoverflow.com/questions/17998179/

Algomorph:
Bump

danieleds:
If someone is willing to work on this, I can provide help as needed.

Thanks for the report.

Algomorph:
@danieleds Can you give us a brief on
1) What source files would we need to look at
2) In general terms, what would need to be changed
I might do it if I find that it takes a reasonable amount of time, I need your input to estimate. Thanks in advance.

danieleds:
The replace dialog is located at `src/ui/Search/frmsearchreplace.cpp`. The "replace" command constructs an object containing information relative to the action and sends it to the Editor (see method `frmSearchReplace::replace()`). In this case, the message is `C_FUN_REPLACE`.

By opening `src/editor/app.js` and searching for `C_FUN_REPLACE` you will find the handler for the replace command. You'll see that the documentation states:

> Replace the currently selected text, then search with a specified regex (calls C_FUN_SEARCH)

so the best thing to do would be to take a look at the search handler and:

  1. see how to obtain captured groups from CodeMirror. [[manual]](https://codemirror.net/doc/manual.html)
  2. cache them so that they can be used in the following "replace" command, if any.

It should be fairly simple, the only obstacle might be CodeMirror. Worst case scenario: a patch to `search/searchcursor.js` in CodeMirror is needed.

Let me know for any kind of doubt :-)

Algomorph:
It took me three whole hours... My fault though, mostly, for not having an appropriate JavaScript IDE and being rusty on JavaScript...

snarum-NIS:
[...]

PR #224, opened 2016-07-10
Added regex group reuse token functionality to Replace & Replace All

Algomorph:
Replace & Replace All now mimics Notepad++ replace functionality
for regular expressions with groups: \[1-9] get replaced by groups
with the corresponding group in the regular expression.

**Example**

Test file contents:
hello C++ goodbye
hello JavaScript goodbye
hello CodeMirror goodbye

Regular expression in "Find" input box of the Search dialog:
```(hell)o\s*([^ ]*)\s*goodbye```

In "Replace with":
\2 is \1

Output:
C++ is hell
JavaScript is hell
CodeMirror is hell

Implements #191

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -251,6 +251,48 @@ UiDriver.registerEventHandler("C_FUN_SEARCH", function(msg, data, prevReturn) {
     return Search(data[0], data[1], data[2]);
 });
 
+/*
+   Determine whether the proposed replacement contains
+   group reuse tokens i.e. \1, \2, etc.
+   (Helper function for search/replace & replace all.)
+ */
+function HasGroupReuseTokens(replacement){
+	var groupReuseRegex = /\\([1-9])/g;
+	return (groupReuseRegex.exec(replacement) !== null);
+}
+/*
+   Substitute group reuse tokens (i.e. \1, \2, etc.) with 
+   the matched groups provided.
+   (Helper function for search/replace & replace all.)
+   groups: contains array of regexp matches, where the first 
+   entry is the whole match and the rest are groups.
+   
+*/
+function ApplyReusedGroups(replacement, groups){
+	//If we got match subgroups, see if we need to alter the replacement
+	for (var iReuseGroup = 1; iReuseGroup < groups.length; iReuseGroup ++){
+		//takes care of non-consecutive group reuse tokens,
+		//i.e. in "\1 \3" with no "\2", the "\3" is ignored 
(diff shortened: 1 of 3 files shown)


danieleds:
Thank you!
Two things:

**1)**
I think the example regex should be `(hell)o\s(.*)\sgoodbye`.

**2)**
I noticed a strange behavior while searching in *plain text*. Try the following steps in your original test file:
 1. Search string: `hello`
 2. Replace string: `\1 \2 \3`
 3. Click the "plain text" option
 4. Replace all

Since it is plaintext you would expect this:

    \1 \2 \3 C++ goodbye
    \1 \2 \3 JavaScript goodbye
    \1 \2 \3 CodeMirror goodbye

but you get this:

    undefined \2 \3 C++ goodbye
    undefined \2 \3 JavaScript goodbye
    undefined \2 \3 CodeMirror goodbye

Could you push a new commit to handle this case? Also, feel free to add your name to the contributors file.

For everything else it works great, thanks for the nice clean code!

Algomorph:
@danieleds 
1) The regex I put in the pull request comment originally suffered from the markdown formatting. I've surrounded it with back-ticks. It should work as it is now. I changed the regex in the git commit message to _almost_ the same thing, with ```[^\s]``` instead of ```[^ ]```, which is the most correct. ```(.*)``` adds an extra space behind "C++" and such. For the pull request comment, I couldn't get that to work due to markdown issue...
2) This is fixed now. I also fixed a couple other things. We shouldn't get "undefined" for any case anymore. Also, if you do the "hello" search you described on the same test file, but in regex mode, the behavior of Notepad++ is different from the plain text behavior: you simply get the two spaces between the group reuse tokens ("   " instead of "\1 \2 \3"), so I made sure to handle that behavior as well.

danieleds:
Thank you! Works great now :)

adeel-raza-azeemi:
Notepadqq regular expression search and replace does not work properly. i.e. we can't reuse one subgroup in replace again.

Algomorph:
@adeel-raza-azeemi , can you file a new bug, please?

adeel-raza-azeemi:
oK i will. But just to inform you. You can test it too. replace text in a string. try to use use \1 or any other more than one time and it will only be replaced once. 

Btw: I didn't know how to file a new bug in github but i will figure it out :)

Algomorph:
@adeel-raza-azeemi ok, thank you sir!

adeel-raza-azeemi:
Filed the bug it #12992

PR #225, opened 2016-07-20
Feature: User customizable keyboard shortcuts.

Teklad:
I came up with a scheme which should allow more or less any menu item to be keybound.  From testing everything appears to be working.

Some tweaking and polish here and there... and I'm sure there's spots that could be more efficient, but this is definitely a good start I think.

The interface is pretty basic, but it definitely gets the job done.


danieleds:
Looks very nice to me!

A few points:

 1. F-keys without modifiers are not grabbed (eg. F1, F2, F3, ...)
 2. About your `TODO: Anyone know how to do this without making a laundry list?`:<br>
Aren't you able to enumerate the actions like you do within MainWindow::updateShortcuts()? This will also allow to automatically have translated strings in the settings dialog. It might require MainWindow to expose a method just for this.

Teklad:
Yea, I had been working on this for quite a while whenever I added that laundry list, lol.  I just pushed another commit that both fixes the F-keys and adds a function to help enumerate through available actions. 

I think that should do it for now.  Slight polishing might be needed on the interface side but aside from that everything appears sound.

danieleds:
Perfect! :) Two more things and I think we can merge:

 * Could you save the settings within a `[Shortcuts]` subsection instead of the `[General]` section? See for example how it is done for the languages settings.
 * A friendly way to restore the default shortcuts would be useful. Maybe with a "Reset All" button, and/or a <kbd>backspace</kbd> key press on a specific line?

Teklad:
Just a quick question.  Do we store default settings somewhere(i.e. a global config file) or will these need to be hard-coded for now?

danieleds:
We don't store them anywhere, you did it right.

It seems everything's working smoothly :) I'll take a better look in a few hours when I have more time and then I'll merge. Thanks!

Issue #200, opened 2016-04-03
Code folding not activating from start


vpop:
Steps to reproduce:
- create new document
- paste C++ code for example
- change language to C++
- code highlighting activates, code folding does not
- if you type a new character or Enter for example, code folding finally pops up

![notepadqq_no_code_folding](https://cloud.githubusercontent.com/assets/3687404/14230312/3d163680-f906-11e5-9a65-be9f91dbc361.png)

Same issue if you save a new file with an extension that represents a language. Hightlighting will turn on after the save, code folding will not unless you input something into the editor.

I have noticed this behaviour after building from source.

Issue #214, opened 2016-04-11
Key Ctrl+Space 


sevenzerolee:
The key Ctrl+Space is in conflict with InputMethod.

danieleds:
Hi, thanks for the report!
Can you add more information and elaborate what you mean?

Daniel

sevenzerolee:
I'm in China and use 'Ctrl+Space' to open Chinese Input Method to input. Is there any way to set the key 'Ctrl+Space' for ‘Alt+/’ or others. Thank you.

Issue #202, opened 2016-04-06
App crashes when I accept "Use spaces" on open, if doc is in other view


petsam:
The situation is :

1. I open two files (xml). 
2. I get a message on the header for the document is using spaces to separate tabs with choices "Ignore" and "Use spaces". 
3. I choose "Use spaces" and the message goes away.
4. I then move one document to "Other view". 
5. The same message re-appears. 
6. If I choose "Ignore" it's OK.
7. If I choose "Use spaces" the program crashes.

I have a general setting to "Use tabs" for my own created docs, so, opening already formatted documents not having tabs for separation is temporary for me.

danieleds:
Is someone able to reproduce this on the master branch?

Teklad:
Not able to reproduce this on my end.  Its highly possible that this was either an upstream type issue or we accidentally fixed it at some point. :|

danieleds:
Hmm okay, I'm closing the issue then.

PR #230, opened 2016-08-05
Fix foldGutter not updating properly on when switching languages.

Teklad:
This should resolve issue #200


src/ui/EditorNS/editor.cpp
```diff
@@ -231,7 +232,8 @@ namespace EditorNS
     void Editor::setLanguage(const QString &language)
     {
         sendMessage("C_CMD_SET_LANGUAGE", language);
-        if (!m_customIndentationMode)
+        sendMessage("C_CMD_REFRESH");
+	if (!m_customIndentationMode)
```
danieleds: Can you fix indentation?

src/ui/mainwindow.cpp
```diff
@@ -428,17 +428,17 @@ void MainWindow::setupLanguagesMenu()
...
-    QList<QMenu*> lst;
-    lst = ui->menuBar->findChildren<QMenu*>();
-
-    foreach (QMenu* m, lst)
-    {
-        addAction(m->menuAction());
-        addActions(m->actions());
-    }
-}
+// void MainWindow::fixKeyboardShortcuts()
```
danieleds: Why has this been commented out?

danieleds:
Thanks! I have added a few notes, can you take a look?

src/ui/EditorNS/editor.cpp
Teklad: fixKeyboardShortcuts is redundant due to the new custom keyboard shortcuts.
I'll see about fixing indentations after I get off work later. If there's
anything else you can think of let me know.

danieleds: Perfect. While you're at it, can you just delete it?

danieleds: Also, remove `#include <QDebug>` if it's not needed.

danieleds:
I've looked a bit more into it and I think the cleanest way is to just include your three-lines "refresh" code within `Languages.setLanguage`, without the need to define an additional `C_CMD_REFRESH`.

Maybe after this line: https://github.com/notepadqq/notepadqq/blob/master/src/editor/classes/Languages.js#L933

Teklad:
We could go that route too as long as you don't think it'll have any issues
once we get sessions implemented.

Teklad:
Totally forgot to ask, but what's wrong with indentation exactly?  I looked around the issues and all I could find that was relevant was #202 .  Is this what you're talking about? I'm about to push an update for the foldgutter thing that also removes fixKeyboardShortcuts entirely.

Just wanting to clarify before I go poking at things unnecessarily.

Teklad:
Going to go a little off topic for a second here.

I'm adding support for changing the keybinding for the "autocomplete" CodeMirror feature, which should take care of issue #214 .  Where would you recommend I add the code inside the editor to make sure the custom keybinding sticks across editor windows?

This appears to work as far as setting the key goes, of course any suggestions are appreciated:
`UiDriver.registerEventHandler("C_FUN_SET_AUTOCOMPLETE_KEY", function(msg, data, prevReturn) {
    var keySequence = new Object;
    keySequence[data] = "autocomplete";
    editor.setOption("extraKeys", keySequence);
});
`

danieleds:
Indentation is wrong here: https://github.com/notepadqq/notepadqq/pull/230/files#diff-d06849a0ad516e0568cef855355d8255R234
Just ignore the changes in `editor.cpp` and you're good.

Also, the "fix" code should be in `Languages.js`, not in `app.js`. Precisely here: https://github.com/notepadqq/notepadqq/blob/master/src/editor/classes/Languages.js#L933
This is because C_CMD_SET_LANGUAGE is not guaranteed to be the only command that changes the language.

I'm not sure abuout your C_CMD_SET_AUTOCOMPLETE_KEY, I think we need a more general solution.

Teklad:
That look better?  Sorry github was being a butthole and I didn't realize it was sending me to the wrong line.  It makes MUCH more sense where you wanted it when I looked again. Haha.

Also, if you can think of a better way to implement the AUTOCOMPLETE keybinding I'd love to hear your ideas.  I'm drawing blanks on how to cleanly implement it.

Teklad:
Also, real fast what's the preferred function for translated strings?  I just added conflict detection to the keybindings preferences menu locally and I want to make sure the "Conflict" string can be translated properly.

I suppose I could just change the color of the input box but I think what I have looks cleaner.

Teklad:
Ignore that last comment, makes more sense to just tint it red.

danieleds:
Merged! I've kept `fixKeyboardShortcuts()` because it looks like it's still needed.

A suggestion for your next pull requests: do your work in a different branch and do a pull request for that branch. This way, by not touching the master branch, it stays clean even in case of rebases, and the next pull requests won't have a storm of unrelated commits :)

PR #235, opened 2016-08-12
Ui improvements for keyboard shortcuts

Teklad:
Added conflict detection support so the user can be warned of duplicate keybindings.


danieleds:
I see you have correctly created a feature branch, but unfortunately your `master` branch is still dirty so I will have a hard time merging it upstream.

What I suggest is to start from scratch: redo the fork and clone the repository again, so that it is in a clean state. Then follow the workflow described here: https://gist.github.com/Chaser324/ce0505fbed06b947d962

Took me a while to learn it too so don't worry :)

Teklad:
Yes unfortunately git is taking me a while to pick up.  It has a lot of
features. Lol.

Issue #220, opened 2016-05-02
Previously unsaved files are not recovering on restart


lokesh1729:
Hi guys,

I have observed in notepad++ that if there is an unsaved file, if i attempt restart and open notepad++ previous files are same as it is. But in notepadqq it's not like that. I am not able to get those.

guekho64:
That would be a great idea for some people who forget to save their files.

+1

PD: What about "Autosave"? Could it be possible someday, alongside recoverable "unsaved files"?

danieleds:
I agree that it would be great to have both (unsaved file recovery and autosave).
The first one is a special case of Sessions.

They surely need to be implemented.

Teklad:
Implemented completely in PR #249

cpagravel:
Is this feature working? I just lost some important info because it crashed and looks to not have recovered the files.

JuBan1:
No, recovery from crashes isn't working yet. 

This requires autosaving to work. I've written the code for it but the underlying editor doesn't expose a feature I need to make it work nicely so I shelved it for the time being.

Sorry that you've lost data. Keeping users' files save should be all the way up the priority list. Hopefully it won't be much longer until this works.

Sariyn:
Hello, is this still a bug in the latest version?  Just lost some important work.  Nice heads up btw.

JuBan1:
What exactly happened? Nqq doesn't do anything to protect your unsaved documents in case of crashes. 

If you restart your computer Nqq should be closed properly and save your documents.

Sariyn:
Hi, yes it is how you said-- system crashed that resulted in loss of unsaved notepadqq work.
To be fair however, it is probably easy to be mistaken for relying on this feature as it not atypical in other editors, and even the Windows notepad++ counterpart supports some level of disaster recovery on crash with unsaved work.  It feels like too an important feature to be missing.  Other than that, great app.

JuBan1:
@mrandreastoth I'm afraid you've got the wrong project :) This is Notepad**qq**, a Linux and macOS alternative to Notepad++. If you want to report an issue with N++ you can probably do that [there](https://github.com/notepad-plus-plus/notepad-plus-plus/issues)

mrandreastoth:
Lol! How embarrassing. But thank you for the link. :)

TimHandy:
This just happened to me. PopOS didn't suspend properly and crashed. On restarting I lost all my text in an unsaved file in Notepadqq. :-(

[...]

PR #239, opened 2016-08-23
Patches for remembering unsaved changes to tabs (Sessions)

JuBan1:
Hello folks,

I discovered Notepadqq a little ago and really missed notepad++'s feature of sessions/remembering tabs. I implemented it myself and am now using it for a short while. Since the feature was requested here https://github.com/notepadqq/notepadqq/issues/220 and here https://github.com/notepadqq/notepadqq/issues/155 I thought I might as well submit them. I hope you'll find some of it useful.

What is added:
* Notepadqq remembers unsaved changes to tabs and caches them at exit instead of asking the user to save or discard the changes. This works for opened files or "new" tabs with no file associated.
* Added an checkbox to enable or disable this feature to the General section of the preferences menu.

I tried to closely emulate Notepad++'s ways. As such, only one "session" exists. If the user has more than one Notepadqq window with unsaved changes, only one will have its tabs remembered. The user will be prompted to save/discard changes when trying to close one of the other windows. No functionality to load/save different sessions.

I have almost no experience with either github or collaborative projects, so any comments are appreciated. In fact, I hope nothing will crash or burn when I click the submit button in a second.

Cheers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ build-notepadqq-Desktop*
 /out
 /src/ui/ui.pro.user
 /src/ui/ui.pro.user.*
+/notepadqq.pro.user
--- a/src/ui/frmpreferences.cpp
+++ b/src/ui/frmpreferences.cpp
@@ -79,13 +79,14 @@ void frmPreferences::loadShortcuts(QSettings* s)
 {
     MainWindow* mw = qobject_cast<MainWindow*>(parent());
     m_shortcuts = new QMap<QString,QString>;
+
     foreach(QAction* a, mw->getActions())
     {
-        if(a->objectName().isEmpty())continue;
-        m_shortcuts->insert(a->objectName(),a->iconText());
+        if(a->objectName().isEmpty()) continue;
+        m_shortcuts->insert(a->objectName(), a->iconText());
     }
 
-    kg = new keyGrabber();
+    kg = new KeyGrabber();
 
     //Build the interface
     QWidget *container = new QWidget();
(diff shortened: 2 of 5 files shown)


Teklad:
This was a feature pretty high up on the wishlist.  If you haven't done so already go ahead and remove #include <QDebug> if its still there(I'm currently busy so didn't get to check the code thoroughly).

I'm sure @danieleds will be interested in looking at this.

JuBan1:
@Teklad The `#include <QDebug>` is definitely still somewhere, but it's also still in use. I know there's at least one call to `qDebug` that isn't coming from my code. I haven't added or removed any includes at all.

Teklad:
Oh well.  I'm sure I'll come across it during my next set of patches.  
Your code looks pretty good from what I can tell though. Look forward to 
seeing it merged. :)

danieleds:
Very nice! I'll test it as soon as I have some time.

I see this as a good time to generalize it to load/save multiple sessions. What's the problem (if any) that you found with saving the tabs for the other windows too?

JuBan1:
@danieleds Technically I think it's not that big of a deal although the extra bookkeeping means we might know want to store all of that inside the QSettings object. It's already a bit inconvenient with two Lists of QVariants.

I modelled the feature after how it works in Notepad++ by default. N++ restricts you to one session window which means the rules for window creation are quite clear: if the session window isn't open, the next window opened will be the session window (and thus have its tabs restored from cache). If the session window is open, the next window will be a normal one.

For multiple sessions things become a little muddy. Will all of them be restored automatically? In what order? How do you close a session? Etc.

I haven't had much use for multiple sessions personally, so I stuck with what I know for the time being.

PR #241, opened 2016-08-23
Fixes #199 and cleans up some code.

### LINKED ISSUE #199: Zenburn theme doesn't properly style new tabs
Changing the theme to zenburn will correctly update all existing tabs, but new ones aren't styled correctly. They have a white background, with black text and the current line is highlighted in black, exactly the same color as the text, making it essentially unusable.

If you go back into the settings, the theme switcher still says zenburn, but if you just hit okay (without changing anything) it will fix the problem, at least until more tabs are opened.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32415258-zenburn-theme-doesn-t-properly-style-new-tabs?utm_campaign=plugin&utm_content=tracker%2F346321&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F346321&utm_medium=issues&utm_source=github).
</bountysource-plugin>

JuBan1:
Some changes to the handling of themes.

* The theme Zenburn doesn't work properly (also referenced in #199). I don't exactly know why this one in particular fails, but I found a workaround. Instead of initially passing the theme to the WebView, we can use `setTheme()` once the view and layout are sufficiently constructed. As far as I can see, all themes works well this way.

* On that occasion I noticed that the preferences menu was overly eager with setting `m_previewEditor`'s theme. It'll now only do so twice (still technically one too many times, but removing the last redundant call is more hassle than it's worth).

* Also, `on_buttonBox_accepted()` is optimized a bit. As far as I can see, calling `invalidateEditorBuffer` and `addEditorToBuffer(1)` just once at the end of the loop works just as well -but faster- than before.

I'm not 100% confident in those chances since I don't quite grasp the bigger picture of the all the EditorNS code works. I haven't found any bugs while testing and everything seems sound, but some extra caution may be advisable.

~ Zenburn looks nice ~


src/ui/frmpreferences.cpp
```diff
@@ -150,30 +150,28 @@ void frmPreferences::on_buttonBox_accepted()
     s.setValue("Extensions/Runtime_Nodejs", ui->txtNodejs->text());
     s.setValue("Extensions/Runtime_Npm", ui->txtNpm->text());
 
+    const Editor::Theme& newTheme = Editor::themeFromName( ui->cmbColorScheme->currentText() );
```
danieleds: You should use `ui->cmbColorScheme->currentData()` here instead of `currentText()`, like it was before. Same thing for some of the other changes in this file.
This is because the "display text" of a control is rarely reliable data and might change based e.g. on user locale.

danieleds:
Hi! I have added a comment to your changes.

In addition to that, I'm not sure if the changes in `editor.cpp` are safe. I mean, I tried them and it works great, but it might create unexpected behaviors later when we will hopefully switch to the newer API.

I'll merge the other changes as soon as you tweak them, and I will explore this problem after that.

JuBan1:
Alright, I've reverted the changes in `Editor.cpp` and instead added a workaround for this specific theme. It's crude, but the only alternative I can come up with other than leaving the theme broken until a proper fix arrives.

Code in Preferences also uses `currentData()` again.

Teklad:
Just FYI.  I've actually noticed a lot of broken themes upon initially loading Notepadqq.  If I manually reset the theme it fixes the issue.  This is most likely related.

JuBan1:
I haven't noticed any themes other than Zenburn being broken, then again I
mostly use just two or three. Can you give me an example?

Teklad:
For instance, when I apply the "Dracula" theme and then restart notepadqq.  Upon restarting the line numbers on the left are the proper color, but the code display section is still white background/black foreground, which is the default.

This is most likely due to the setTheme being called before the codemirror instance is fully loaded... resulting in odd behavior.  Its more noticeable in dark background themes I think, and some themes are more prone to the issue than others for some odd reason.

A possible solution to the issue would be to call the internal theme function within the javascript code once certain conditions are met.  I've been poking around at CodeMirror for a while now and it tends to have odd behaviors when you do things at the wrong time.

This is only an assumption, however. I've not actually tested the theory thoroughly since I typically use the default theme during testing.

danieleds:
@Teklad Right now, the theme to use is specified as part of the configuration in CodeMirror's constructor. So no issue here, because no setTheme is called manually.

You're right though to say that this is a timing issue. My hypothesis is that something is causing CodeMirror to load *before* the theme files are fully loaded. And that's odd. This is why I would like to fix it at the CodeMirror level, not at the C++ level @JuBan1

Teklad:
One idea would be to create an addon for CodeMirror which would work as 
a wrapper to ensure the theme requested is properly loaded up prior to 
setting it, CodeMirror was originally meant to have all the themes 
loaded in css prior.  That would most likely solve any timing issues, 
since addons won't run until CodeMirror is fully loaded.

As it stands, we dynamically add the stylesheets within SET_THEME, 
which, in all honestly, should work fine... and indeed does work once 
notepadqq has an editor fully intiailized.

It simply acts like the theme file wasn't appended properly on initial load.

JuBan1:
Well, that sucks. I tried to fiddle with the CodeMirror js files to understand why the load order sometimes screws up, but to no avail. Even manually calling setTheme() during the construction of the editor (like I initially suggested) doesn't work all the time. It just covered up the issue most of the time.

IMO fixing this inconsistent and buggy behavior should probably be pretty far up the to-do list.

Teklad:
I was actually planning to look into it before reworking the search
feature. Just finishing up my show-symbols branch.

I have a good idea on how to fix it properly.

danieleds:
@Teklad I was taking a look and I think I fixed it. I had to move the stylesheet loader *after* the js had loaded.

Pushing in a moment.

danieleds:
@JuBan1 I have merged your pull request here: 97b93b319e8b43310c7513b76100f84b7ef93373. Thank you.

I'm keeping this open to get confirmation from you and @Teklad about my fix for Zenburn. If it works for you too, then we can close this PR.

JuBan1:
Zenburn now works on my end. The bugs aren't entirely gone. Some bugs still show up. These bugs seem to be limited to first-time loading in the settings preview window though. They work well afterwards and I haven't had the bugs in the actual editor.

danieleds:
Okay, so that happens only when the left column with line numbers has a bigger width than normal, if I get it right.

Teklad:
Also confirmed on this side.  The editor appears to function properly 
upon initial start now.  Good work!

danieleds:
Good. I'll open an issue for the problem with the preview editor.

PR #242, opened 2016-08-23
[Resubmit] Patches to enable sessions

JuBan1:
I'm fighting with git but I think I finally got the hang of it. I've reopend this from https://github.com/notepadqq/notepadqq/pull/239 based on a different branch so I can have a clean master. Code's the same.

------

Hello folks,

I discovered Notepadqq a little ago and really missed notepad++'s feature of sessions/remembering tabs. I implemented it myself and am now using it for a short while. Since the feature was requested here https://github.com/notepadqq/notepadqq/issues/220 and here https://github.com/notepadqq/notepadqq/issues/155 I thought I might as well submit them. I hope you'll find some of it useful.

What is added:
* Notepadqq remembers unsaved changes to tabs and caches them at exit instead of asking the user to save or discard the changes. This works for opened files or "new" tabs with no file associated.
* Added an checkbox to enable or disable this feature to the General section of the preferences menu.

I tried to closely emulate Notepad++'s ways. As such, only one "session" exists. If the user has more than one Notepadqq window with unsaved changes, only one will have its tabs remembered. The user will be prompted to save/discard changes when trying to close one of the other windows. No functionality to load/save different sessions.

I have almost no experience with either github or collaborative projects, so any comments are appreciated. In fact, I hope nothing will crash or burn when I click the submit button in a second.

Cheers.


danieleds:
Perfect!

You're doing good, just for reference you might find this useful: https://gist.github.com/Chaser324/ce0505fbed06b947d962

PR #243, opened 2016-08-23
Fixes #240; Improved view pane handling.

### LINKED ISSUE #240: Program crash when trying to close a tab.
Steps to reproduce:

* Open a fresh window.
* Open a second tab.
* Right click on second tab and select "Move to Other View"
* Middle click(!) on the tab you just moved to close it.
* Segmentation Fault

Using the 'x' button to close the tab works normally. Doesn't occur every time, but most of the time. Just repeat steps 3+4 if it doesn't work immediately.

PS: I have a fix for that lined up, but I'm fighting more with git than with the code. It'll come eventually.
PPS: Fix is here.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37199488-program-crash-when-trying-to-close-a-tab?utm_campaign=plugin&utm_content=tracker%2F346321&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F346321&utm_medium=issues&utm_source=github).
</bountysource-plugin>

JuBan1:
Fixes the segfault described in #240. 

Also, when a new view is created through "Move to other View" both views will now share their space equally. Before that, the new view would be squished and text was basically unreadable until the view was resized manually.

### DIFF
--- a/src/ui/editortabwidget.cpp
+++ b/src/ui/editortabwidget.cpp
@@ -258,23 +258,15 @@ void EditorTabWidget::on_editorMouseWheel(QWheelEvent *ev)
 
 void EditorTabWidget::mouseReleaseEvent(QMouseEvent *ev)
 {
-    int openTabs = count();
-
     if (ev->button() == Qt::MiddleButton) {
         int index = tabBar()->tabAt(ev->pos());
 
         if (index != -1) {
             emit tabCloseRequested(index);
-
-            --openTabs;
-
-            //Closing the tab may result in the TabWidget being deleted if it was the last tab.
-            //In this case we must not use mouseReleaseEvent or other members anymore!
-            if(openTabs <= 0)
-                return;
+            ev->accept();
+            return;
         }
     }
-
 
     QTabWidget::mouseReleaseEvent(ev);
 }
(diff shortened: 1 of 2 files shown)


src/ui/editortabwidget.cpp
```diff
@@ -258,13 +258,24 @@ void EditorTabWidget::on_editorMouseWheel(QWheelEvent *ev)
...
+
+            //Closing the tab may result in the TabWidget being deleted if it was the last tab.
+            //In this case we must not use mouseReleaseEvent or other members anymore!
+            if(openTabs <= 0)
+                return;
         }
     }
 
+
     QTabWidget::mouseReleaseEvent(ev);
```
danieleds: You should be able to just move this line to the very top of the method and it should work without any other change :)

JuBan1:
Your suggested didn't work, surprisingly. I assume because `QTabWidget::mouseReleaseEvent()` calls `event->ignore()` which probably makes Qt look for children of the tabwidget to handle the event after `mouseReleaseEvent()` returns. But at that point the widget will already be deleted if it had only one tab.

But if I recall correctly passing the event along is something we shouldn't do anyways if we handle it. So the last commit got rid of all the extra code in favor of simply accepting the event if we handle it. I think this is probably the best solution.

danieleds:
Perfect! Merged.

PR #244, opened 2016-08-24
Show symbols

Teklad:
This is the notepadqq portion of the show-symbols feature.  I'm still debating on what to do about the CodeMirror addon.

I've more or less scrapped all the original code I was going to use, so the entire way the addon functions is different from the one I was initially going to use from CodeRaiser.

Pretty simple changes mostly, basically all this does is exposes the functionality of the addon to the notepadqq instance.

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -71,6 +71,11 @@ UiDriver.registerEventHandler("C_CMD_SHOW_END_OF_LINE", function(msg, data, prev
     editor.refresh();
 });
 
+UiDriver.registerEventHandler("C_CMD_SHOW_WHITESPACE", function(msg, data, prevReturn) {
+    editor.setOption("showWhitespace",data == true);
+    editor.refresh();
+});
+
 UiDriver.registerEventHandler("C_FUN_GET_INDENTATION_MODE", function(msg, data, prevReturn) {
     return { useTabs: editor.options.indentWithTabs, size: editor.options.indentUnit };
 });
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -66,6 +66,11 @@ UiDriver.registerEventHandler("C_CMD_SET_INDENTATION_MODE", function(msg, data, 
     editor.refresh();
 });
 
+UiDriver.registerEventHandler("C_CMD_SHOW_END_OF_LINE", function(msg, data, prevReturn) {
+    editor.setOption("showEOL",data == true);
+    editor.refresh();
+});
+
 UiDriver.registerEventHandler("C_FUN_GET_INDENTATION_MODE", function(msg, data, prevReturn) {
     return { useTabs: editor.options.indentWithTabs, size: editor.options.indentUnit };
 });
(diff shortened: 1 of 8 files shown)


danieleds:
Ok no problem with the addon, if the code is totally different.

The glitch I was talking about is this: when I click on "Show All Characters", I expect the "Show Tabs" and "Show End of Line" items in the "View -> Show Symbol" menu to get checked automatically.

So "Show All Characters" is checked **iff** "Show Tabs" and "Show End of Line" are _both_ checked.

In addition, we're going to need an additional menu item: "Show Spaces" :)

Teklad:
I've made some changes which incorporate all but the "Show Spaces" menu option.  

Also changed things up so notepadqq doesn't bother grabbing settings on startup for the others in the case that showSymbols is enabled.  This was simply a side effect of adding the feature, so it all worked out for the better.

Show All Characters properly restores the previous settings as well when it is toggled off.

I'll add the final "Show Spaces" action tomorrow.  If there is anything else you can think of give me a heads up and I'll try to get it worked in.

Teklad:
Okay.... I think I'm all done... just double check to see if the behavior is how you expect it to be and I believe that should cover everything.

danieleds:
This is the other glitch I was talking about:

 1. Enable "Show all characters"
 2. Hide tabs from "View -> Show Symbols -> Show Tabs"
 3. Tabs *and* spaces disappear, instead of only tabs.

And this is another:

 1. In "View -> Show Symbols", enable tabs, spaces and EOL
 2. Enable "Show all charcters"
 3. Disable "Show all characters"
 4. No special characters are shown (and it is correct), but in "View -> Show Symbols" they're all checked so to show only spaces I have to deselect it and select it again which is odd.

So maybe this is a better behavior for this second glitch:

 * If the user enables all the three options (tabs, spaces and EOL), then "Show all symbols" should be automatically enabled
 * If the user disables "Show all symbols" and the previous settings said that tabs, spaces and EOL were all enabled, disable them all.

Teklad:
Okay. I'll see what I can do. Currently it just restores previous settings.
I'll add some checks to even out the behavior. :)

Teklad:
Alright... made a few changes.... see how you like it... I'll need to update the CodeMirror repository in order to add the ability to color the tab character.  Figured I'd slip the update in here since its a very small change.

Anyways, I've made some changes to the "Show All Characters" behavior.... I think it matches your requests.

Let me know what you think.

danieleds:
The tab arrows look awesome :D
And the glitches are fixed.

danieleds:
I have merged the code, but can you provide some documentation for `bool MainWindow::updateSymbols(bool on)`?

Like parameters and output meanings.

You can just post it here and I'll take care to commit it.

Teklad:
/*
  * @brief Update symbol options using parameter "on" and 
Show_All_Characters toggle status
  * @param on true or false based on the calling element's toggle status.
  * @return bool: true if on is false and Show_All_Characters is 
checked.  False otherwise.
  *               on a true return, default symbol saving behavior is 
modified.
  */

Teklad:
Sorry about the code formatting.  I tend to fall back to my own formatting from time to time when I'm writing these things.  I'll try to keep the nuances of the style in mind for the next pull request. :P

danieleds:
Thanks!

About the code formatting, we actually need some rules (right now, there aren't). I don't want to be strict about this, because I believe that a too rigid set of rules will do much more harm than good, but maybe a set of basic guidelines would be good. I'll think about it.

Issue #152, opened 2015-10-12
Change default editor font


Swyter:
This is basic functionality. I still haven't found an easy way to change it.

It's a bit funny that a complex, fully-fledged CSS-based code highlight framework like _Codemirror_ does not expose something that even _Notepad_ has.

I think it should be high on the to-do list. Anyway, keep it up.

xnko:
While there is no settings for that, you can change font-family in the file `/usr/share/notepadqq/editor/libs/codemirror/lib/codemirro.css`

akien-mga:
I'd be interested in this feature too, the default font is not monospace so it's the first thing I wanted to change when starting notepadqq for the first time.

danieleds:
@akien-mga It should be. What distro are you using?

akien-mga:
Mageia 6; I just packaged notepadqq for its official repositories: http://svnweb.mageia.org/packages/cauldron/notepadqq/current/SPECS/notepadqq.spec?view=markup&pathrev=987360

Building against CodeMirror 5.3.0. Here's how the font looks like (with custom colours that I chose while testing, but it does not impact the font):

![spectacle m24589](https://cloud.githubusercontent.com/assets/4701338/13616217/38381228-e579-11e5-80b3-369d8b92e5cf.png)

akien-mga:
I'll test a clean build from source without the packaging step, it might be related to something in my packaging I guess.

akien-mga:
Same thing with a build from HEAD using updated git submodules.

danieleds:
Are you able to see monospace fonts in e.g. Google Chrome / Chromium?

OhDaeto:
Updated Arch to qt 5.6 and now I do not know how to return the previous font.

Teklad:
This issue was resolved in: #245 .... can now be closed.

PR #245, opened 2016-08-24
Allow users to choose their own font family and size.

JuBan1:
A picture says more than 1000 words.

The default font isn't always sensible, and using zoom to enlarge text isn't quite as great since it blows up everything. Here's what is needed to change fonts. It's done in the same way themes are set. The change is kept in the config.

I initially shelfed this because it was somewhat unstable, but with @danieleds 's simple Zenburn fix all issues went away. I didn't find any bugs with this. Happy testing.

Also requested in #152.

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -406,12 +406,11 @@ UiDriver.registerEventHandler("C_CMD_SET_FONT", function(msg, data, prevReturn) 
     var styleTag = document.getElementById('userFont');
 
     if( styleTag  ){ 
-        styleTag.innerHTML = ".editor > .CodeMirror { " + fontFamily + fontSize + " }";
-        styleTag.appendTo('head');
+        styleTag.innerHTML = "div.editor > .CodeMirror { " + fontFamily + fontSize + " }";
     }else{
         styleTag = document.createElement("style");
         styleTag.id = 'userFont';
-        styleTag.innerHTML = ".editor > .CodeMirror { " + fontFamily + fontSize + " }";
+        styleTag.innerHTML = "div.editor > .CodeMirror { " + fontFamily + fontSize + " }";
         document.getElementsByTagName("head")[0].appendChild( styleTag );
     }
 
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -403,18 +403,16 @@ UiDriver.registerEventHandler("C_CMD_SET_FONT", function(msg, data, prevReturn) 
 
     var fontFamily = data.family != "" ? ("font-family:'" + data.family + "';") : "";
 
-
     var styleTag = document.getElementById('userFont');
 
-    if( styleTag  ){
+    if( styleTag  ){ 
         styleTag.innerHTML = ".editor > .CodeMirror { " + fontFamily + fontSize + " }";
-        return true;
+        styleTag.appendTo('head');
     }else{
         styleTag = document.createElement("style");
         styleTag.id = 'userFont';
         styleTag.innerHTML = ".editor > .CodeMirror { " + fontFamily + fontSize + " }";
         document.getElementsByTagName("head")[0].appendChild( styleTag );
-        return true;
     }
 
 });
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -397,6 +397,28 @@ UiDriver.registerEventHandler("C_CMD_SET_THEME", function(msg, data, prevReturn)
     editor.setOption("theme", data.name);
 });
 
+UiDriver.registerEventHandler("C_CMD_SET_FONT", function(msg, data, prevReturn) {
+
+    var fontSize = (data.size != "" && data.size>0) ? ("font-size:" + data.size + "px;") : "";
+
+    var fontFamily = data.family != "" ? ("font-family:'" + data.family + "';") : "";
+
+
+    var styleTag = document.getElementById('userFont');
+
+    if( styleTag  ){
+        styleTag.innerHTML = ".editor > .CodeMirror { " + fontFamily + fontSize + " }";
+        return true;
+    }else{
+        styleTag = document.createElement("style");
+        styleTag.id = 'userFont';
+        styleTag.innerHTML = ".editor > .CodeMirror { " + fontFamily + fontSize + " }";
+        document.getElementsByTagName("head")[0].appendChild( styleTag );
+        return true;
+    }
+
+});
+
 UiDriver.registerEventHandler("C_CMD_SET_OVERWRITE", function(msg, data, prevReturn) {
     editor.toggleOverwrite(data);
 });
--- a/src/ui/frmpreferences.cpp
+++ b/src/ui/frmpreferences.cpp
@@ -169,6 +169,9 @@ void frmPreferences::on_buttonBox_accepted()
             // Set new theme
             editor->setTheme(newTheme);
 
+            //Set font override
+            editor->applyGlobalFontOverride();
+
             // Reset language-dependent settings (e.g. tab settings)
             editor->setLanguage(editor->language());
 
--- a/src/ui/frmpreferences.cpp
+++ b/src/ui/frmpreferences.cpp
@@ -46,7 +46,7 @@ frmPreferences::frmPreferences(TopEditorContainer *topEditorContainer, QWidget *
     ui->chkWarnForDifferentIndentation->setChecked(s.value("warnForDifferentIndentation", true).toBool());
 
     loadLanguages(&s);
-    loadColorSchemes(&s);
+    loadAppearanceTab(&s);
     loadTranslations(&s);
     loadShortcuts(&s);
 
@@ -142,7 +142,7 @@ void frmPreferences::on_buttonBox_accepted()
     s.setValue("warnForDifferentIndentation", ui->chkWarnForDifferentIndentation->isChecked());
 
     saveLanguages(&s);
-    saveColorScheme(&s);
+    saveAppearanceTab(&s);
     saveTranslation(&s);
     saveShortcuts(&s);
     s.setValue("Search/SearchAsIType", ui->chkSearch_SearchAsIType->isChecked());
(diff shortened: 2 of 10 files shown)


danieleds:
Thanks! I'll check it out soon

danieleds:
Hi! I did some tests and it works great :)

Could you add some documentation over your methods?
Also, I think the change in the submodule at `src/editor/libs/codemirror` is not needed. Can you remove it from the pull request?

src/editor/app.js
```diff
@@ -390,6 +390,26 @@ UiDriver.registerEventHandler("C_CMD_SET_THEME", function(msg, data, prevReturn)
...
+
+    var fontSize = (data.size != "" && data.size>0) ? ("font-size:" + data.size + "px;") : "";
+
+    var fontFamily = data.family != "" ? ("font-family:'" + data.family + "';") : "";
+
+    var styleTag = document.getElementById('userFont');
+
+    if( styleTag  ){ 
+        styleTag.innerHTML = ".editor > .CodeMirror { " + fontFamily + fontSize + " }";
+        styleTag.appendTo('head');
```
danieleds: This looks wrong. Isn't styleTag already in the DOM?

Teklad: Actually, it would probably be best to change the styles directly like I 
did for the symbols CodeMirror add-on.  This removes the need to keep 
appending tags to override the previous one.

danieleds: He's already doing that*, but for some reason he forgot to remove line 403, I think.

*kind of... you were changing the style directly, this is replacing the innerHTML.

Teklad: Well I guess it's the same effect, lol.  I just worry about using 
innerHTML overwriting something important in other use cases.  It may 
not apply here though.

danieleds: @Teklad I agree that your solution is way more elegant. In this simple case though I think it's irrelevant whether to use one solution over another.

JuBan1:
I'll add some documentation a little later. 

About the submodule change... I'm not sure how that happened. I don't even know what that file does. Must've been changed automagically somehow. I'll remove it when I add documentation.

src/editor/app.js
JuBan1: I used the setTheme() function as a guide. The append To() call is also in line 383 in setTheme(), it basically serves to move the style tag to the very end of the head tag. That's needed because setTheme() does the same. If the style tag weren't moved down as well, changing themes would override the contents of the style tag. 

Could someone link me to @Teklad 's alternative? I'd love to take a look.

danieleds: @JuBan1 Pseudocode:

    for (var sheet of document.styleSheets) {
        if (sheet.ownerNode && sheet.ownerNode.id === 'userFont') {
            sheet.cssRules[0].style.fontFamily = something;
            sheet.cssRules[0].style.fontSize = something
        }
    }

I feel like this is more efficient, but I didn't test it. It all depends if it's more efficient to loop through DOM nodes or to re-parse CSS code.

danieleds: @JuBan1 `appendTo` is not a member function of `styleTag`, it actually is a jQuery function, so right now the only thing it does is to throw an error :)

I tried to comment it and everything still works, I still miss why would it be needed in any place other than `C_CMD_SET_THEME`. Can you explain it again?

JuBan1: > appendTo is not a member function of styleTag, it actually is a jQuery function, so right now the only thing it does is to throw an error :)

Welp, goes to show that my interactions with javascript generally involve a lot of duct tape and luck.

> Can you explain it again?

So this is how I understood it: the C_CMD_SET_THEME call appends its link tag to <head>, then C_CCMD_SET_FONT appends its style tag. So after that the end of <head> looks like this:

        ....
        <link type="css" href="theme.css" blah blah/>
        <style> .editor > .CodeMirror { font-family: "this"; font-size: "that" } </style>
    </head>

If the user changes the theme afterwards, C_CMD_SET_THEME is called again and the updates link tag is again appended to head:

        ....
        <style> .editor > .CodeMirror { font-family: "this"; font-size: "that" } </style>
        <link type="css" href="new_theme.css" blah blah/>
    </head>

However, if new_theme.css modifies font-family or font-size on their own (only a few themes do that), the data from the <style> tag will be overridden since new_theme.css comes after it. That's why I'm calling C_CMD_SET_FONT again after any change of themes, with the intention to move the <style> tab down again.

danieleds: Even if a theme modifies the font on its own, the selector `.editor > .CodeMirror` is more specific (and can be made even more specific at will) so there is no problem, it always win.

If you can provide an example of a theme that modifies the font, then we can test if the current selector is specific enough.

JuBan1: pastel-on-dark changes font size:

```
//from pastel-on-dark.css
.cm-s-pastel-on-dark.CodeMirror {
	background: #2c2827;
	color: #8F938F;
	line-height: 1.5;
	font-size: 14px;
}
```

danieleds: Ok then, change your selector from `.editor > .CodeMirror` to `div.editor > .CodeMirror` and it should work. No need to move stylesheets around ;-)

Anyway, that's something that is wrong within the theme itself. It should have specified a relative size, not an absolute one. I'll submit a bug report upstream.

danieleds:
Font size bug report submitted here: https://github.com/codemirror/CodeMirror/issues/4206

danieleds:
The font-size bug has been fixed upstream.

@JuBan1 news?

JuBan1:
I *think* I've got it now. Changes made.

danieleds:
Thanks.

If you take a look at the changes here on GitHub, you'll notice that there are a lot of unrelated changes. This is because you updated your fork with the newest commits (which is a good thing) but you did a merge instead of a rebase.

Right now it's going to be difficult for me to review the actual important changes. Read [here](https://gist.github.com/Chaser324/ce0505fbed06b947d962) for instructions on how to properly keep your fork up to date.

If you can manage to fix it it's great :)
Otherwise, tomorrow night I'll try to fix it by myself for this time.

JuBan1:
Jesus. So I understand what happened, but have no idea how to reverse it. But closing and re-creating the pull request should remove the offending changes, should it not?

danieleds:
I have no idea if this will work, but you can try:

```
git fetch upstream
git checkout master
git merge upstream/master

git checkout userfonts
git rebase master
```

(replace "upstream" if you named yours differently)

And then push.

If it doesn't work, I'll take care tomorrow

JuBan1:
Seems like it works now. I rebased a copy of userfonts on master and force pushed the changes back up into my fork. Or something like that. Phew.

Teklad:
I keep the document Daniel linked to you a while ago saved in my 
bookmarks for when I'm doing something.  It helps quite a bit, 
especially during rebasing and things of that nature. Following that 
guide alone has made git so much easier to work with. :)

danieleds:
@JuBan1 nice and clean!

src/ui/frmpreferences.cpp
```diff
@@ -284,9 +310,18 @@ void frmPreferences::saveLanguages(QSettings *s)
...
     s->setValue("Appearance/ColorScheme", ui->cmbColorScheme->currentData().toString());
+
+    QString font = ui->cmbFontFamilies->isEnabled() ? ui->cmbFontFamilies->currentFont().family() : "";
+    int size = ui->spnFontSize->isEnabled() ? ui->spnFontSize->value() : 0;
+
+    Editor::setGlobalFontFamily( font );
+    Editor::setGlobalFontSize( size );
+
+    s->setValue("overrideFontFamily", font);
+    s->setValue("overrideFontSize", size);
```
danieleds: These should be under `Appearance/OverrideFont*`, as happens with "ColorScheme" (they're both in the Appearance tab).

src/editor/app.js
```diff
@@ -397,6 +397,25 @@ UiDriver.registerEventHandler("C_CMD_SET_THEME", function(msg, data, prevReturn)
     editor.setOption("theme", data.name);
 });
 
+UiDriver.registerEventHandler("C_CMD_SET_FONT", function(msg, data, prevReturn) {
+
+    var fontSize = (data.size != "" && data.size>0) ? ("font-size:" + data.size + "px;") : "";
+
+    var fontFamily = data.family != "" ? ("font-family:'" + data.family + "';") : "";
```
danieleds: `data.family` should be escaped in case it contains quotes.
Replacing `'` with `\'` should be enough.

And just to be on the safe side, at line 400 add a `+` before data.size, like this:

    ... ("font-size:" + (+data.size) + "px;") ...

danieleds:
Ok I'm fixing these last problems myself, merging soon.

danieleds:
Merged as e246e8158ad58da049cd11667dc1d2d156d40b04!
I hope to release a new version soon with this and @Teklad 's support for special symbols.

src/ui/main.cpp
```diff
@@ -94,7 +94,6 @@ int main(int argc, char *argv[])
     // There are no other instances: start a new server.
     a.startServer();
 
-    Editor::addEditorToBuffer();
```
danieleds: Why was this removed?

JuBan1: That was a mistake. The cose I submitted it worked too, but it should have read:

```
    // Apply font settings. Best to do that before any editors are opened to prevent
    // unneeded work. Otherwise we'd have to apply it to all editors currently created.
    QString fontFam = settings.value("overrideFontFamily", "").toString();
    int fontSz = settings.value("overrideFontSize", 0).toInt();

    Editor::setGlobalFontFamily(fontFam);
    Editor::setGlobalFontSize(fontSz);

    Editor::addEditorToBuffer();
```

This was meant to set the font override before any editors were added. But with your changes to how fonts are applied, this isn't necessary anymore.

danieleds: Oh, got it!

By the way, I've just pushed the new nqq version with your changes to the ppa. The updated packages will become available in a few minutes ;)

PR #249, opened 2016-08-28
Sessions, Finally.

JuBan1:
The Pull Request is back- and this time with a vengeance.

* Sessions are saved as XML files to where ever you want them. They only keep track of opened files (so no temporary documents) and they do not cache any unsaved changes, just like in N++. Scroll positions and different Views are saved as well.

* Load/Save Session from the Menu now work. 

* "Remember my tabs on closing" feature now uses Sessions behind the curtains, but still saves any unsaved changes to a cache. As with the old PR, this feature can be en/disabled in the options menu and is enabled by default.

### DIFF
--- a/src/ui/Sessions/sessions.cpp
+++ b/src/ui/Sessions/sessions.cpp
@@ -1,0 +1,115 @@
+#include "include/Sessions/sessions.h"
+
+/* Session XML structure:
+ *
+ * <Notepadqq>
+ *		<View>
+ *			<Tab filePath="xxx" .../>
+ *			<Tab filePath="xxx" ...b/>
+ *		</View>
+ *		<View>
+ *			...
+ *		</View>
+ *		...
+ * <Notepadqq>
+ *
+ * */
+
+std::vector<ViewData> SessionReader::readData(bool* outSuccess) {
+    std::vector<ViewData> result;
+
+    if (m_reader.readNextStartElement()) {
+        if (m_reader.name() == "Notepadqq"){
+            result = readViewData();
+        }
+        else
+            m_reader.raiseError(QObject::tr("Error reading session file"));
+    };
+
+    if(outSuccess != nullptr)
+        *outSuccess = !m_reader.error();
+
+    return result;
+}
+
+QString SessionReader::getError(){
+    return m_reader.errorString();
+}
+
+std::vector<ViewData> SessionReader::readViewData() {
+    std::vector<ViewData> result;
+
+    while(m_reader.readNextStartElement()){
+        if(m_reader.name() == "View"){
+            ViewData vd;
+            vd.tabs = readTabData();
+            result.push_back(vd);
+        }
+        else
+            m_reader.skipCurrentElement();
+    }
+
+
+    return result;
+}
+
+std::vector<TabData> SessionReader::readTabData() {
+    std::vector<TabData> result;
+
+    while(m_reader.readNextStartElement()){
+        if(m_reader.name() == "Tab"){
+            const QXmlStreamAttributes& attrs = m_reader.attributes();
+
+            TabData td;
+            td.filePath = attrs.value("filePath").toString();
+            td.cacheFilePath = attrs.value("cacheFilePath").toString();
+            td.scrollX = attrs.value("scrollX").toString().toInt();
+            td.scrollY = attrs.value("scrollY").toString().toInt();
+            result.push_back(td);
+
+            m_reader.readElementText();
+        }
+        else
+            m_reader.skipCurrentElement();
+    }
+
+    return result;
+}
+
+SessionWriter::SessionWriter(QFile& destination)
+    : m_writer(&destination)
+{
+    m_writer.setAutoFormatting(true);
+
+    m_writer.writeStartDocument();
+    m_writer.writeStartElement("Notepadqq");
+}
+
+SessionWriter::~SessionWriter(){
+    m_writer.writeEndElement();
+    m_writer.writeEndDocument();
+}
+
+void SessionWriter::addViewData(const ViewData& vd){
+    if(vd.tabs.empty()) return;
+
+    m_writer.writeStartElement("View");
+
+    for(auto&& tab : vd.tabs)
+        addTabData(tab);
+
+    m_writer.writeEndElement();
+}
+
+void SessionWriter::addTabData(const TabData& td){
+    m_writer.writeStartElement("Tab");
+
+    QXmlStreamAttributes attrs;
+    attrs.push_back( QXmlStreamAttribute("filePath", td.filePath) );
+    attrs.push_back( QXmlStreamAttribute("cacheFilePath", td.cacheFilePath) );
+    attrs.push_back( QXmlStreamAttribute("scrollX", QString::number(td.scrollX)) );
+    attrs.push_back( QXmlStreamAttribute("scrollY", QString::number(td.scrollY)) );
+    m_writer.writeAttributes(attrs);
+
+    m_writer.writeEndElement();
+}
--- a/src/ui/Sessions/sessions.cpp
+++ b/src/ui/Sessions/sessions.cpp
@@ -5,13 +5,22 @@
  * <Notepadqq>
  *      <View>
  *          <Tab filePath="xxx" .../>
- *          <Tab filePath="xxx" ...b/>
+ *          <Tab filePath="xxx" .../>
  *      </View>
  *      <View>
  *          ...
  *      </View>
  *      ...
  * <Notepadqq>
+ *
+ *
+ * All currently available attributes for <Tab>:
+ * -> string filePath - path to the file if it exists
+ * -> string cacheFilePath - path to the cache file if it exists
+ * -> int scrollX - horizontal scroll position
+ * -> int scrollY - vertical scroll position
+ * -> long int lastModified - optional, last modification date (in msecs since epoch) of the file point to in filePath
+ * -> int active - optional, value is "1" if this tab is the open one in the tabview, otherwise "0".
  *
  * */
 
@@ -63,9 +72,11 @@ std::vector<TabData> SessionReader::readTabData() {
             TabData td;
             td.filePath = attrs.value("filePath").toString();
             td.cacheFilePath = attrs.value("cacheFilePath").toString();
-            td.scrollX = attrs.value("scrollX").toString().toInt();
-            td.scrollY = attrs.value("scrollY").toString().toInt();
+            td.scrollX = attrs.value("scrollX").toInt();
+            td.scrollY = attrs.value("scrollY").toInt();
             td.lastModified = attrs.value("lastModified").toLongLong();
+            td.active = attrs.value("active").toInt() != 0;
+
             result.push_back(td);
 
             m_reader.readElementText();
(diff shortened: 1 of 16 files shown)


danieleds:
Cool! B) I'll try this soon.

JuBan1:
Sessions are in an alright state by now. It's rebased and I cleaned some things up. 

Also, when attempting to save a tab that was restored from cache Nqq will now warn you if the file was modified in the meantime, just like it already does when files are modified while opened in Nqq.

There is one bug that I'm having trouble fixing:

* Open two tabs, each in one View. Write something into them so they'll get cached.
* Close and reopen Nqq.
* The restored tabs now don't behave properly. First, the blinking caret is present in both views as if they both have focus. Next, try to click on the left tab and CTRL+S to save it. You'll get a save prompt for the right tab.

For some reason the focus/currently active tab or tabwidget isn't properly handled. Clicking on the left and right tab in alternation a few times resolves the problem (you can see in Nqq's titlebar how the file name isn't properly shown at first, but after a few times switching tabs it works again).

The whole editor/tab thing is pretty complicated with all the signals being passed around, so I'd appreciate it if you could take a look, @danieleds.

danieleds:
Taking a look now

danieleds:
I think there is a problem with the settings, what I choose in "Preferences" is not persisted (the flag always end up being disabled)

danieleds:
I'm trying to find that "focus bug" too, in the meantime it could help if you provide even more documentation... I'll highlight the pieces that I feel need more love.

src/ui/include/docengine.h
```diff
@@ -51,6 +51,7 @@ class DocEngine : public QObject
 
     bool loadDocuments(const QList<QUrl> &fileNames, EditorTabWidget *tabWidget);
     bool loadDocument(const QUrl &fileName, EditorTabWidget *tabWidget);
+    bool loadDocumentSilent(const QUrl &fileName, EditorTabWidget *tabWidget);
```
danieleds: Big comment block needed here explaining what "silent" means.

JuBan1:
Looks like the two lines for setting/getting the option in the preferences menu were accidentally purged when I resolved the merge conflicts. Will submit them together with the requested documentation.

src/ui/include/docengine.h
```diff
@@ -74,7 +75,7 @@ class DocEngine : public QObject
     bool read(QFile *file, Editor *editor);
     bool read(QFile *file, Editor *editor, QTextCodec *codec, bool bom);
     // FIXME Separate from reload
-    bool loadDocuments(const QList<QUrl> &fileNames, EditorTabWidget *tabWidget, const bool reload, QTextCodec *codec, bool bom);
+    bool loadDocuments(const QList<QUrl> &fileNames, EditorTabWidget *tabWidget, const bool reload, QTextCodec *codec, bool bom, bool rememberLastSelectedDir=true);
```
danieleds: Need a comment block to explain what `rememberLastSelectedDir` does.

Don't worry about documenting the other parameters, I'll need to clean my own mess some day.

src/ui/include/mainwindow.h
```diff
@@ -205,6 +216,42 @@ private slots:
...
+
+	/**
+	 * @brief Saves a session as an XML file
+	 * @param filePath Path to where the XML file should be created.
+	 * @param cacheModifiedFiles If true, dirty tabs will be written into the cache directory.
+	 *        Only use this for the remember-my-tabs feature. Multiple sessions writing to
+	 *        the cache directory will end up in data loss.
+	 * @return Whether the save has been successful.
+	 */
+    bool                saveSession(QString filePath, bool cacheModifiedFiles);
```
danieleds: Warning: tab indentation sneaking all around

src/ui/include/mainwindow.h
```diff
@@ -205,6 +216,42 @@ private slots:
...
+	 *        in the cache.
+	 */
+    bool                saveTabsToCache();
+    void                restoreTabsFromCache();
+
+	/**
+	 * @brief Closes all open tabs but asks the user for permission before discarding changes.
+	 * @return Whether all files have been properly closed.
+	 */
+    bool                closeAllTabs();
```
danieleds: It's important here to specify that this methods doesn't *actually* close the tabs. It acts *as-if*, even saving the changes if the user says so, but never actually closing them.

src/ui/mainwindow.cpp
```diff
@@ -2226,3 +2428,61 @@ void MainWindow::on_actionToggle_Smart_Indent_toggled(bool on)
...
+    QString recentFolder = QUrl::fromLocalFile(
+                                m_settings.General.getLastSelectedSessionDir())
+                                .toLocalFile();
+
+    QString fileName = QFileDialog::getOpenFileName(
+                this,
+                "Open Session...",
+                recentFolder,
+                "Session file (*.xml);;Any file (*)",
+                0, 0);
```
danieleds: Missing two tr()

danieleds:
Okay I'm finished with the boring notes for now

src/ui/include/docengine.h
JuBan1: I was thinking about giving loadDocument a once-over with another Pull Request. I left it out of this one because it'd just make things messier. So hopefully the `rememberLastSelectedDir` won't actually stay there for long.

JuBan1:
Right, got everything. Other than these changes I also renamed `closeAllTabs` to `finalizeAllTabs` to remove ambiguity.

src/ui/frmpreferences.ui
```diff
@@ -114,6 +114,13 @@
           </widget>
          </item>
          <item>
+          <widget class="QCheckBox" name="chkRememberSession">
+           <property name="text">
+            <string>Remember openened tabs when closing Notepadqq</string>
```
danieleds: "openened"... I'm pointing out these trivial issues now because otherwise I'm sure I'm going to forget about them.

src/ui/mainwindow.cpp
```diff
@@ -358,6 +367,227 @@ void MainWindow::createStatusBar()
...
+    m_dontUpdateRecentDocs = false;
+
+    //Stop if we haven't added any views at all, otherwise we have to clean up after ourselves.
+    if(viewCounter <= 0)
+        return;
+
+    //If the last tabwidget still has no tabs in it at this point, we'll have to delete it.
+    EditorTabWidget* lastTabW = m_topEditorContainer->tabWidget( m_topEditorContainer->count() -1);
+    if (lastTabW->count() == 0)
+        delete lastTabW;
```
danieleds: use `MainWindow::removeTabWidgetIfEmpty()`

danieleds:
Okay I've got it. I will update the master branch with a generic fix.

JuBan1:
Great. With all issues here being fixed, I think this is ready to be merged on my end.

danieleds:
I've found a segfault which is caused by the "focus issue".

**Setup steps:**
 1. Ensure "remember tabs" is enabled
 2. In the only empty editor, write something to make it dirty
 3. Add a new tab, move it to the other view
 4. Write something in the new tab to make it dirty
 5. Now that you have two dirty tabs side by side, close notepadqq.

**Segfault steps**
 1. Launch notepadqq. It will restore the session of the setup steps
 2. Close the tab on the right
 3. Add a new tab and it crashes

I've encountered some unexpected issues with fixing that bug, I'll finish later than I thought.

danieleds:
Okay, I need more time, so for now I've merged some support for a workaround.

Rebase on master and then add these lines to your mainwindow.cpp, into `MainWindow::loadSession`:

```
             }
 
             editor->setScrollPosition(tab.scrollX, tab.scrollY);
+            editor->clearFocus();
 
         } //end for
 
+        tabW->clearFocus();

         //In case a new tabwidget was created but no tabs were actually added to it,
         //we'll attempt to re-use the widget for the next view.

........

     if(viewCounter <= 0)
         return;
 
+    m_topEditorContainer->tabWidget(0)->editor(0)->setFocus(); // Or whatever you want to focus, perhaps the last one

     //If the last tabwidget still has no tabs in it at this point, we'll have to delete it.
     EditorTabWidget* lastTabW = m_topEditorContainer->tabWidget( m_topEditorContainer->count() -1);
     if (lastTabW->count() == 0)
         delete lastTabW;
```

JuBan1:
Done.

danieleds:
Perfect!

Can you make it save, on close, the tabs of the last closed window instead of those of the first opened window?

JuBan1:
That brings up the question of how to handle several windows, however.

Let's say I have two windows open. I close one which then counts as the last closed window and its tabs are saved. Then I close the other one. But now that *other* one is the last closed window. I can't save its tabs, however, because it would override the first save.

danieleds:
You're right, I didn't explain myself well.

> Can you make it save, on close, the tabs of the last closed window instead of those of the first opened window?

What I meant is: 
Can you make it save, *when the last window of the process is closed*, the tabs of that last window instead of those of the first opened window?

Clearer?

JuBan1:
Got it! Done.

danieleds:
It works! :)

In a few hours I'll get the chance try the load/save session menu too.

danieleds:
It's good!

I noticed that right now the "what's the active tab" information is not persisted within the session, and neither are the positions of the selections.

Do you think it is worth adding them in this PR, or instead it's better to leave them for later?

src/ui/Sessions/sessions.cpp
```diff
@@ -0,0 +1,123 @@
+#include "include/Sessions/sessions.h"
+
+/* Session XML structure:
+ *
+ * <Notepadqq>
+ *      <View>
+ *          <Tab filePath="xxx" .../>
+ *          <Tab filePath="xxx" ...b/>
+ *      </View>
```
danieleds: Add an example of the other attributes too.
Also, there's a "b".

src/ui/Sessions/sessions.cpp
```diff
@@ -0,0 +1,123 @@
...
+ *      <View>
+ *          ...
+ *      </View>
+ *      ...
+ * <Notepadqq>
+ *
+ * */
+
+std::vector<ViewData> SessionReader::readData(bool* outSuccess) {
```
danieleds: Use "pass by reference" for the out parameter

src/ui/frmpreferences.ui
```diff
@@ -86,7 +86,7 @@
      <item>
       <widget class="QStackedWidget" name="stackedWidget">
        <property name="currentIndex">
-        <number>0</number>
+        <number>5</number>
        </property>
```
danieleds: Try cleaning this manually in text mode

src/ui/frmpreferences.ui
```diff
@@ -464,8 +471,8 @@
              <rect>
               <x>0</x>
               <y>0</y>
-              <width>403</width>
-              <height>196</height>
+              <width>565</width>
+              <height>389</height>
```
danieleds: Try cleaning this manually in text mode

src/ui/mainwindow.cpp
```diff
@@ -358,6 +362,236 @@ void MainWindow::createStatusBar()
...
+
+        } // end for
+    } // end for
+
+    // Write all information to a session file
+    QFile file(filePath);
+    file.open(QIODevice::WriteOnly);
+
+    if (file.isOpen()) {
+        SessionWriter sessionWriter(file);
```
danieleds: If we can't open the file in write mode we should report it to the caller!

Suppose, as an example, that I have no write permissions over my `~/.config/Notepadqq/session.xml`. I make some important changes without saving them because I think that notepadqq will save them in the session, but instead it will close without saying anything, losing my changes.

In this case, nqq should show a dialog saying that it's not able to save the session, and then abort the closing procedure thus giving the user a chance to manually save his files.

JuBan1:
Active tab and caret position aren't big changes. If by 'selections' you mean selected text.. that's more annoying, especially because the file's content can change between saving and loading a session. The selection texts could be completely wrong then. (The scroll position and caret position possibly too, but I think that's much less noticeable.)

danieleds:
No, I was thinking about just the positions

JuBan1:
Ok, then I'd say I'll make amends later today/tomorrow and we'll merge after that. A few extra lines of code won't make this PR any fatter.

JuBan1:
Ok, all mentioned changes made with the exception of saving the cursor/caret position. That didn't work properly, probably because of your workaround? It seems like only the tab that currently has focus retains its cursor position.

danieleds:
Yeah, it's because of my workaround. I didn't think about this possible issue.

Anyway, if it's ok for you I'm going to merge... then we'll try to address this problem later (maybe using proper code instead of a dirty workaround).

JuBan1:
Good, good.

PR #250, opened 2016-08-29
Search overhaul

Teklad:
So these are the performance improvements for the Search in File functions.  I kept the scope pretty low and just focused on rewriting how the search itself is done.  

The performance improvement over the old method is really noticeable, especially on larger files.  Pretty sure I managed to squash all the odd behaviors.

Just needs a good testing to make sure there's no odd bugs being introduced here.  :)

### DIFF
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -53,6 +53,7 @@ void SearchInFilesWorker::run()
     //int totalFileMatches = 0;
     // Number of files that contain matches
     //int totalFiles = 0;
+    bool multiLine = m_regex.pattern().contains(QStringLiteral("\\n"));
 
     while (it.hasNext()) {
         m_stopMutex.lock();
@@ -70,7 +71,6 @@ void SearchInFilesWorker::run()
         QFile f(fileName);
         DocEngine::DecodedText decodedText;
         bool retry;
-        bool multiLine = m_regex.pattern().contains(QStringLiteral("\\n"));
         do {
             retry = false;
             decodedText = DocEngine::readToString(&f);
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -4,7 +4,7 @@
 #include <QDirIterator>
 #include <QMessageBox>
 #include <QThread>
-
+#include <QIODevice>
 SearchInFilesWorker::SearchInFilesWorker(const QString &string, const QString &path, const QStringList &filters, const SearchHelpers::SearchMode &searchMode, const SearchHelpers::SearchOptions &searchOptions)
     : m_string(string),
       m_path(path),
@@ -133,10 +133,8 @@ FileSearchResult::FileResult SearchInFilesWorker::searchSingleLineRegExp(const Q
     int i = 0;
     int column; 
     int absoluteColumn = 0;
-    int eolSize = 1;
     QRegularExpressionMatch match;
 
-    if (content.indexOf("\r\n") != -1) eolSize = 2;
     while (!(line=stream.readLine()).isNull()) {
         if (m_stop) break;
         match = m_regex.match(line);
@@ -148,10 +146,15 @@ FileSearchResult::FileResult SearchInFilesWorker::searchSingleLineRegExp(const Q
             match = m_regex.match(line, column + match.capturedLength());
             column = match.capturedStart();
             m_matchCount++;
-            if (m_matchCount%50) QThread::msleep(1);
+            if (m_matchCount%50) QThread::usleep(1);
         }
         i++;
-        absoluteColumn += line.length() + eolSize;
+        absoluteColumn += line.length();
+
+        //Check line ending per line to be safe.
+        if (content.midRef(absoluteColumn,2) == "\r\n") absoluteColumn += 2;
+        else absoluteColumn++;
+        
     }
     return structFileResult;
 
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -237,7 +237,7 @@ FileSearchResult::Result SearchInFilesWorker::buildResult(const int &line, const
         res.previewBeforeMatch = lineContent.mid(0,column);
     }
     res.match = lineContent.mid(column,matchLen);
-    res.previewAfterMatch = lineContent.mid(column + matchLen,matchLen + 100);
+    res.previewAfterMatch = lineContent.mid(column + matchLen,matchLen + 50);
     res.matchStartLine = line;
     res.matchStartCol = column;
     res.matchEndLine = line;
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -133,7 +133,6 @@ FileSearchResult::FileResult SearchInFilesWorker::searchSingleLineRegExp(const Q
                 structFileResult.results.append(buildResult(i, column, streamPosition + column, line, resultLength));
                 column = line.indexOf(m_string, column + resultLength, caseSensitive);
                 m_matchCount++;
-                if (m_matchCount%50) QThread::usleep(1);
             }
             i++;
             if (content->midRef(streamPosition + lineLength, 2) == "\r\n") streamPosition += lineLength + 2;
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -190,7 +190,7 @@ FileSearchResult::FileResult SearchInFilesWorker::searchMultiLineRegExp(const QS
         int i;
         line = -1;
         for (i=1; i<lineStart.size(); i++) {
-            if (lineStart[i] > column) {
+            if (lineStart.at(i) > column) {
                 line = i-1;
                 break;
             }
@@ -198,8 +198,9 @@ FileSearchResult::FileResult SearchInFilesWorker::searchMultiLineRegExp(const QS
         if (line == -1) {
             break;
         }
-        structFileResult.results.append(buildResult(line, (column - lineStart[line]), fullDoc.mid(lineStart[line], column - lineStart[line])+match.captured(), match.capturedLength()));
-        match = tmpRegExp.match(fullDoc, column + match.capturedLength());
+        int matchLen = match.capturedLength();
+        structFileResult.results.append(buildResult(line, (column - lineStart.at(line)), fullDoc.mid(lineStart.at(line), column - lineStart.at(line))+match.captured(), matchLen));
+        match = tmpRegExp.match(fullDoc, column + matchLen);
         column = match.capturedStart();
         m_matchCount++;
         // NOTE: This sleep is here so that the main thread will get a chance to
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -4,7 +4,6 @@
 #include <QDirIterator>
 #include <QMessageBox>
 #include <QThread>
-#include <QIODevice>
 SearchInFilesWorker::SearchInFilesWorker(const QString &string, const QString &path, const QStringList &filters, const SearchHelpers::SearchMode &searchMode, const SearchHelpers::SearchOptions &searchOptions)
     : m_string(string),
       m_path(path),
@@ -22,20 +21,20 @@ SearchInFilesWorker::~SearchInFilesWorker()
 
 void SearchInFilesWorker::run()
 {
-    // Search string converted to a regex
-    QString rawSearch = frmSearchReplace::rawSearchString(m_string, m_searchMode, m_searchOptions);
-
+    QString rawSearch = m_string;
+    bool multiLine = false; 
+    
     QFlags<QRegularExpression::PatternOption> options = QRegularExpression::NoPatternOption;
-    if (m_searchOptions.MatchCase == false) {
-        options |= QRegularExpression::CaseInsensitiveOption;
+    if(m_searchMode != SearchHelpers::SearchMode::PlainText) {
+        if (m_searchOptions.MatchCase == false) {
+            options |= QRegularExpression::CaseInsensitiveOption;
+        }
+        rawSearch = frmSearchReplace::rawSearchString(m_string, m_searchMode, m_searchOptions);
+        m_regex.setPattern(rawSearch);
+        m_regex.setPatternOptions(options);
+        m_regex.optimize();
+        multiLine = m_regex.pattern().contains(QStringLiteral("\\n"));
     }
-
-    m_regex.setPattern(rawSearch);
-    m_regex.setPatternOptions(options);
-
-    //Pre-optimize here since we are typically going to be traversing many files.
-    m_regex.optimize();
-
 
     // Search result structure
     FileSearchResult::SearchResult structSearchResult;
@@ -48,12 +47,6 @@ void SearchInFilesWorker::run()
 
     // Iterator used to find files in the specified directory
     QDirIterator it(m_path, m_filters, QDir::Files | QDir::Readable | QDir::Hidden, dirIteratorOptions);
-
-    // Total number of matches in all the files
-    //int totalFileMatches = 0;
-    // Number of files that contain matches
-    //int totalFiles = 0;
-    bool multiLine = m_regex.pattern().contains(QStringLiteral("\\n"));
 
     while (it.hasNext()) {
         m_stopMutex.lock();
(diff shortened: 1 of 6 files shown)


danieleds:
On what kind of files did you try it? These are my results:

data

I have tried two different test data, one was the notepadqq repository itself and the other was another large repository of mine.

I have repeated 3 tests for each version and test data, for a total of 12 tests, but I've not been able to notice improvements (actually, for what I can see the current implementation is faster).

(side note: you're missing some hits, probably from multiple matches on the same line)

danieleds:
Forgot to mention, I didn't use any filename filter.

If you can provide information on your test data, that would be useful :)

src/ui/Search/searchinfilesworker.cpp
```diff
@@ -89,40 +91,18 @@ void SearchInFilesWorker::run()
...
-            QRegularExpressionMatch match = i.next();
-            QStringList matches = match.capturedTexts();
-
-            if (!matches[0].isEmpty()) {
-                structFileResult.results.append(buildResult(match, &content));
-
-                curFileMatches++;
-                //totalFileMatches++;
-            }
+        if (m_regex.pattern().contains(QStringLiteral("\\n"))) {
```
JuBan1: The result of the `contains` call could be computed once before the loop instead of being computed once every iteration.

JuBan1:
So I quickly ran the search for the term "QThread" on the Qt5 source, 139,316 items, totalling 5.3 GB.

Branch | Config | Result | Finished in
----------|-----------|---------|------------
New | Debug | 22447 hits in 1267 files | 02mins 49secs
New | Release | 22447 hits in 1267 files | 01mins 24secs
Current | Debug | ---- | I didn't even bother
Current | Release | ---- | `terminate called after throwing an instance of 'std::bad_alloc'`. It managed to run for over 30 minutes and chew through *all of my 16GiB RAM* before dying.

I think the results speak for themselves.

I can echo the observation about larger files. There are a lot of >50 MiB files which is where the current search really choked up.

On my machine it wasn't only faster but also more efficient. CPU utilization was much lower. The current implementation basically maxes out one core permanently. The new branch doesn't even do that, yet is so much faster.

I wonder if this could be even more improved by running several workers in parallel and splitting the workload between them.

Teklad:
So the idea behind the branch isn't to get improvements for just text 
files.... the point was to give a dramatic increase in large 
files/binary where the current search chokes up....

For example, if you compile openmw from github then try to search 
without a filter it literally won't complete for many minutes, if not 
longer.  The new branch is able to do this within a minute.

I still need to smooth out a few things from the conversion, but the 
overall improvement is definitely there.  It appears that the match 
iterator has extremely variable performance across different systems, 
which is really odd.  One particular file when searching within openmw 
would literally only give me 2 results per second due to the extremely 
long lines.

I'll do some more modifications to improve this further.... as far as 
more threads go... because the CPU is lowered this won't make much of a 
difference because at that point the only thing that is bottle-necking 
is the hard drive I would assume... multiple threads might actually make 
it slower in this case.

I've tried various different searches... but the current implementation 
is slower for me, filtered or not.

Teklad:
On a side note. I'm going to move a few things around to see if I can
improve performance further once I get the bugs ironed out. Any suggestions
for further performance improvements are welcome.

This is just the initial implementation.  I'm sure squeezing a few extra
bits of performance is very possible in places.

Teklad:
Okay, so I've made a few minor tweaks to the code.  Try your test again danieleds and see if the performance is getting better.  It should be pretty close or faster than the current implementation with these changes.

JuBan1:
Branch | Config | Result | Finished in
----------|-----------|---------|------------
Previous | Release | 22447 hits in 1267 files | 01mins 24secs
Updated | Release | 23032 hits in 1267 files | 00mins 58secs

Note that the number of hits has gone up with the recent additions. Unfortunately I can't determine the true result. I ran the directory through Sublime but it skips the binary files and only gets ~18000 results.

On that note, an option to dis-/enable searching through non-text files might be a good idea.

Teklad:
Yea... The results changed as a result of the tweaks. Should be more
accurate now.

It would probably be wise to allow the maximum line length to be
modifiable  in the preferences menu based on their own needs/system specs.

An advanced settings menu would come in handy for settings that shouldn't
be touched often.

Teklad:
Definitely should probably do text mode search by default. Binary file
searching isn't a thing somebody uses under normal circumstances.

danieleds:
Sorry, I've not had a chance to try it yet, however:

> For example, if you compile openmw from github then try to search without a filter it literally won't complete for many minutes, if not longer.

Ok, so you're saying that the current implementation is slower *only* if the search involves binary files?

> as far as more threads go... because the CPU is lowered this won't make much of a difference because at that point the only thing that is bottle-necking is the hard drive I would assume... multiple threads might actually make it slower in this case.

True.

> It would probably be wise to allow the maximum line length to be modifiable in the preferences menu based on their own needs/system specs.

Uhm, why is there a maximum line length? I would hate an editor that says to me "hey, I've found *n* results", while actually they are *n*+*k*.

I think I'll have some time later to test your latest version :)

JuBan1:
> Uhm, why is there a maximum line length? I would hate an editor that says to me "hey, I've found n results", while actually they are n+k.

The line length is for the result string that is shown in the docklet. Usually it shows you the entire line in which the result was found, but binary files don't have line breaks and so the result string would be the entire length from the position of the result until the eof.

This isn't just useful for binary files though. Like some of those .js files that are optimized for size where all unneeded characters such as line endings have been removed.

Teklad:
Its pretty common practice to limit line length when traversing large bits
of text...

Actually the current implementation lacks any concept of this.  Because the
search is threaded, the current implementation causes out of memory errors
on very large files.

Also, for me the current one is slow even with just text files.  I thought
it was just my machine until we got the bug report.

danieleds:
> I think the line length is for the result string that is shown in the docklet.

Now it makes sense, the limit is only applied to the text shown in the UI.
I didn't get it at first.
This is obviously good, and a nice improvement would be to show not the first *n* characters but, instead, the *n* characters *around* the matched text.

danieleds:
> Not necessarily. First, I ran those tests on a 3.5 GHz Xeon processor - faster than most people's workstations. Second, I ran this on a normal HDD. SSDs are relatively common nowadays and offer much higher read speeds. 

The actual advantage has to be verified. Honestly I think that parallelizing an IO-bound algorithm like this is not going to improve much and, as @Teklad said, could likely slow it down. Even if an improvement was noticeable, I don't think it would be so great to justify the increased complexity required in the code.

JuBan1:
> The actual advantage has to be verified. 

Absolutely! I didn't want to sound like I was advocating for implementing in right on the spot. But I'd be interested in the result... maybe I'll give it a go if I run out of interesting features to implement.

Teklad:
I'll make the changes to get the x amount of the surrounding string.  I'm a bit concerned about loading the entire file using decodeText...  I know its extremely unlikely, but if some goofball loads up some ridiculously oversized file... its going to cause some issues.

I would really prefer to use a QFile iterator here if at all possible, that way I can have more control over my read(i.e. peek is handy, but not required).

Is the decoded content of the file absolutely necessary for getting correct results?

danieleds:
> Is the decoded content of the file absolutely necessary for getting correct results?

Unfortunately yes, we have to decode it. A text search without knowing the encoding is definitely wrong.
And this is a problem by itself, because to detect the correct encoding we basically have to *guess* with some heuristics... text files don't have headers.

A viable option could be to to detect the encoding from the first 1KB only (see this [FIXME](https://github.com/notepadqq/notepadqq/blob/master/src/ui/docengine.cpp#L491)) and then loading the file piece by piece with the correct encoding. But let's keep this for another pull request, I don't want to clutter this one with delicate changes.


Speaking of testing, I've built this bash script so that we can generate a reproducible set of test files: http://pastebin.com/DLW70BgD
I'll test the old vs new implementation over those files in a moment. Let me know if you think you need different, additional test files.

The tests I suggest to perform, so that we can actually compare our results, are
  * Search "Lorem ipsum" with Filter="long-line-500KB"
  * Search "Lorem ipsum" with Filter="long-line-5MB"
  * Search "Lorem ipsum" with Filter="multiple-lines-500KB"
  * Search "Lorem ipsum" with Filter="multiple-lines-5MB"
  * Search "abcdefg" with Filter="binary-file-500KB"

Teklad:
That seems fair enough.  The end goal of this pull request is to make it faster under all circumstances, not just the binary use case described as the issue.  I'll keep making small changes until we reach that goal.  It'll take a bit to perfect.

danieleds:
My results:

| File | Implementation | Time | Hits | Comments |
|------|---------------------|--------|-------|---------------|
| long-line-500KB | master | 04.22s | 721 | |
| long-line-500KB | ba7fe14 | 0.10s | 3 | Should search in the whole line |
| long-line-5MB | master | (unfinished) | (unfinished) | |
| long-line-5MB | ba7fe14 | 0.10s | 3 | Should search in the whole line |
| multiple-lines-500KB | master | 2.92s | 721 | |
| multiple-lines-500KB | ba7fe14 | 0.10s | 721 | |
| multiple-lines-5MB | master | > 120.00s | - | |
| multiple-lines-5MB | ba7fe14 | 0.92s | 7195 | |
| binary-file-500KB | master | 20.55s | 3908 | |
| binary-file-500KB | ba7fe14 | 0.57s | 3908 | |
| binary-file-20MB | master | (unfinished) | (unfinished) | |
| binary-file-20MB | ba7fe14 | 12.83 | 156250 | |

The new implementation is actually quite fast. But because of the fact that you're ignoring long lines, for now long-line-500KB and long-line-5MB tests are not significant. I'll repeat them as soon as you're able to read the whole lines efficiently.

JuBan1:
Mine just finished as well:


This branch

File | Result | Time
-----|--------|------
long-line-100KB | 3 hits in 1 files | 0s 0ms
long-line-5MB | 3 hits in 1 files | 0s 20ms
long-line-20MB | 3 hits in 1 files | 0s 80ms
multiple-lines-100KB | 145 hits in 1 files | 0s 10ms
multiple-lines-5MB | 7195 hits in 1 files | 0s 340ms
multiple-lines-20MB | 28777 hits in 1 files | 1s 900ms
binary-file-20MB | 156250 hits in 1 files | 9s 500ms

Master

File | Result | Time
-----|--------|------
long-line-100KB | 145 hits in 1 files | 0s 100ms
long-line-5MB | App crash | ---
long-line-20MB | --- | ---
multiple-lines-100KB | 145 hits in 1 files | 0s 80ms
multiple-lines-5MB | 7195 hits in 1 files | 2m 50s
multiple-lines-20MB | ---| ---
binary-file-20MB | --- | Too long, didn't finish

I didn't bother running the 20MB files considering the circumstances.

danieleds:
@JuBan1 Ok, so looks like the behavior is confirmed :)

The new version is extremely more efficient for large files (well, the current version is unusable). The only issue now is the "3 hits".

Teklad:
After a lot of poking around I found a memory leak in the way we typically build results, which is why long searches would fail.

Though its not exactly a memory leak in the literal sense... we weren't capping previewAfterMatch or previewBeforeMatch at all, which would in the case of binary files... quickly run a system out of memory when a match was found often.

After patching the leak I have a working search which is still fairly fast... Its a bit slower on binary files than it was for obvious reasons... but adding an option for text-only search would be cheap and easy to implement later as a GUI feature.

For reasons unknown the iterator method is slower than this method, which I've yet to wrap my head around as to why.

Give the tests a try once more and give me your results please. :)

Teklad:
So the search working really well now has surfaced an issue that wasn't apparent before.  It looks like there's an actual memory leak that happens when search results are displayed.

Upon clearing the results, the memory isn't reclaimed.  On top of that... I kept running the search on the binary-file-20MB over and over and was eventually able to cause an out of memory kill on notepadqq.

This isn't really related to this pull request, but its a bug that I'll need to investigate regardless.  Something isn't getting deleted properly... the question is what.

Teklad:
The only things I can determine right now is that there is a small memory leak somewhere DocEngine, and another happens somewhere while passing the file results struct around...

The memory leak is by no means massive, and would take a considerable amount of file searches to hit.  In theory clearing the results widget should have freed up the memory allocated by the file results struct... but that doesn't happen for some reason.

I wouldn't even know where to start looking in DocEngine though.  I've not looked at that code since we used QScintilla, lol.

JuBan1:
You could try a memory debugger like valgrind.

Teklad:
@Juban1 & @danieleds , care to give this latest commit a test?  I want to get this merged pretty soon so I can hunt down those memory leaks. :P

JuBan1:
Ah yes, there you go:

File | Result | Time | Previous Time 
----|----|----|---
long-line-100KB | 145 hits in 1 files | 20ms | 0ms
long-line-5MB | 7195 hits in 1 files | 80ms | 20ms
long-line-20MB | 28777 hits in 1 files | 300ms | 80ms
multiple-lines-100KB | 145 hits in 1 files | 10ms | 10ms
multiple-lines-5MB | 7195 hits in 1 files | 100ms | 340ms
multiple-lines-20MB | 28777 hits in 1 files | 300ms | 1900ms
binary-file-20MB | 156250 hits in 1 files | 600ms | 9500ms

Teklad:
Good deal.  I can live with these results.  As far as the memory leak goes... looks like DocEngine is good, just kinda how it works or something another.

I can't figure out the connection between the results being displayed and the memory not being returned when the list is cleared.

danieleds:
Sorry I've been busy all day. Fast results confirmed. Impressive!

@Teklad About the memory leak, does it happen in the master branch too?

Teklad:
@danieleds , yes, but you'd have to wait a long long time for it to show 
in the master branch due to the extremely slow search on binary files.

Its almost like the QStandardItem objects aren't getting deleted 
properly... but I don't think that's the case here.  I can definitely 
confirm the memory is allocated AFTER the search is finished.... not 
during the search.

Teklad:
Actually, to be more correct... its more like the search result data itself isn't cleared out of memory, but the QStandardItem is.

Teklad:
I was curious so I did this
```cpp
void FileSearchResultsWidget::addSearchResult(const FileSearchResult::SearchResult &searchResult)
{
    return;

```

The memory leak appears to disappear... so its gotta be somewhere in this file... or a related file.

Teklad:
Try doing a search on "binary-file-20MB" about 15 times over and keep "top" or a similar program open to watch the memory... then clear the results.  

Something isn't clearing when the results are emptied.

Sorry for all the replies, this issue has been frustrating trying to track down the exact cause.

JuBan1:
Feels like this must be the QStrings that contain the result strings not being cleaned up properly. They're ref counted internally, so if an orphan reference of them stays behind somewhere, the data stays in memory. 
I can't see where/why this could be happening though.

danieleds:
One thing that I observed is that if I do a search on "binary-file-20MB" and then I clear the results before doing any additional search, I can go on forever with these two steps (search, clear, search, clear, ...) and the memory usage stays stable.

If instead I clean after a lot of searches, then that memory is gone forever.

Teklad:
So I've made quite a bit of headway with this.  Apparently the reason it was choking so badly was partially due to the way we were setting up the threads.  I added a SearchInFilesController class which helps separate the GUI from the backend stuff.

This gave me a chance to clean up the code quite a bit for the searchInFiles function in frmsearchreplace.cpp.... I still need to do the same to the replaceInFiles I think... just to make it more tidy.

Now, the leak is better.... as-in... after quite a bit of testing I found out our threads weren't cleaning up completely like they should.  The changes look something like this... its a bit messy at the moment and probably needs a few changes, but it works well.

```cpp
//searchinfilesworker.cpp
SearchInFilesController::SearchInFilesController(QObject* parent, const QString &string, const QString &path, const QStringList &filters, const SearchHelpers::SearchMode &searchMode, const SearchHelpers::SearchOptions &searchOptions)
{
    SearchInFilesWorker *worker = new SearchInFilesWorker(string, path, filters, searchMode, searchOptions);
    worker->moveToThread(&workerThread);
    connect(&workerThread, SIGNAL(started()),worker, SLOT(run()));
    connect(worker, SIGNAL(error(QString)), parent, SLOT(searchInFilesUpdate(QString)));
    connect(worker, SIGNAL(errorReadingFile(const QString&, int&)), parent, SLOT(displayThreadErrorMessageBox(const QString&, int&)), Qt::BlockingQueuedConnection);
    connect(worker, SIGNAL(progress(QString)), parent, SLOT(searchInFilesUpdate(QString)));
    connect(&workerThread,SIGNAL(finished()),worker,SLOT(deleteLater()));
    connect(worker, SIGNAL(finished(bool, FileSearchResult::SearchResult)), parent, SLOT(searchInFilesFinished(bool, FileSearchResult::SearchResult)));
    connect(parent, SIGNAL(stopSearch()),worker,SLOT(stop()));
    connect(parent, SIGNAL(stopSearch()),this,SLOT(deleteLater()));
    workerThread.start();
}
```
```cpp
//frmsearchreplace.cpp
void frmSearchReplace::searchInFiles(const QString &string, const QString &path, const QStringList &filters, const SearchHelpers::SearchMode &searchMode, const SearchHelpers::SearchOptions &searchOptions)
{
//    cleanFindInFilesPtrs();

        m_session = new SearchInFilesSession(this);

        //A signal fires everytime msgBox is closed which will clean it up properly.
        m_session->msgBox = new dlgSearching(this);
        m_session->msgBox->setTitle(tr("Searching..."));
        m_session->msgBox->setWindowTitle(m_session->msgBox->title());

        SearchInFilesController* searchControl = new SearchInFilesController(this, string, path, filters, searchMode, searchOptions);
        m_session->msgBox->exec();

        // If we're here, the search finished or the user wants to cancel it.
        emit stopSearch();
}

```
As you noticed, the memory increases and stays up when multiple results are displayed, regardless of whether they are cleared or not.  I'm still not sure what causes this half of the issue... any insight on that part of the issue would be immensely helpful.

Teklad:
Okay... so I don't think its leaking anymore... just holding onto memory it allocated.  I can do 10 searches and clear the list... and the memory won't go up anymore unless I do 11 searches.

I'm not really sure whether there is a fix for this specific behavior or not, but its a heck of a lot better than the leaky mess it was before, lol.

danieleds:
Is the same thing needed for replaceInFiles?

Teklad:
Can you merge based on a previous commit?  If not I'll git revert and 
send the previous one. :)

danieleds:
It's better if you revert the two latest commits

Teklad:
All right. When I get back tomorrow I will move the last 2 commits to a
different branch and finalize this one.  The next request will focus on
code cleanup and optimization.

Teklad:
Alright, so I removed the two commits and pushed all the previous changes.

PR #252, opened 2016-08-31
NqqSettings to replace QSettings

JuBan1:
I thought of NqqSettings as a replacement for QSettings where:

* only one instance exists, just like the Notepadqq class.
* settings are type-safe.
* the details of each setting (save string, default value, etc) are all centralized and easy to see/change.
* we don't have to care about providing default values with every settings->value() call.

How it currently works:

* New settings are added to the nqqsettings.h files as a macro: `NQQ_SETTING(SettingName, SettingType, DefaultValue)`. The compiler then auto-generates a few functions: `getSetting`, `setSetting` and `resetSetting`. 
* A few things such as language settings or shortcuts are a bit more complicated because there are so many individual keys that are saved. They need to be implemented by hand.

Currently I've replaced most QSetting calls in `mainwindow.cpp` and transitioned the whole shortcut stuff to NqqSettings as a proof of concept.

The direct impact on code:

* Creating a new Settings objects is now done with `NqqSettings& settings = NqqSettings::getInstance()` or similar.
* Fetching settings went from `settings->value("WordWrap", false).toBool()` to `settings->getWordWrap()`.
* Setting settings went from `settings->setValue("WordWrap, true)` to `settings->setWordWrap(true)`.

As I said, there are some ugly workarounds in the code simply so I could get it to compile and run properly. Once we have the details worked out these bits will look cleaner.

### DIFF
--- a/src/ui/Search/frmsearchreplace.cpp
+++ b/src/ui/Search/frmsearchreplace.cpp
@@ -3,7 +3,7 @@
 #include "ui_frmsearchreplace.h"
 #include <QLineEdit>
 #include <QMessageBox>
-#include <QSettings>
+#include "include/nqqsettings.h"
 #include <QFileDialog>
 #include <QThread>
 #include <QCompleter>
@@ -25,14 +25,15 @@ frmSearchReplace::frmSearchReplace(TopEditorContainer *topEditorContainer, QWidg
     ui->cmbSearch->completer()->setCaseSensitivity(Qt::CaseSensitive);
     ui->cmbReplace->completer()->setCaseSensitivity(Qt::CaseSensitive);
 
-    QSettings s;
-    ui->cmbSearch->addItems(s.value("Search/searchHistory", QStringList()).toStringList());
+    NqqSettings& s = NqqSettings::getInstance();
+
+    ui->cmbSearch->addItems(s.Search.getSearchHistory());
     ui->cmbSearch->setCurrentText("");
-    ui->cmbReplace->addItems(s.value("Search/replaceHistory", QStringList()).toStringList());
+    ui->cmbReplace->addItems(s.Search.getReplaceHistory());
     ui->cmbReplace->setCurrentText("");
-    ui->cmbLookIn->addItems(s.value("Search/fileHistory", QStringList()).toStringList());
+    ui->cmbLookIn->addItems(s.Search.getFileHistory());
     ui->cmbLookIn->setCurrentText("");
-    ui->cmbFilter->addItems(s.value("Search/filterHistory", QStringList()).toStringList());
+    ui->cmbFilter->addItems(s.Search.getFilterHistory());
     ui->cmbFilter->setCurrentText("");
 
     connect(ui->cmbSearch->lineEdit(), &QLineEdit::textEdited, this, &frmSearchReplace::on_searchStringEdited);
@@ -657,9 +658,9 @@ void frmSearchReplace::on_radSearchWithSpecialChars_toggled(bool checked)
 
 void frmSearchReplace::on_searchStringEdited(const QString &/*text*/)
 {
-    QSettings s;
+    NqqSettings& s = NqqSettings::getInstance();
 
-    if (s.value("Search/SearchAsIType", true).toBool()) {
+    if (s.Search.getSearchAsIType()) {
         if (ui->actionFind->isChecked()) {
             Editor *editor = currentEditor();
 
@@ -729,36 +730,49 @@ QStringList frmSearchReplace::fileFiltersFromUI()
     return filters;
 }
 
-void frmSearchReplace::addToHistory(QString string, QString type, QComboBox *comboBox)
-{
-    if (string != "") {
-        QSettings s;
-        QStringList history = s.value("Search/" + type + "History", QStringList()).toStringList();
-        history.prepend(string);
-        history.removeDuplicates();
-        history = history.mid(0, 10);
-        s.setValue("Search/" + type + "History", history);
-        comboBox->clear();
-        comboBox->addItems(history);
-    }
+void addToHistory(QStringList& history, QString string, QComboBox *comboBox) {
+    if(string.isEmpty())
+        return;
+
+    history.prepend(string);
+    history.removeDuplicates();
+    history = history.mid(0, 10);
+    comboBox->clear();
+    comboBox->addItems(history);
 }
 
 void frmSearchReplace::addToSearchHistory(QString string)
 {
-    addToHistory(string, "search", ui->cmbSearch);
+    NqqSettings& s = NqqSettings::getInstance();
+
+    auto history = s.Search.getSearchHistory();
+    addToHistory(history, string, ui->cmbSearch);
+    s.Search.setSearchHistory(history);
 }
 
 void frmSearchReplace::addToReplaceHistory(QString string)
 {
-    addToHistory(string, "replace", ui->cmbReplace);
+    NqqSettings& s = NqqSettings::getInstance();
+
+    auto history = s.Search.getReplaceHistory();
+    addToHistory(history, string, ui->cmbSearch);
+    s.Search.setReplaceHistory(history);
 }
 
 void frmSearchReplace::addToFileHistory(QString string)
 {
-    addToHistory(string, "file", ui->cmbLookIn);
+    NqqSettings& s = NqqSettings::getInstance();
+
+    auto history = s.Search.getFileHistory();
+    addToHistory(history, string, ui->cmbSearch);
+    s.Search.setFileHistory(history);
 }
 
 void frmSearchReplace::addToFilterHistory(QString string)
 {
-    addToHistory(string, "filter", ui->cmbFilter);
+    NqqSettings& s = NqqSettings::getInstance();
+
+    auto history = s.Search.getFilterHistory();
+    addToHistory(history, string, ui->cmbSearch);
+    s.Search.setFilterHistory(history);
 }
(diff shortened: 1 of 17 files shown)


danieleds:
Nice! I can't see any issue in the general architecture. Continue in this way.

For categories, do you think we'll be able to provide some kind of hierarchical structure to the code?
For example, if I have a "Title" setting under CategoryA, and another "Title" setting under CategoryB, it would be nice to be able to do something like `settings->CategoryA->getTitle()` and `settings->CategoryB->getTitle()`.


Also, as you are changing the first letter of some setting from lowercase to uppercase (which is good), let's not forget to maintain compatibility by reading the old settings names as a fallback.

Teklad:
Well it probably wouldn't hurt to ignore case when pulling a setting. Or
doing a conversion of the settings file on initial run depending on whether
Version is less than blah.

JuBan1:
> For categories, do you think we'll be able to provide some kind of hierarchical structure to the code?

Good idea. Should be no problem.

QSettings should already ignore case. From the docs:

> Note that the Windows registry and INI files use case-insensitive keys, whereas the CFPreferences API on OS X and iOS uses case-sensitive keys.

JuBan1:
```
    //Settings of the "General" category
    NQQ_SETTING(LastSelectedDir,      QString,    ".")
    NQQ_SETTING(RecentDocuments,      QList<QVariant>, QList<QVariant>())

    //Settings of the "Appearance" category
    BEGIN_CATEGORY(Appearance)
        NQQ_SETTING(OverrideFontFamily, QString,    "")
        NQQ_SETTING(OverrideFontSize,   int,        0)
    END_CATEGORY(Appearance)
```

In code:
```

    editor->setFont(m_settings->Appearance.getOverrideFontFamily(),
                    m_settings->Appearance.getOverrideFontSize());
```

Teklad:
Perfect. No problem then. I'm so used to the Linux file system I assume
everything is case sensitive. Lol.

danieleds:
Perfect, let's wrap the general stuff into their own category too.

> I'm so used to the Linux file system I assume everything is case sensitive. Lol.

Yeah me too :)

JuBan1:
Right, so apparently the Qt documentation is a dirty liar. The keys *are* case-sensitive, clearly evidenced by the fact that I now have duplicates with different case of most keys in my .ini. Guess I'll cook up some compatibility code.

On that note, it might be a good idea to wait with further merges until this PR is ready since it modifies many files and is going probably going to lead to a bunch of merge conflicts.

I'm probably done with this tomorrow. The NqqSettings class is basically done now and all settings are migrated over to it. But since the new system is a little more rigid I had to change some parts of the frmpreferences files and I'll have to thoroughly test them. Also, preserving backwards compatibility.

Teklad:
Git rebase is your friend if conflicts do arise, but I doubt they 
will... I'm working on a completely different part that's kind of 
disconnected from the rest of notepadqq. :X

JuBan1:
That's true. The search/replace GUI has a whole bunch of QSettings calls in it but there are none under the hood.

JuBan1:
Alright, that should be all.

Because QSettings was used somewhat creatively in a few different places I had to rewrite them. I took the opportunity to clean up a lot of the frmpreferences code since I had to half do that anyways to get NqqSettings to work well with it.

So ultimately quite some code has changed. As usual I hunted down all bugs I can find but due to the extensive nature of the changes some extra testing is definitely needed. Especially the Languages and Shortcuts parts have changed significantly.

There is now a function (`NqqSettings::ensureBackwardsCompatibility()`) that is called at program start. It looks at the settings file to determine whether the keys still have old capitalization and replaces them with the properly capitalized keys.

src/ui/Search/frmsearchreplace.cpp
```diff
@@ -729,36 +730,49 @@ QStringList frmSearchReplace::fileFiltersFromUI()
...
-        QSettings s;
-        QStringList history = s.value("Search/" + type + "History", QStringList()).toStringList();
-        history.prepend(string);
-        history.removeDuplicates();
-        history = history.mid(0, 10);
-        s.setValue("Search/" + type + "History", history);
-        comboBox->clear();
-        comboBox->addItems(history);
-    }
+void addToHistory(QStringList& history, QString string, QComboBox *comboBox) {
```
danieleds: why is this not declared in frmsearchreplace.h?

JuBan1: Because it didn't really need to. It's pretty much just a small helper so I don't have to write the same piece of code 3x. It doesn't access any of `frmSearchReplace`'s data nor is it going to be of any interesting to any code outside of it.

danieleds:
Hi! I'm testing.

One thing I noticed is that shortcuts are broken. For example, if you start with an empty config file, CTRL+F doesn't work. I guess the defaults are wrong.

JuBan1:
You're right. Silly me, I forgot to check whether the key exists before setting it. That way, *all* defaults are overridden by empty key sequences. Oops.

src/ui/Search/frmsearchreplace.cpp
danieleds: Ok, now I see, you had to refactor the method because you couldn't pass a "setting name" as before. :+1: 

Can you add some docs over this method?

src/ui/frmpreferences.cpp
```diff
@@ -289,42 +252,63 @@ void frmPreferences::loadTranslations(QSettings *s)
...
-        QString langId = map.value("id", "");
-        QString keyPrefix = "Languages/" + langId + "/";
 
-        s->setValue(keyPrefix + "useDefaultSettings", m_langsTempSettings->value(keyPrefix + "useDefaultSettings"));
+    //Build the interface
+    QWidget *container = ui->page_5;
+    QVBoxLayout *layout = new QVBoxLayout();
+    QPushButton *resetDefaults = new QPushButton("Restore Defaults");
+    QObject::connect(resetDefaults,SIGNAL(clicked()),this,SLOT(resetShortcuts()));
+    resetDefaults->setFixedWidth(128);
```
danieleds: For convention, we use [this](https://github.com/notepadqq/notepadqq/blob/master/src/ui/mainwindow.cpp#L54) style for connecting signals.

src/ui/frmpreferences.ui
```diff
@@ -463,8 +464,8 @@
              <rect>
               <x>0</x>
               <y>0</y>
-              <width>403</width>
-              <height>196</height>
+              <width>395</width>
+              <height>205</height>
              </rect>
             </property>
             <layout class="QVBoxLayout" name="verticalLayout_3">
```
danieleds: this whole file shouldn't have changed :)

src/ui/frmpreferences.cpp
JuBan1: This is actually code that was already in there. You'll find it again in the file a little further up, where it was "deleted".
I actually would like the entire block to go away. All it does is create the contents of the Shortcuts page, but the layout and button could just be added in the designer instead. I just couldn't figure out how to. The StackedWidget is really stubborn in the designer. I think that's what the original author of the code thought too, which is why they just added the widgets this way.

Anyways, I'll change the connect call.

src/ui/nqqsettings.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+void NqqSettings::ensureBackwardsCompatibility()
+{
+    QSettings s;
+
+    const QStringList keys = s.allKeys();
+
+    //If this string is in the settings we either have the new ini settings
+    if(keys.contains("BackwardsCompatible"))
+            return;
```
danieleds: I think we should take this opportunity to add, within the settings, a special key which represents the version of Notepadqq that wrote that file. The value should be updated each time Nqq is started.

So, instead of `keys.contains("BackwardsCompatible")`, we will check if `NotepadqqVersion` key is present or not.

I think it will come useful in the future, for situations like this ;)

src/ui/frmpreferences.ui
JuBan1: There's a new Widget in line 439. The page widget for the shortcuts page, called page_5.

But the other changed tags? Yea, they sometimes sneak in :x. I don't really know what prompted the designer to subtly change the height..

src/ui/Search/frmsearchreplace.cpp
JuBan1: Will do.

src/ui/frmpreferences.cpp
danieleds: If you can find a way to do it from the designer, the change is more than welcome :D
It's not that important though, not worth losing much time.

src/ui/frmpreferences.ui
danieleds: Ah! Sorry, I had missed it.

By the way, now that you made me focus on that file I notice that the names of those pages suck a bit. Could you for now change `page_5` to something like `pageShortcuts` so that, even if from the designer it's a blank page, there is some clue about what it does?

Only for `page_5`, so that this pull request stays clean and we can address the others later.

danieleds:
Default shortcuts now working :)

src/ui/frmpreferences.ui
danieleds: And yeah, Qt Designer can often be a dick

danieleds:
A `s.setValue("BackwardsCompatible", true);` is sneaking at the end of nqqsettings.cpp

danieleds:
Wow, that should be all.

I'll keep testing this branch for a little while, but I believe it's fine, and I really like the work.

Hope to merge soon! :)

JuBan1:
Good to hear.

By the way, since the version number of this branch is still 0.53.0, the backwards compat function will trigger every time. But as soon as everything's merged and you bump that number up a notch it'll behave.

danieleds:
Merged as 9e14f05d7c9b23741ecc15f1a4ced22377999e97

Issue #253, opened 2016-09-03
Smart Indent


slypiggies:
Hello!
Is there a way to turn off _Smart Indent_? For example, when I am editing a C++ file, I type `{` and hit `Enter`, an extra indentation will be added. Also, when I missed the `;` on the previous line and hit `Enter`, an extra indentation will be added, too. I just want to keep the indentation on the previous line (_Basic Auto Indent_). Where can I find a toggle to turn this (_Smart Indent_) off?

### Suggestion (if there isn't a way)
Add a way to allow users to turn off _Smart Indent_, but keeping _Basic Auto Indent_ at the same time.

Thanks!

JuBan1:
Hey, 
right now there's no option for this as far as I know. It's been requested in #170 as well. It seems to be a very easy addition so I'll throw a patch together.

slypiggies:
Thanks a lot!

danieleds:
Fixed by @JuBan1 at 8210b15a1dd73e0c05e9e552e24fa92b3d86e6eb

Issue #170, opened 2016-01-09
How to turn off auto indent in notepadqq?


EastAPOLO:
Whenever I go to a new line it automatically indents for me and I was wondering if there is any way to turn this off, because I can't find it anywhere in the settings.

Thank you,

- EastAPOLO

danieleds:
No, right now there is not any setting for this.

xproofx:
This would be a very welcome feature.  Thanks!

danieleds:
Fixed by @JuBan1 at 8210b15a1dd73e0c05e9e552e24fa92b3d86e6eb

PR #254, opened 2016-09-03
Allow smart indentation to be enabled or disabled by the user.

JuBan1:
Requested in #253, #170.

Adds a new menu option `Edit->Indentation->Enable Smart Identation`. It is enabled by default. 

If disabled, codemirror will fall back to the "prev"-style of indentation and disregard any smart indent coming from the language modes.

Right now the setting won't be saved, but will be set to true every time Nqq is opened. I think this is a sensible default.

This is a per-window setting right now. I'm not sure if this is the best choice but we're a bit inconsistent with this anyways. But that's a story of another Issue.

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -417,6 +417,12 @@ UiDriver.registerEventHandler("C_CMD_SET_OVERWRITE", function(msg, data, prevRet
     editor.toggleOverwrite(data);
 });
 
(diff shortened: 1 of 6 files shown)


danieleds:
Thanks, merged as 8210b15a1dd73e0c05e9e552e24fa92b3d86e6eb

PR #256, opened 2016-09-04
Search overhaul

Teklad:
All right, so this is a recreation of my last pull request since I messed up a bit when reverting.  This includes all the changes prior to the threading improvements.  This is much easier to integrate for the time being.

### DIFF
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -53,6 +53,7 @@ void SearchInFilesWorker::run()
     //int totalFileMatches = 0;
     // Number of files that contain matches
     //int totalFiles = 0;
+    bool multiLine = m_regex.pattern().contains(QStringLiteral("\\n"));
 
     while (it.hasNext()) {
         m_stopMutex.lock();
@@ -70,7 +71,6 @@ void SearchInFilesWorker::run()
         QFile f(fileName);
         DocEngine::DecodedText decodedText;
         bool retry;
-        bool multiLine = m_regex.pattern().contains(QStringLiteral("\\n"));
         do {
             retry = false;
             decodedText = DocEngine::readToString(&f);
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -196,10 +196,10 @@ FileSearchResult::FileResult SearchInFilesWorker::searchMultiLineRegExp(const QS
         for(int i=0; i < lineStart.length(); i++){
             if(lineStart[i] > column) {
                 line = i-1;
+                fileResult.results.append(buildResult(line, column - lineStart[line], column, content, match.capturedLength()));
                 break;
             }
         }
-        fileResult.results.append(buildResult(line, column, column, content, match.capturedLength()));
         match = m_regex.match(content, column + match.capturedLength());
         column = match.capturedStart();
     }
@@ -220,14 +220,16 @@ FileSearchResult::SearchResult SearchInFilesWorker::getResult()
 FileSearchResult::Result SearchInFilesWorker::buildResult(const int &line, const int &column, const int &absoluteColumn, const QString &lineContent, const int &matchLen)
 {
     FileSearchResult::Result res;
+    bool multiLine = lineContent.contains('\n') || lineContent.contains('\r');
+    int displayColumn = multiLine ? absoluteColumn : column;
 
-    if (column > 50) {
-        res.previewBeforeMatch = lineContent.mid(column-50,50);
+    if (displayColumn > 50) {
+        res.previewBeforeMatch = lineContent.mid(displayColumn-50,50);
     }else {
-        res.previewBeforeMatch = lineContent.mid(0,column);
+        res.previewBeforeMatch = lineContent.mid(0,displayColumn);
     }
-    res.match = lineContent.mid(column,matchLen);
-    res.previewAfterMatch = lineContent.mid(column + matchLen,matchLen + 50);
+    res.match = lineContent.mid(displayColumn,matchLen);
+    res.previewAfterMatch = lineContent.mid(displayColumn + matchLen,matchLen + 50);
     res.matchStartLine = line;
     res.matchStartCol = column;
     res.matchEndLine = line;
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -169,15 +169,13 @@ FileSearchResult::FileResult SearchInFilesWorker::searchSingleLine(const QString
 
 }
 
-FileSearchResult::FileResult SearchInFilesWorker::searchMultiLineRegExp(const QString &fileName, const QString &content)
+FileSearchResult::FileResult SearchInFilesWorker::searchMultiLineRegExp(const QString &fileName, QString &content)
 {
     FileSearchResult::FileResult fileResult;
     fileResult.fileName = fileName;
     
     int column = 0;
     int line = 0;
-    int lastLine = 0;
-    QString fullDoc;
     QVector<int> lineStart;
     QRegularExpressionMatch match;
 
@@ -193,7 +191,7 @@ FileSearchResult::FileResult SearchInFilesWorker::searchMultiLineRegExp(const QS
     }
 
     m_regex.setPatternOptions(m_regex.patternOptions() | QRegularExpression::MultilineOption);
-    match = m_regex.match(fullDoc);
+    match = m_regex.match(content);
     column = match.capturedStart();
     while(column != -1 && match.hasMatch()) {
         //Get the current line we are working on.
@@ -203,8 +201,8 @@ FileSearchResult::FileResult SearchInFilesWorker::searchMultiLineRegExp(const QS
                 break;
             }
         }
-        fileResult.results.append(buildResult(line, column, column, fullDoc, match.capturedLength()));
-        match = m_regex.match(fullDoc,column + match.capturedLength());
+        fileResult.results.append(buildResult(line, column, column, content, match.capturedLength()));
+        match = m_regex.match(content, column + match.capturedLength());
         column = match.capturedStart();
     }
     
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -4,7 +4,7 @@
 #include <QDirIterator>
 #include <QMessageBox>
 #include <QThread>
-
+#include <QIODevice>
 SearchInFilesWorker::SearchInFilesWorker(const QString &string, const QString &path, const QStringList &filters, const SearchHelpers::SearchMode &searchMode, const SearchHelpers::SearchOptions &searchOptions)
     : m_string(string),
       m_path(path),
@@ -133,10 +133,8 @@ FileSearchResult::FileResult SearchInFilesWorker::searchSingleLineRegExp(const Q
     int i = 0;
     int column; 
     int absoluteColumn = 0;
-    int eolSize = 1;
     QRegularExpressionMatch match;
 
-    if (content.indexOf("\r\n") != -1) eolSize = 2;
     while (!(line=stream.readLine()).isNull()) {
         if (m_stop) break;
         match = m_regex.match(line);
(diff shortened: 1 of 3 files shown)


danieleds:
Old PR ref: #250

danieleds:
Tip: it's ok to force-push changes to a feature branch used in a pull request

danieleds:
Looks like "Match whole word only" flag is broken?

Teklad:
Yea... I never re-implemented it properly.  I pushed a new commit which includes the function.  Test again and make sure it's behaving as expected.

danieleds:
Just some notes:

Currently, the `master` branch considers `^` as the start of the *file* and `$` as the end of the *file*.
Your branch, instead, considers them as respectively the start and the end of the *line*, which I think is better. But if I throw a `\n` in there, it suddenly starts to behave like master, which to me is confusing. But I would like some opinions (@JuBan1 ?).

In addition, searching for `\n` is not a safe way to detect if a search is multiline or not.
E.g. I could put an escape sequence in there: if you search `interdum\.\x0a` as a regex over `multiple-lines-100KB`, you'll notice that the master branch works ok but this branch finds zero results because it thinks that the regex is single-line when actually it is multiline (`\x0a` is equivalent to `\n`).

src/ui/Search/searchinfilesworker.cpp
```diff
@@ -143,71 +96,158 @@ void SearchInFilesWorker::stop()
...
-    // String composed by all the lines that contain the found word.
-    QString previewString = content->mid(firstLinePosStart, lastLinePosEnd - firstLinePosStart);
+}
 
-    // All the lines in wholeLine
-    QStringList matchLines = previewString.split(newLine, QString::KeepEmptyParts);
+FileSearchResult::FileResult SearchInFilesWorker::searchMultiLineRegExp(const QString &fileName, QString content)
+{
+    FileSearchResult::FileResult structFileResult;
+    structFileResult.fileName = fileName;
```
danieleds: Why this whole work? Wasn't the flag `QRegularExpression::MultilineOption` sufficient for multiline regexps?

Teklad: searchMultiLineRegex is still rough around the edges...  Its not been 
touched much since the initial implementation.  If you like, you could 
simply throw the old slow method in there for the moment... otherwise 
it'll take me some time to finish.

On that note, I believe we have an option specifically in 
frmSearchReplace.cpp for multiLineRegex....  Would it be okay to hook it 
to that instead to detect newline regex?

JuBan1:
I think `^` and `$` really only make sense as line beginning and ending, in all cases. File begin/end is rarely very useful.

src/ui/Search/searchinfilesworker.cpp
danieleds: >  If you like, you could simply throw the old slow method in there for the moment

God no it sucks :D

> On that note, I believe we have an option specifically in frmSearchReplace.cpp for multiLineRegex.... Would it be okay to hook it to that instead to detect newline regex?

Hmm, what option are you referring to?

Teklad: Nevermind,
I misread the special characters option as regex w/ newlines.... could 
we add a multiLine checkbox to the frm?  I'm not exactly very happy with 
the way multiLineRegExp is done at the moment... it was ugly like that 
when I found it in kate and I never did find time to pick it apart much.

I could probably do a much better job.... but I'm wanting to get the 
singleLineRegExp pushed so I can grab the latest changes from master and 
do the thread improvements with the next pull request (All the 
Nqqsettings changes cause massive amounts of merge conflicts outside of 
the files I'm in.).

multiLineRegex isn't going to be massively used for text document 
searches anyways, so you could disable it until I can finish my next set 
of patches since its focused on improving these features further.

danieleds:
> I could probably do a much better job.... but I'm wanting to get the singleLineRegExp pushed so I can grab the latest changes from master and do the thread improvements with the next pull request (All the Nqqsettings changes cause massive amounts of merge conflicts outside of the files I'm in.). multiLineRegex isn't going to be massively used for text document searches anyways, so you could disable it until I can finish my next set of patches since its focused on improving these features further.

Do the thread improvements depend on these changes? Because there are still some issues with multiline even in plain text mode, and I would like to avoid merging regressions even if they will be corrected later.

So if the thread improvements do not depend on this feature, we can merge them first so that we get rid of those merge conflicts.

Teklad:
The issue is I can't mess with any other files but the two that I've 
been changing due to all the changes from nqqsettings.

I suppose I could write some pseudo code for a multiline option in 
frmpreferences.cpp, but it would either need to be re-implemented or the 
merge conflicts resolved when the pull request is finished.

danieleds:
Ok I think I get what you're saying.

1. are we sure we need a multiline option in frmpreferences.cpp? Maybe we can do better at autodetecting it.

2. can't you rebase this branch on top of master so that you get the latest changes and no conflicts occur?

    git checkout search-overhaul
    git rebase master
    git push -f origin search-overhaul

JuBan1:
> The issue is I can't mess with any other files but the two that I've been changing due to all the changes from nqqsettings.

I had a feeling something like this would happen, if you remember :P

But you should be able to rebase without much trouble since I haven't messed with the actual search classes much/at all.

Teklad:
True, I didn't think about it but I could rebase before making 
changes... I'm an idiot. lol

Teklad:
Okay, so I re-implemented the multiLineRegex.... its not as fast as the other methods, mainly due to how regex works in QT.  

It should be able to get the job done without any regressions though.  Its still faster than the original implementation we had.

Teklad:
Also, travis doesn't appear to be aware of QRegularExpression's optimize function... so it keeps failing my builds. xD

danieleds:
Okay so looks like Travis is already bearing fruit ;-)
"Optimize" is only available since Qt 5.4. Maybe you can wrap it in a conditional macro.

Teklad:
Alright, just removed the optimize thing... since QT auto-optimizes anyways after so many passes on the fly.

Build passed. Just needs some more testing now.

Teklad:
So I experimented with std regex for a bit.... and it yields much better 
results than QRegularExpression in terms of efficiency. Would you be 
against moving over to the C++11 standard regex library for taking care 
of this particular issue?

danieleds:
I'm perfectly fine with switching to C++11 regex implementation if it's significantly better in terms of performance

Teklad:
Well that didn't go as well as I'd hoped... std::regex isn't very 
forgiving for user-inputted strings... the error handling it has is 
along the lines of segfaulting when you input something really funky, or 
anything that simply gets too large of a result. -_-

Teklad:
So apparently, the issue is a known "bug" in the std::regex implementation.  The documentation is listed: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61582.  

For the time being we will be stuck using QRegularExpression for the regex half of the matching, mainly because I don't think its the right decision at this time to use external libraries for something so trivial.

I can't think of anything else I need to do for this pull request unless there's more bugs I haven't squashed.

Teklad:
Alright.  I'm a lot happier with this setup...  much more fluid and manage-able in the long run.

The plain text searching got another speed boost with this commit... While also allowing QRegularExpression to maintain its proper functionality for the ^ and $ thingies.

I re-implemented the "special characters" search function for the plaintext side of things so it didn't dip into the regex stuff when it got turned on.

It works by creating a list of line numbers rather than iterating through the text via getline.... and uses the data to create the results as it goes along.

The RegEx searching is decently fast... still slow on binaries, but meh.  Its biggest issue appears to be when it hits tons of results it slows down a bit.

Something we can do further down the line to improve speed a bit more would be to either push results directly to the GUI, or build the results outside of the main thread.

I need this to be thoroughly tested for defects. :)

danieleds:
It's way better now!

src/ui/Search/searchinfilesworker.cpp
```diff
@@ -22,53 +22,43 @@ SearchInFilesWorker::~SearchInFilesWorker()
 
 void SearchInFilesWorker::run()
 {
-    // Search string converted to a regex
-    QString rawSearch = frmSearchReplace::rawSearchString(m_string, m_searchMode, m_searchOptions);
+    FileSearchResult::SearchResult searchResult;
+    searchResult.search = m_string;
 
     QFlags<QRegularExpression::PatternOption> options = QRegularExpression::NoPatternOption;
```
danieleds: The multiline regex option was actually a good thing. Change line 28 to this:
        
    QFlags<QRegularExpression::PatternOption> options = QRegularExpression::MultilineOption;

src/ui/Search/searchinfilesworker.cpp
```diff
@@ -22,53 +22,43 @@ SearchInFilesWorker::~SearchInFilesWorker()
...
+    if (m_searchMode == SearchHelpers::SearchMode::Regex) {
+        if (m_searchOptions.MatchCase == false) {
+            options |= QRegularExpression::CaseInsensitiveOption;
+        }
+        QString rawSearch = frmSearchReplace::rawSearchString(m_string, m_searchMode, m_searchOptions);
+        m_regex.setPattern(rawSearch);
+        m_regex.setPatternOptions(options);
+    }else if (m_searchMode == SearchHelpers::SearchMode::SpecialChars) {
+        QString rawSearch = frmSearchReplace::rawSearchString(m_string, m_searchMode, m_searchOptions);
+        m_string = unescapeString(m_string);
```
danieleds: `rawSearch` is not used?

Teklad: The way the search functions work won't allow getting the rawSearch 
string outside of the scope of "run"... I suppose we could make 
rawSearch a class member, but at that rate it'd be better to convert 
m_string directly after we apply all the data.

I'll go ahead and add MultiLine option back.... want me to just do 
m_regex.setPattern( frmSearchReplace::rawSearchString(m_string, 
m_searchMode, m_searchOptions)); instead?

danieleds: > want me to just do m_regex.setPattern( frmSearchReplace::rawSearchString(m_string, m_searchMode, m_searchOptions)); instead?

No, it's fine, just change the initialization at line 28 like I showed.

danieleds:
When I search `\x4corem` in `multiple-lines-100KB` with special characters mode, it should search for `Lorem` but it doesn't find anything.

I don't get why; the code looks fine.

Teklad:
I'll look into it... its probably an issue with the data increment.... 
more than likely I have it skipping too far ahead.

src/ui/Search/searchinfilesworker.cpp
danieleds: > The way the search functions work won't allow getting the rawSearch string outside of the scope of "run"... I suppose we could make rawSearch a class member, but at that rate it'd be better to convert m_string directly after we apply all the data.

Hmm I don't understand, so line 39 is a mistake and can be removed?

Teklad: It was just leftover from the old code... its not really used anymore 
outside of the scope you see there.... not necessary to keep it at all 
really.

Teklad: Just pushed a new commit.... care to try that \x4corem test again? 
Pretty sure this should fix it.

danieleds: Yep, fixed! :)

Teklad: Barring any other issues. I think this is finally ready to be merged.

I learned a lot of nifty tricks during this little project.  Our search 
functions are actually faster than kate's now from testing. :)

Also,  I did an experimental test of threading the searches just to 
check.  In cases of SSD hard drives, it actually increases the 
performance even further.  It would be a nice little feature to have as 
optional later down the line, but this should be plenty fast for now.

danieleds:
A minor graphical issue with the search results:

When the whole matched string is in one single line, the result should show *only portions of that line*.
When the matched string spans multiple lines, instead, then it should behave as it is doing now.

Right now, instead, if for example in `multiple-lines-100KB` I search "Lorem ipsum", then I see the end of the previous lines too (which I don't care about).

Teklad:
So for results on a single line it just needs to show that line only?  
It shouldn't be that hard to accomplish, but I left it like that to keep 
performance high.

I'll see if I can detect the newline without impacting performance too 
much.  Should be able to do that within the code right before the 
results are displayed onscreen...  clipping previewBefore|AfterMatch 
should do the trick.

Teklad:
Okay, I think I got a fix for the graphical issue... care to reveal your 
test case that you used so I can make sure this little addition works 
for clipping preview data?

Teklad:
Nevermind... its definitely fixed... Give this commit a go and make sure it behaves better. :)

Teklad:
I tested this for a good while and am not really seeing anymore issues with
it.

:)

danieleds:
Me neither! I'll take a final look to the code later

danieleds:
You forgot to remove the unused `rawSearch` line.

Also, I'm now getting a crash caused by an assertion failure. You are probably reading beyond the index of something.

danieleds:
The issue has been introduced with d40fff4.

Anyway that work should not be done in `searchresultsitemdelegate.cpp`. It should be done in `SearchInFilesWorker::buildResult`: that's because it's where `previewBeforeMatch`/`previewAfterMatch` are supposed to be created.

danieleds:
I'm merging without those last 2 commits, then I'll see if I can push a quick fix.

danieleds:
Finally done. :)
Thank you!

PR #258, opened 2016-09-05
Replaced the shortcuts table with a tree.

JuBan1:
image

Since the number of menu actions grows slowly but steadily, reorganizing them into a tree is probably an easier way for users to find the actions they want to change. Other than that, functionality has not changed.

On that note - should the user get a MsgBox telling them to resolve key conflicts when they attempt to press Accept with conflicts present?

### DIFF
--- a/src/ui/frmpreferences.cpp
+++ b/src/ui/frmpreferences.cpp
@@ -94,8 +94,7 @@ void frmPreferences::on_buttonBox_accepted()
 {
     if(m_keyGrabber->hasConflicts()){
         //Try our best to show the error to the user immediately.
-        ui->stackedWidget->setCurrentWidget( ui->pageShortcuts );
-        ui->treeWidget->setCurrentItem( ui->treeWidget->topLevelItem( ui->stackedWidget->currentIndex() ) );
+        ui->treeWidget->setCurrentItem(ui->treeWidget->topLevelItem(ui->stackedWidget->indexOf(ui->pageShortcuts)));
         m_keyGrabber->scrollToConflict();
 
         QMessageBox msgBox;
--- a/src/ui/frmpreferences.cpp
+++ b/src/ui/frmpreferences.cpp
@@ -93,16 +93,16 @@ void frmPreferences::on_treeWidget_currentItemChanged(QTreeWidgetItem *current, 
 void frmPreferences::on_buttonBox_accepted()
 {
     if(m_keyGrabber->hasConflicts()){
+        //Try our best to show the error to the user immediately.
+        ui->treeWidget->setCurrentItem( ui->treeWidget->topLevelItem(4) );
+        m_keyGrabber->scrollToConflict();
+
         QMessageBox msgBox;
         msgBox.setWindowTitle(QCoreApplication::applicationName());
         msgBox.setIcon(QMessageBox::Warning);
         msgBox.setText("<h3>" + tr("Keyboard shortcut conflict") + "</h3>");
         msgBox.setInformativeText(tr("Two or more actions share the same shortcut. These conflicts must be resolved before your changes can be saved."));
         msgBox.exec();
-
-        ui->treeWidget->setCurrentItem( ui->treeWidget->topLevelItem(4) );
-        m_keyGrabber->scrollToConflict();
-
         return;
     }
 
--- a/src/ui/frmpreferences.cpp
+++ b/src/ui/frmpreferences.cpp
@@ -94,7 +94,8 @@ void frmPreferences::on_buttonBox_accepted()
 {
     if(m_keyGrabber->hasConflicts()){
         //Try our best to show the error to the user immediately.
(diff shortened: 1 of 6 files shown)


danieleds:
Nice! I'll take a look as soon as I get home.
@Teklad as the original author, do you want to double-check?

danieleds:
And yes, I think we should at least warn the user about unresolved conflicts that he might not have noticed because e.g. they were outside the scroll area.

src/ui/keygrabber.cpp
```diff
@@ -2,73 +2,86 @@
...
-    setSelectionBehavior(QAbstractItemView::SelectRows);
-    setEditTriggers(QAbstractItemView::NoEditTriggers);
-    setAlternatingRowColors(true);
     setColumnCount(2);
-    setHorizontalHeaderItem(0, new QTableWidgetItem("Action"));
-    setHorizontalHeaderItem(1, new QTableWidgetItem("Keyboard Shortcut"));
-    connect(this, &QTableWidget::itemChanged, this, &KeyGrabber::itemChanged);
+    setColumnWidth(0, 400);
+    setAlternatingRowColors(true);
+    setHeaderLabels(QStringList() << "Action" << "Keyboard Shortcut");
```
danieleds: `tr()` is missing for header labels

src/ui/frmpreferences.cpp
```diff
@@ -92,6 +92,15 @@ void frmPreferences::on_treeWidget_currentItemChanged(QTreeWidgetItem *current,
 
 void frmPreferences::on_buttonBox_accepted()
 {
+    if(m_keyGrabber->findConflicts()){
+        QMessageBox msgBox;
+        msgBox.setWindowTitle(QCoreApplication::applicationName());
+        msgBox.setIcon(QMessageBox::Warning);
+        msgBox.setInformativeText(tr("Two or more actions share the same shortcut. These conflicts must be resolved before your changes can be saved."));
+        msgBox.exec();
+        return;
```
danieleds: I'd add a 

    msgBox.setText("<h3>" + tr("Keyboard shortcut conflict") + "</h3>");

and redirect the user to the shortcuts tab *right before* showing the messagebox.

Teklad: It might be unwise to let them leave the preferences menu with 
conflicts, even optionally.... as it will break both of the shortcut 
keys in those cases due to how Qt works.  As a side note, auto-scrolling 
to the conflict would be a nice touch.

The original implementation was just to get conflict detection 
working... I never did get a chance to polish it very well. :)

danieleds: Auto-scrolling would be really nice!

JuBan1:
Ok, fixed the smaller things and added auto-scrolling to the conflict. 

It would be nice if the code tried to fit both conflicting shortcuts in the view but that doesn't seem worth the extra work considering there could be several different conflicts and/or more than two actions sharing the same key sequence.

src/ui/frmpreferences.cpp
```diff
@@ -92,10 +92,15 @@ void frmPreferences::on_treeWidget_currentItemChanged(QTreeWidgetItem *current,
 
 void frmPreferences::on_buttonBox_accepted()
 {
-    if(m_keyGrabber->findConflicts()){
+    if(m_keyGrabber->hasConflicts()){
+        //Try our best to show the error to the user immediately.
+        ui->treeWidget->setCurrentItem( ui->treeWidget->topLevelItem(4) );
```
danieleds: Use the ui reference instead of the index `4` and we're good to merge!

Teklad:
Well, if there's more than 2... it'll just scroll to the next conflict 
when they press accept.... a small indiciator text by the reset button 
would probably be fine to let them know there are still more conflicts 
remaining.... could also add a "next conflict" button in these cases as 
well.

src/ui/frmpreferences.cpp
JuBan1: What's the UI reference? I can't use ui->page_4 because this is setting the item in the left-hand list, rather than changing the page of the stackedWidget. I don't think they have any references.

I do agree that using magic numbers here sucks. But getting rid of them is a bit more work unless I'm missing something.

danieleds: It's true that the other code uses indices but the point is that they're not hardcoded, so we maintain a relationship between the treeWidget and the stackedWidget indices. Try this:

    ui->treeWidget->setCurrentItem(ui->treeWidget->topLevelItem(ui->stackedWidget->indexOf(ui->pageShortcuts)));

JuBan1: I could do this:
`ui->stackedWidget->setCurrentWidget( ui->pageShortcuts );
        ui->treeWidget->setCurrentItem( ui->treeWidget->topLevelItem( ui->stackedWidget->currentIndex() ) `);``

It's a bit ugly, but at least it avoids any literals.

danieleds:
"Removed integer literal, added insanity" elected as best commit message of the year :D

Merging.

PR #262, opened 2016-09-07
Update of Hungarian translation file

SandorPetrovics:
Sorry about it.
I did an updated again, because I found significant differences between own translation file and other translation files. Cause of the problem may be that I have used the lupdate 4.8.7 version. So today I ran an update with lupdate 5.5.1 version


danieleds:
No problem!

PR #263, opened 2016-09-09
Code cleanup and small fixes for Search

Teklad:
This request will be focused on cleaning up the code/bug fixing in frmSearchReplace, SearchInFilesWorker, ReplaceInFilesWorker, etc.  As well as fixing a few long-standing bugs in the process.

For the moment, I've gotten the search code quite a bit cleaner, and fixed a bug regarding threads not cleaning up properly.

I've still got to hook the threads for replaceInFiles... but so far its going very smoothly.  The end result should be something that's quite a bit easier to manage.

This change also allowed sending the results directly from the worker thread to the main thread via a signal, as well as getting rid of the thread/worker model.

This should be about 150 or so less lines of code while still getting the job done efficiently(At least judging by the current progress).

I'm also going to spend some time documenting the undocumented functions in the header files while I go through this, just to keep things clean and consistent.

### DIFF
--- a/src/ui/Search/dlgsearching.cpp
+++ b/src/ui/Search/dlgsearching.cpp
@@ -42,5 +42,5 @@ QString dlgSearching::text() const
 
 void dlgSearching::on_btnCancel_clicked()
 {
-    reject();
+    accept();
 }
--- a/src/ui/Search/filesearchresultswidget.cpp
+++ b/src/ui/Search/filesearchresultswidget.cpp
@@ -83,11 +83,15 @@ void FileSearchResultsWidget::addSearchResult(const FileSearchResult::SearchResu
 void FileSearchResultsWidget::on_doubleClicked(const QModelIndex &index)
 {
     int type = SearchResultsItemDelegate::rowItemType(index);
-
+    QPoint start = SearchResultsItemDelegate::resultRowStartData(index);
+    QPoint end   = SearchResultsItemDelegate::resultRowEndData(index);
     if (type == SearchResultsItemDelegate::ResultTypeMatch) {
         emit resultMatchClicked(
                     SearchResultsItemDelegate::fileResultRowData(index.parent()),
-                    SearchResultsItemDelegate::resultRowData(index));
+                    start.x(),
+                    start.y(),
+                    end.x(),
+                    end.y());
     }
 }
 
--- a/src/ui/Search/filesearchresultswidget.cpp
+++ b/src/ui/Search/filesearchresultswidget.cpp
@@ -35,8 +35,7 @@ void FileSearchResultsWidget::setupActions()
 void FileSearchResultsWidget::addSearchResult(const FileSearchResult::SearchResult &searchResult)
 {
     // Row, in the model, relative to this search
-    QList<QStandardItem *> searchRow;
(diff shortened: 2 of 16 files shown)


Teklad:
Didn't go quite as planned, but the new threading model is in place... everything cleans itself up properly, and got rid of a bit of redundant code.

I think maybe the searchstring class might be a little redundant though.  It might be better suited to go inside SearchHelpers if you don't see a problem with it.

Outside of that, this pretty much wraps up the request... Just needs some regression testing to make sure no new bugs were introduced. :)

danieleds:
Good! 

But I think you didn't push some files.
Also, I think you're not using an IDE so you're not noticing this, but block comments should start with `/**` (see the others).

Teklad:
Care to elaborate on me not pushing some files?  Everything is building properly on travis/via my end, so I'm not sure what I missed, lol.

danieleds:
Sorry I'm stupid... I didn't find `searchstring.h` in the changeset. I even cloned the whole pull request without finding it, but it's there.

Busy days make me crazy :-P

Teklad:
Fixed the block comments. :)

Teklad:
So I've been investigating the memory leak in filesearchresultswidget.  I made the memory usage better by pulling out only the data we need from the struct and applying it to the data roles instead of the entire structure.

What I've noticed since changing the threading around is that the data will cap itself off at the last total amount of results. As an example:

I search binary-file-20MB 5 times, then clear the results widget.  I'll be able to search 5 more times without the memory being altered.  After that, the memory will begin increasing again.

This seems to be normal behavior for QStandardItem (I tested by creating a pseudo function that added a few million of these, and then clearing the data), which is quantified when data is attached to it.

So, the best solution here would most likely be to limit the maximum amount of searches which can be displayed at any given time.  I highly doubt anyone would need 10 different searches displayed anyhow.  

On a lower memory system, not doing something about it could cause a big problem if they never bothered to close notepadqq or clear the results. 

Just wanting to kind of get your input on this.

Teklad:
Also, I'm finished with all the changes related to this request I think.  I'm happy with the results.  I'd like to get back to adding features now. Bug hunting can be painful. 

If you don't think the very last commit was necessary I can revert it before merging, but it seemed like a nice small change.

Teklad:
Any update on this @danieleds ?  :P

danieleds:
Hey sorry, I'm not dead (yet).
Couldn't find the time.
I'll try to review these changes in the weekend.

src/ui/Search/dlgsearching.cpp
```diff
@@ -42,5 +42,5 @@ QString dlgSearching::text() const
 
 void dlgSearching::on_btnCancel_clicked()
 {
-    reject();
+    accept();
```
danieleds: why?

danieleds:
it's quite cleaner!

Teklad:
Reject always returned 0.... accept returns 1 or 0 depending on how the 
message box is handled(when the GUI closes itself it returns 0, if the 
user closes, returns 1)... it helped me differentiate between user and 
system input.

danieleds:
@Teklad Finally merged :) Thank you

Issue #234, opened 2016-08-12
Replace "in selection" functionality is missing


Algomorph:
There is no "in selection" checkbox in the Find & Replace dialog.
Notepad++ has this checkbox there, which is used to limit the behavior of "Replace All" to the current selection only. 
Unlike in Notepad++, I would go even as far as having it checked by default when the Find & Replace Dialog is brought up with a _selection in the text that spans more than 1 line._ The current behavior is, when the Find & Replace Dialog is brought up, "fill in the find textbox with the contents of selection", which is most appropriate when the selection spans less than one line. I propose that we differentiate between these cases based on the line number in the selection.

danieleds:
I totally agree.

kitovyj:
Ouch, I just stubled at the missing feature in Linux. Pain.

aaronfranke:
I would also like this feature. It would make my life much easier.

davehouser1:
Any update on this? This is one of the main reasons I use NotePad++

mrjemson:
+1000 for this feature. Used this extensively in Notepad++.

PR #273, opened 2016-09-23
Adding support for replace in selection (#234)

AndersonDaniel:
Hi there!

Addition of the replace "in selection" functionality in notepad++.

### DIFF
--- a/src/ui/Search/frmsearchreplace.ui
+++ b/src/ui/Search/frmsearchreplace.ui
@@ -323,7 +323,7 @@
           <property name="frameShadow">
             <enum>QFrame::Plain</enum>
           </property>
-          <layout class="QVBoxLayout" name="verticalLayout_3">
+          <layout class="QVBoxLayout" name="verticalLayout_5">
             <item>
              <widget class="QPushButton" name="btnSelectAll">
               <property name="text">
--- a/src/ui/Search/frmsearchreplace.ui
+++ b/src/ui/Search/frmsearchreplace.ui
@@ -334,6 +334,13 @@
         </property>
        </widget>
       </item>
+	  <item row="2" column="3">
+       <widget class="QCheckBox" name="chkReplaceInSelection">
+        <property name="text">
+         <string>In selection</string>
+        </property>
+       </widget>
+      </item>
       <item row="2" column="1">
        <widget class="QLabel" name="lblSpacer1">
         <property name="text">
--- a/src/ui/Search/frmsearchreplace.ui
+++ b/src/ui/Search/frmsearchreplace.ui
@@ -294,16 +294,7 @@
         </property>
        </widget>
       </item>
-      <item row="0" column="3">
-       <widget class="QPushButton" name="btnSelectAll">
-        <property name="text">
-         <string>Select all</string>
-        </property>
-        <property name="autoDefault">
-         <bool>true</bool>
-        </property>
-       </widget>
-      </item>
+      
       <item row="1" column="1">
        <widget class="QPushButton" name="btnReplacePrev">
         <property name="text">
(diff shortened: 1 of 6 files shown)


danieleds:
Hi, thank you. 
If I am not mistaken, this only works with the "Replace All" button. I think we need to add support for search in selection too, otherwise the feature is confusing.

AndersonDaniel:
Yes, this currently only works with the "Replace All" button.
You mean that it should also work with the "Search All" button, and the effect would be to shrink the current selection to the search term?
For example, given the following text and selection (selection in bold):

> Lorem ipsum dolor sit amet
> **Lorem ipsum dolor sit amet**
> Lorem ipsum dolor sit amet

and the search term "lor", after clicking "Search All" with "in selection" checked the selected text will be:

> Lorem ipsum dolor sit amet
> **Lor**em ipsum do**lor** sit amet
> Lorem ipsum dolor sit amet

JuBan1:
Yes, I'm pretty sure that's wat he meant. 

Would be great if it worked the same way Replace works. Meaning the "Search In Selection" is automatically checked if the user opens the search dialog while having some text selected.

danieleds:
Yes, it's like @JuBan1 said.

And it should also work with "Search prev / search next / replace prev / replace next"

AndersonDaniel:
Okay, so:

1. Implemented the "Search all in selection" functionality
2. Differentiating between two possibilities when the user searches and there is a selection:
    a) Multi-line selection - "In Selection" checkbox is automatically checked, selection isn't copied to search text
    b) Single line selection - "In Selection" is unchecked and selection is copied to search text.
3. While "In Selection" is checked, editing the search text (in "Find" tab) doesn't trigger the "Find next" operation (otherwise the selection is lost and there's no meaning to the "search all in selection" operation)

Concerning search next / prev, replace next / prev - any of them replaces the current selection, so having "in selection" affect them means the operation will only work once (after which the selection will be the first search /replace result). Seems to me a bit strange behavior... What do you think?

JuBan1:
'Replace All' works great. "Select All" still selects *all* instances of the searched term for me.

> Concerning search next / prev, replace next / prev - any of them changes the current selection, so having "in selection" affect them means the operation will only work once (after which the selection will be the first search /replace result). Seems to me a bit strange behavior... What do you think?

Yes, that's a bit of a problem. If it turns out Find Next/Prev don't work then they should probably be greyed out if 'In Selection' is checked.

Simply remembering the start and end of the selection and still allowing Find Next/Prev is probably a bad idea since it would be confusing. Users should be able to clearly see the part of the text that is searched. 

A possible solution is to use some highlight color (similar to what happens when you select a word and all other occurrences of this word are highlighted as well). When the user uses Find Next/Prev the selection could be used as usual (to show the found occurrence) and the search area could be highlighted with a different color. 

On the downside, it's probably a little more involved to implement.

AndersonDaniel:
Added the part of graying out the Next/Prev buttons when 'In Selection' is active.

About "Select All" still selecting all instances, it could be a bit confusing since when even a single instance is selected, all of them are highlighted (not only in selection). For me it works, could you please verify this is not the case? (For example by using "Select All" with "In Selection" checked and then editing the selected text.)

Thanks!

JuBan1:
Yes, you're right. I confused the highlighting with selection, derp.

JuBan1:
Alright, looking good. I have two more suggestions.

First, In Selection should probably be toggleable when the search window is opened without any text selected. So users can turn on In Selection search without having to close the window, select some text, then reopen it. Right now the checkbox is disabled.

Next, I thought wrapping the search and replace buttons as well as the selection checkbox in a QFrame might be a nice way to visually group them. Here's a mock-up:



Although I don't know how good that would look with the way the search UI is constructed (= buttons being re-used and layout changed depending on which tab is selected).

AndersonDaniel:
Fixed both, QFrame looks fine I think:
qframe_find
qframe_replace

What do you think?

src/ui/include/mainwindow.h
```diff
@@ -293,6 +293,7 @@ private slots:
     QString             currentWordOrSelection();
     void                currentWordOnlineSearch(const QString &searchUrl);
     QString             getNewDocumentName();
+    void                showSearchReplace(frmSearchReplace::Tabs tab);
```
JuBan1: One last round of nagging and I'll be content.

Could you add a short documentation to this new function?

src/ui/Search/frmsearchreplace.ui
```diff
@@ -324,15 +315,44 @@
...
-       </widget>
+      <item row="0" column="3" rowspan="3">
+        <widget class="QFrame" name="frameSearchReplaceAll">
+          <property name="frameShape">
+            <enum>QFrame::StyledPanel</enum>
+          </property>
+          <property name="frameShadow">
+            <enum>QFrame::Plain</enum>
+          </property>
+          <layout class="QVBoxLayout" name="verticalLayout_3">
```
JuBan1: I'm getting the following warning:

> Search/frmsearchreplace.ui: Warning: The name 'verticalLayout_3' (QVBoxLayout) is already in use, defaulting to 'verticalLayout_31'.

Name's already in use in line 96.

JuBan1:
I like it. I tested it on my end as well and it feels pretty good to use.

danieleds:
This is how Qt Creator does it:

ex

I like it a lot, but I'm not sure how easy it is to implement.
If it's not, then the current behavior is good but I'd like to think about the UI before merging.

AndersonDaniel:
Fixed both things, thanks :)

AndersonDaniel:
Started playing with CodeMirror highlighting. Surrounding the selection with a border like Qt Creator seems rather difficult; we could highlight the previously selected text something like this:
example
But I don't know if it looks intuitive enough, and I also think it would require a lot of code to handle special cases with replace when the replacement length is not equal to the replaced text's length (also for regexp replacing).

JuBan1:
How does Replace In Selection handle it right now? Shouldn't it already have to handle the change of the selected code?

AndersonDaniel:
Yes, CodeMirror internally takes care of that when replacing. From the docs:

> **replace**(text: string, ?origin: string)
> Replaces the currently found match with the given text and _adjusts the cursor position to reflect the replacement._

Teklad:
If you accessed the drawing functions a bit, you could probably paint an overlay using QPainter, but I debate how hacky that would be, lol.

Anyways, this may have a few conflicts with #263 in terms of merging, so it'd probably be better if this pull request was merged first.  The changes in this request are much smaller so I shouldn't have any problems fixing conflicts when rebasing if I run into any.

valera-rozuvan:
@AndersonDaniel Any progress?

AndersonDaniel:
@valera-rozuvan I didn't pursue the previous-selection-highlighting direction further (I think it would be too complicated to do it properly). The rest was ready for merge since the last commit, I see conflicts have accumulated since then.

valera-rozuvan:
I am guessing this PR is no longer desirable, and too far behind the `master` branch? Maybe close it?

@danieleds , @AndersonDaniel , @JuBan1 - Your thoughts?

PR #276, opened 2016-09-29
Fix document icons on KDE. Resolves #275.

### LINKED ISSUE #275: no indication if file was edited on KDE
When I edit a file, I can't see anywhere if the changes were already saved or not. I never had this issue on Unity. Now I'm using Kubuntu 16.04 (KDE 5.18.0 / Plasma 5.5.5, Qt 5.5.1).

edited

JuBan1:
`QIcon::hasThemeIcon` is used to query whether an icon should be loaded from an installed theme or from Notepadqq's icon files. In the case of "notepadqq-document-saved" and "notepadqq-document-unsaved" the function erroneously returns `true`.

The reason appears to be that the two icon names start with "notepadqq". I assume that is enough for KDE to act like an icon called "notepadqq" was requested - which obviously exists.

Solution is to rename the two icons.


Teklad:
Boy these pull requests are piling up.  I hope @danieleds gets some free time soon otherwise merging all these might become painful. xD

JuBan1:
Thankfully they're mostly in entirely different places in the codebase. 

Best we can do is have them in presentable condition when he has the time to look at and merge them.

danieleds:
Thanks guys!

I'm trying to merge small and quick PRs like these as soon as possible. Sorry @Teklad for #263 , it's big so I'm taking a look one bit at a time, I hope to merge it soon

Issue #281, opened 2016-10-01
the translation does not work 


david-geiger:
Hi,

I have recently pull-requested the French translation and tested notepadqq with the latest git master snapshot, but after installing the package and starting it I see that the language is still on default English.
So I set "French" language in Settings -> Preferences... -> General -> Localization and restart notepadqq but same default English language is still in use, no action done.
So after a quick search I found what is going wrong, here the attached patch:

[fix-translations.patch.txt](https://github.com/notepadqq/notepadqq/files/504365/fix-translations.patch.txt)

Also another question: would it be possible at first run to autodetect the language from users system? so users do not need to select themselves the languages.

Best Regards,
David

Teklad:
I've sent in a pull request #282  to patch up this issue for you as well as include the language auto-detection.  :)

Teklad:
Could you elaborate on how you installed the package?  I'm trying to grasp what is going on with your system which causes this.  

Also, can you verify if the directories /usr/local/share/notepadqq/translations or /usr/share/notepadqq/translations exist for you?

david-geiger:
Yeah! I understand better now, we installed all binaries in "/usr/bin/" as it should be and not in "/usr/lib/project_name/", this is why it can't find translations in "/usr/share/notepadqq/translations/"

$ ls /usr/bin/ |grep notepadqq
notepadqq*
$
$ ls /usr/share/notepadqq/translations/
notepadqq_de.qm  notepadqq_fr.qm  notepadqq_hu.qm  notepadqq_pl.qm  notepadqq_ru.qm
$

So why install the binary in /usr/lib/notepadqq/ ?
And why not include all translations directly compiled in the binary?

david-geiger:
Yes I think so, embed the translations into the binary will be the best solution :)

danieleds:
I'm going to bundle them into the binary.

PR #282, opened 2016-10-01
Language detection and a patch

Teklad:
This resolves issue #281 in full.  It tries to detect default system locale by default, and if the translation for that locale doesn't exist defaults back to English. 

I'll need a couple of people to test this to make sure the patch @david-geiger sent in doesn't cause any issues for anyone else.  I used `export LANG=fr_CA.UTF-8` to test and it seemed to work okay on my end.

### DIFF
--- a/src/ui/include/nqqsettings.h
+++ b/src/ui/include/nqqsettings.h
@@ -63,8 +63,7 @@ class NqqSettings {
 public:
 
     BEGIN_GENERAL_CATEGORY(General)
-        NQQ_SETTING(FirstRun,                       bool,       true)
-        NQQ_SETTING(Localization,                   QString,    "en")
+        NQQ_SETTING(Localization,                   QString,    "")
         NQQ_SETTING(CheckVersionAtStartup,          bool,       true)
         NQQ_SETTING(WarnForDifferentIndentation,    bool,       true)
 
--- a/src/ui/include/nqqsettings.h
+++ b/src/ui/include/nqqsettings.h
@@ -63,6 +63,7 @@ class NqqSettings {
 public:
 
     BEGIN_GENERAL_CATEGORY(General)
+        NQQ_SETTING(FirstRun,                       bool,       true)
         NQQ_SETTING(Localization,                   QString,    "en")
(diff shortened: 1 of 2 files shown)


src/ui/include/nqqsettings.h
```diff
@@ -63,6 +63,7 @@ class NqqSettings {
 public:
 
     BEGIN_GENERAL_CATEGORY(General)
+        NQQ_SETTING(FirstRun,                       bool,       true)
         NQQ_SETTING(Localization,                   QString,    "en")
```
danieleds: Instead of a "FirstRun" setting, can't you change the default value for "Localization" to an empty string and check if that is empty?

Teklad: Yea, but I was thinking about other settings that might need to be initialized in the future on a first run basis... if you don't think this is an issue I can default it to empty instead.

Teklad:
I think that should cover your request.  No point in trying to think too far ahead here, lol.

src/ui/main.cpp
```diff
@@ -60,10 +70,12 @@ int main(int argc, char *argv[])
     } else if (translator.load(QLocale(langCode),
                                QString("%1").arg(qApp->applicationName().toLower()),
                                QString("_"),
-                               QString("%1/../../share/%2/translations")
+                               QString("%1/../share/%2/translations")
```
danieleds: Why this change? When installed, notepadqq is located at `/usr/lib/notepadqq/notepadqq-bin`, so the double `../` was ok.
Maybe add a comment next to that line, saying `// The Notepadqq binary gets installed at /usr/lib/notepadqq/notepadqq-bin`

Teklad: You're right... this actually breaks the installation on my end (when installed)...  But according to #281 its already broken for another user, though I'm sure there's special circumstances involved.

I'll go ahead and revert this small change so it'll just be a feature request rather than a bug fix, but we need to see about a more reliable way to detect this once installed on the system in the future.

danieleds:
Merged: 0a40d97edf71c505074210290b0291f3d6d3f72e

PR #283, opened 2016-10-01
REAMDE: coreutils are always present

darealshinji:


danieleds:
I prefer to keep it in the readme if there is not a strong reason to remove it. It is a dependency, and having it clearly stated can't do any harm.

Feel free to reopen if you need!

PR #284, opened 2016-10-01
Sessions Refactored

JuBan1:
In order to make code review and merging less painful I have split #274 into two parts. This part is purely refactoring and should not have any visible effects on how Nqq works. This makes the autosaving patch significantly easier to review.

Changes done:

* `PersistentCache` is not responsible for clearing cache directories anymore. That was moved to `saveSession`.
* `saveSession` and `loadSession` were moved from `MainWindow`to `Sessions`
* `saveSession` can now use any directory to write cache files to (in preparation for autosaving).
* Lots of helper code (like `TabData` or `SessionWriter`) were moved from `Sessions.h` to `Sessions.cpp` because nothing else needs them.
* A few functions of `MainWindow` were made public because they're needed for loading or saving sessions.
* `MainWindow::on_documentLoaded` got another parameter to determine whether the loaded document should be added to the Recently Used Documents list.

90% of changes are just moving around code which will hopefully make it easy to merge.


danieleds:
Thanks a lot :)

danieleds:
Are you able to remove the dependency on MainWindow from Sessions?

This is what you're currently using:
 * `window->topEditorContainer()`: can be passed as a parameter
 * `window->getDocEngine()`: can be passed as a parameter
 * `window->getNewDocumentName()`: can be passed as a std::function parameter
 * `window->removeTabWidgetIfEmpty()`: can be done inline
 * `window->refreshEditorUiInfo()`: MainWindow can call it right after `loadSession()` returns (it should be able to determine what is the focused editor by itself)

JuBan1:
Theoretically yes but to what end? Sounds to me like we'd just pluck apart MainWindow by doing that. 

`#include "MainWindow.h" is only in `Sessions.cpp` so everything that is pulled in by that `include` stays in this compilation unit.

danieleds:
This is just to facilitate unit testing in the future.

JuBan1:
Okay, so how about:

> window->topEditorContainer()
> window->getDocEngine()
> window->refreshEditorUiInfo()

As you said.

> window->getNewDocumentName():

Turn this into `static(?) QString DocEngine::getNewDocumentName()` since this probably doesn't belong in `MainWindow` anyways. And it removes the need to use a `std::function`.

> window->removeTabWidgetIfEmpty();

Turn this into `void EditorTabWidget::removeIfEmpty()` since it also probably shouldn't be in `MainWindow`. Either `EditorTabWidget` or `TopEditorContainer` is more suitable.

danieleds:
> Turn this into static(?) QString DocEngine::getNewDocumentName() since this probably doesn't belong in MainWindow anyways. And it removes the need to use a std::function.

Yes, better!

> Turn this into void EditorTabWidget::removeIfEmpty() since it also probably shouldn't be in MainWindow. Either EditorTabWidget or TopEditorContainer is more suitable.

Okay. I can see it either as a static method of EditorTabWidget, or as an instance method of TopEditorContainer. It's your call.

danieleds:
> Turn this into static(?) QString DocEngine::getNewDocumentName() since this probably doesn't belong in MainWindow anyways. And it removes the need to use a std::function.

Actually, I'd prefer an instance method given that the result will change with each invocation.

JuBan1:
So while doing the modifications I noticed something.

Passing a `DocEngine*` and a `TopEditorContainer*`to the session functions really only serves to shoot programmers in the foot. You can't just pass any DE* and TEC* to the function because they have to be "a pair" (e.g. things get weird if a DE* starts saving documents that weren't also loaded through it) and both DE* and TEC* are connected to a `MainWindow*` instance through a number of signals that may even end up calling functions from the respective other again (e.g. TEC* sends a signal, it is processed by MW* which calls functions of DE* again).

These three classes are so strongly interwoven that unit testing one without the other is going to be nigh impossible.

Teklad:
Some of the code base is quite a bit older in terms of practices used (I 
was around when we still used QScintilla).  I don't think @danieleds 
would be against figuring out ways to detach these systems from one 
another as best as possible. I know I'm not.

JuBan1:
Easier said than done, sadly.

Well, the modifications are done. In addition to what is outlined above I also had to move an `enum` into `DocEngine` and I found a way of calling `refreshEditorUiInfo()` without needing an instance of `MainWindow`

danieleds:
@Teklad I agree.

Merging!

PR #289, opened 2016-10-02
Run menu

Teklad:
This adds fun support for run menu stuff.  Putting this pull request up to get some feedback on what else I might need to do to polish this up.

Things I know I need:
- QLabel describing use of %fullpath% %filename%, etc.
- A less ugly(more functional) "Run..." dialog.

Any other suggestions to functionality changes are welcome... One of the main things is it works very similar to if you were to use bash (make sure to put quotes around arguments with spaces).

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -19,6 +19,8 @@
 #include "include/Sessions/persistentcache.h"
 #include "include/Sessions/sessions.h"
 #include <QFileDialog>
+#include <QLineEdit>
+#include <QInputDialog>
 #include <QMessageBox>
 #include <QClipboard>
 #include <QUrl>
@@ -1943,17 +1945,23 @@ void MainWindow::on_actionInterpret_as_triggered()
 #include <QDebug>
 void MainWindow::setupRunMenu()
 {
-    QHash <QString, QString> runners = m_settings.Run.getCommands();
-    QHashIterator<QString, QString> i(runners);
+    QMap <QString, QString> runners = m_settings.Run.getCommands();
+    QMapIterator<QString, QString> i(runners);
     ui->menuRun->clear();
+
+    QAction *a = ui->menuRun->addAction(tr("Run..."));
+    connect(a, SIGNAL(triggered()), this, SLOT(runCommand()));
+    ui->menuRun->addSeparator();
+
     while (i.hasNext()) {
         i.next();
-        QAction *a = ui->menuRun->addAction(i.key());
+        a = ui->menuRun->addAction(i.key());
         a->setData(i.value());
+        a->setObjectName("RunCmd"+a->text());
         connect(a, SIGNAL(triggered()), this, SLOT(runCommand()));
     }
     ui->menuRun->addSeparator();
-    QAction *a = ui->menuRun->addAction(tr("Modify Run Commands"));
+    a = ui->menuRun->addAction(tr("Modify Run Commands"));
     connect(a, SIGNAL(triggered()), this, SLOT(modifyRunCommands()));
 
 }
@@ -1961,14 +1969,31 @@ void MainWindow::setupRunMenu()
 void MainWindow::modifyRunCommands()
 {
     RunPreferences p;
-    p.exec();
+    if(p.exec() == 1) {
+        setupRunMenu();
+    }
 }
 
 void MainWindow::runCommand()
 {
     QAction *a = qobject_cast<QAction*>(sender());
+    QString cmd;
+
+    if (a->data().toString().size()) {
+        cmd = a->data().toString();
+    } else {
+        bool ok;
+        //TODO: Nicer command input, with save option.
+        cmd = QInputDialog::getText(this, tr("Enter command text"),
+                                    tr("Command:"), QLineEdit::Normal,
+                                    "", &ok);
+        if(!ok || cmd.isEmpty()) {
+            return;
+        }
+    }
+
     Editor *editor = currentEditor();
-    QString cmd = a->data().toString();
+
     QUrl url = currentEditor()->fileName();
     QStringList selection = editor->selectedTexts();
     if (!url.isEmpty()) {
@@ -1979,7 +2004,9 @@ void MainWindow::runCommand()
     if(!selection.first().isEmpty()) {
         cmd.replace("\%selection\%",selection.first());
     }
-    QProcess::startDetached(cmd);
+    if(!QProcess::startDetached(cmd)) {
+        
+    }
 }
 
 void MainWindow::on_actionPrint_triggered()
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -1975,6 +1975,52 @@ void MainWindow::modifyRunCommands()
     }
 }
 
+QStringList parseCombinedArgString(const QString &program)
+{
+    QStringList args;
+    QString tmp;
+    int quoteCount = 0;
+    bool inQuote = false;
+    // handle quoting. tokens can be surrounded by double quotes
+    // "hello world". three consecutive double quotes represent
+    // the quote character itself.
+    for (int i = 0; i < program.size(); ++i)
+    {
+        if (program.at(i) == QLatin1Char('"'))
+        {
+            ++quoteCount;
+            if (quoteCount == 3)
+            {
+                // third consecutive quote
+                quoteCount = 0;
+                tmp += program.at(i);
+            }
+            continue;
+        }
+        if (quoteCount)
+        {
+            if (quoteCount == 1)
+                inQuote = !inQuote;
+            quoteCount = 0;
+        }
+        if (!inQuote && program.at(i).isSpace())
+        {
+            if (!tmp.isEmpty())
+            {
+                args += tmp;
+                tmp.clear();
+            }
+        }
+        else
+        {
+            tmp += program.at(i);
+        }
+    }
+    if (!tmp.isEmpty())
+        args += tmp;
+    return args;
+}
+
 void MainWindow::runCommand()
 {
     QAction *a = qobject_cast<QAction*>(sender());
@@ -2005,6 +2051,7 @@ void MainWindow::runCommand()
     if(!selection.first().isEmpty()) {
         cmd.replace("\%selection\%",selection.first());
     }
+    cmd.replace("\\\"","\"\"\"");
     if(!QProcess::startDetached(cmd)) {
         
     }
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -1943,7 +1943,7 @@ void MainWindow::on_actionInterpret_as_triggered()
 
     dialog->deleteLater();
 }
-#include <QDebug>
+
 void MainWindow::setupRunMenu()
 {
     QMap <QString, QString> runners = m_settings.Run.getCommands();
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -160,6 +160,7 @@ MainWindow::MainWindow(const QString &workingDirectory, const QStringList &argum
     }
 
     setupLanguagesMenu();
+    setupRunMenu();
 
     showExtensionsMenu(Extensions::ExtensionsLoader::extensionRuntimePresent());
 
@@ -1939,6 +1940,37 @@ void MainWindow::on_actionInterpret_as_triggered()
 
     dialog->deleteLater();
 }
+#include <QDebug>
+void MainWindow::setupRunMenu()
+{
+    QHash <QString, QString> runners = m_settings.Run.getCommands();
+    QHashIterator<QString, QString> i(runners);
+    ui->menuRun->clear();
+    while (i.hasNext()) {
+        i.next();
+        QAction *a = ui->menuRun->addAction(i.key());
+        a->setData(i.value());
+        connect(a, SIGNAL(triggered()), this, SLOT(runCommand()));
+    }
+}
+
+void MainWindow::runCommand()
+{
+    QAction *a = qobject_cast<QAction*>(sender());
+    Editor *editor = currentEditor();
+    QString cmd = a->data().toString();
+    QUrl url = currentEditor()->fileName();
+    QStringList selection = editor->selectedTexts();
+    if (!url.isEmpty()) {
+        cmd.replace("\%fullpath\%", url.toString(QUrl::None));
+        cmd.replace("\%path\%", url.path(QUrl::FullyEncoded));
+        cmd.replace("\%filename\%", url.fileName(QUrl::FullyEncoded));
+    }
+    if(!selection.first().isEmpty()) {
+        cmd.replace("\%selection\%",selection.first());
+    }
+    QProcess::startDetached(cmd);
+}
 
 void MainWindow::on_actionPrint_triggered()
 {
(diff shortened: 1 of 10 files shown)


Teklad:
Okay, so I have some free time to finish this up on the weekend.  Just kind of needing some opinions on the interface for the run menu.

I know I still need to implement the "save" option for the "Run..." dialog.  Aside from that, I was wondering if the way the add/remove dialog works is okay or maybe it needs to be changed up a little bit.

Right now it automatically creates the next entry to type into when you add a new entry.  Wondering if maybe I should have an "Add" button instead perhaps.

Any other opinions welcomed as well.

JuBan1:
Looks neat. I think automatically adding a new line is nicer than manually having to "Add" items.

I'd say the font size of the table should be a tiny bit larger- it's unusually small, is it not? And the accept button should probably read "OK" instead of "Okay" to bring it in line with the other buttons.

Teklad:
@danieleds care to give this a shot? Just wanting to get a read on what else needs to be polished up.

JuBan1:
Okay, sorry, I see now what was irritating me. It wasn't the small font size but the small section size of the table that made everything look squished. Now that I compare the old vs new build it becomes apparent.

IMO I like it best with the default font size and `vh->sectionResizeMode(QHeaderView::Fixed);   vh->setDefaultSectionSize(18 or 20);`. UI font size should probably not scale through FontOverride since that one is for the editor. UI font size is usually determined by the operating system (to properly handle high DPI although I don't know if Nqq does that well right now).

Right now the `Remove` button is default focused when opening the dialog, so if I simply hit Return/Enter it'll delete one command. Same if I just write something into one of the cells and hit Return/Enter twice. That's probably a bad behavior. It'd be great if we had a small delete icon next to each line, just like the Open File icon is there as well.

Other than that... once you add an info label for possible %parameters% I don't see anything else to nitpick about.

Teklad:
Okay, all done with those changes... only problem is I don't think we have a proper "Remove" icon to put in its place.... any suggestions there? :P

JuBan1:
You can access theme icons with `IconProvider::fromTheme("someString")`. A few examples that seem to fit are "remove", "delete-comment", or "edit-delete".

Teklad:
Alright, so where should I position the remove button?  Before or after the open document button... or maybe you have a better placement idea in mind?

JuBan1:
I think I'd put them to the right of the table to visually differentiate it from the editing options.

Teklad:
I'll give it a shot.  Need to figure out how to get it to stop clipping my painter so I can draw outside the widget.

Teklad:
So drawing outside the widget would introduce extra dependencies I'm not really comfortable having in a preferences dialog... We could either use it as it is now or try to modify the paint event of the table itself to constrain the right margin with enough space to place the remove buttons nicely. The latter would obviously be a lot more work, but it would look pretty nice I think. Alternatively could try using an extra table column which would mimic the background color of the dialog.

danieleds:
I like it!

> Alternatively could try using an extra table column which would mimic the background color of the dialog.

I'd put them into an extra column.

Also, for some reason I'm not able to launch a terminal using the following command:

    gnome-terminal -e "bash -c \"echo hello; exec bash\""

I don't know if I'm missing something

Teklad:
I figured out the issue with that... we'll need to add a preference which will point to the shell interpreter they prefer to use for executing commands(this is a lot like other programs do this).

Give this latest commit a try and make sure the behavior is more along the lines of what you expect just to be on the safe side.

JuBan1:
First of all, it looks very good.

I think clicking the delete button on the lowest row should only clear the entries, not entirely delete the row. The way it is right now I can delete the lowest row and then I'm unable to add any new commands until I reopen the window. Instead, just clearing the text is probably better.

Also, it might actually be a good idea to move the two buttons into a separate column. If you have a rather long command it is overlapped by the icons.

This here shows both problems:

Image

Teklad:
Alright. Both  should be pretty easy tasks I think.  Should have something a bit more fluid up in the next couple of hours.

danieleds:
Hmm, with the latest commit I'm not able to launch anything: for me `bash` is in `/bin`, not `/usr/bin`.

Actually, I found the problem with your previous version of the code (which I prefer). The issue is that [Qt is stupid](http://doc.qt.io/qt-5/qprocess.html#start) and in order to execute this:

    gnome-terminal -e "bash -c \"echo hello; exec bash\""

one has to pass this:

    gnome-terminal -e "bash -c \\"""echo hello; exec bash\\""""

So, for now, this is a minor issue and I'd rollback to fd938b64244f52629f0551fe4ec3b.

Possible solutions:
 * let the user specify each argument separately (and take the opportunity to allow specifying a working directory for the command)
 * automatically translate the first command into the second one (it's not obvious if it's practical or not)

So, forgetting just for a moment about this, it works great for me. Are stylistic/usability issues the only missing things?

Teklad:
@JuBan1 I've made some changes which should address the issues you noticed, please give it a test and make sure it is up to standard for you.  

@danieleds, some brainstorming is necessary to figure out all the nuances of getting it to behave like it should without an interpreter. I reverted the previous commit as requested, however.

Having all the arguments individually specified in the interface would probably look ugly and might seem rather counter-intuitive for most people.  I do agree that having a way to set the working directory for a command would be useful, however.

I'm currently at a bit of a loss on how to efficiently fix the odd syntax issue without forcing an interpreter of some sort, though I'm sure it's a lot simpler than it seems on paper.

With that said, I think that squares off more or less everything for this.  Features like "save from run dialog" are really extras which could be easily added later, not to mention I don't want this pull request getting much fatter. >.>

danieleds:
@Teklad yeah, we should think about it. I believe that a small parser could be a good solution, but let's see if there is a simpler way. 

Small note: remember to apply the text color for selected items in paint(). For the search results widget, we did it like this:

        // set text color for selected item
        if (option.state & QStyle::State_Selected)
            ctx.palette.setColor(QPalette::Text, options.palette.highlightedText().color());

JuBan1:
> please give it a test and make sure it is up to standard for you. 

From a UI perspective I give my stamp of approval(TM).

I haven't given the string parsing much thought but [here](http://stackoverflow.com/questions/22685877/passing-argument-string-which-contains-spaces-and-quotes) is a SO thread that discusses the issue and outlines a potential solution.

Teklad:
Okay.... highlighting color applied... I just adjusted the pen prior to drawing the text if selected, which gets the same result really.

Now the only remaining issues are the ones revolving command syntax.

Teklad:
As far as QT is concerned, and from testing using the SO link @JuBan1 provided, the QT syntax for running your previous command is quite literally:
```bash
gnome-terminal -e "bash -c """echo hello; exec bash""""
```

If this syntax doesn't bother you too bad, we could go with it and just add a bit of information in the RunPreferences menu to let them know that 3 double quotes = 1 literal quote.

Teklad:
Actually, give this latest commit a test and do your typical command scenario, i.e.:
```
gnome-terminal -e "bash -c \"echo hello; exec bash\""
```

If you don't run into any issues with it, we can just use it like this... I don't think it will cause any compatibility issues but just need to be sure.

danieleds:
@Teklad your latest commit doesn't handle for example this:

    gnome-terminal -e "bash -c \"cd \\\"$(dirname '%path%')\\\"; exec bash\""

which is used to launch a terminal in the directory of the current file.

I've prepared a small parser which works fine, I'll send it to you soon.

src/ui/mainwindow.cpp
```diff
@@ -1940,40 +1944,87 @@ void MainWindow::on_actionInterpret_as_triggered()
...
+        }
+    }
+
+    Editor *editor = currentEditor();
+
+    QUrl url = currentEditor()->fileName();
+    QStringList selection = editor->selectedTexts();
+    if (!url.isEmpty()) {
+        cmd.replace("\%fullpath\%", url.toString(QUrl::None));
+        cmd.replace("\%path\%", url.path(QUrl::FullyEncoded));
```
danieleds: Why the escape for `%`?

src/ui/mainwindow.cpp
```diff
@@ -1940,40 +1944,87 @@ void MainWindow::on_actionInterpret_as_triggered()
...
-        args << fileName.toString(QUrl::None);
-        QProcess::startDetached("firefox", args);
+    QAction *a = qobject_cast<QAction*>(sender());
+    QString cmd;
+
+    if (a->data().toString().size()) {
+        cmd = a->data().toString();
+    } else {
+        bool ok;
+        //TODO: Nicer command input, with save option.
```
danieleds: Yep, an ad-hoc dialog would be better. Can you do it within this PR?

danieleds:
(not sure how this "github reviews" works... just trying ;P )

danieleds:
Here's the parser:

```c++
/**
 * @brief Given a command string, extracts the application name and the list of arguments.
 *        The arguments are separated by one or more spaces. Arguments containing spaces
 *        must be quoted to be correctly supplied to the new process. You can quote either
 *        with `'` or `"`. Within a quoted argument, you can escape the quote character by
 *        prepending a `\`. Within a double-quoted argument, you can also escape `\` by
 *        prepending another `\`.
 * @param cmd
 * @return
 */
QStringList parseCommandString(QString cmd) {
    QStringList parts;
    const char OUT = '\0';
    char quote = OUT;
    QString curr = "";
    for (int i = 0; i < cmd.length(); i++) {
        if (cmd[i] == '"') {
            if (quote == '"') {
                quote = OUT;
                parts.append(curr);
                curr = "";
            } else if (quote == '\'') {
                curr += '"';
            } else {
                quote = '"';
            }

        } else if (cmd[i] == '\'') {
            if (quote == '\'') {
                quote = OUT;
                parts.append(curr);
                curr = "";
            } else if (quote == '"') {
                curr += '\'';
            } else {
                quote = '\'';
            }

        } else if (cmd[i] == '\\') {

            if (i+1 < cmd.length()) {
                i++;
                if (quote == '\'') {
                    if (cmd[i] == '\'') {
                        curr += '\'';
                    } else {
                        curr += '\\';
                        curr += cmd[i];
                    }
                } else if (quote == '"') {
                    if (cmd[i] == '"') {
                        curr += '"';
                    } else if (cmd[i] == '\\') {
                        curr += '\\';
                    } else {
                        curr += '\\';
                        curr += cmd[i];
                    }
                } else {
                    curr += cmd[i];
                }
            } else {
                curr += '\\';
            }

        } else if (cmd[i] == ' ') {
            if (quote == OUT) {
                if (curr.length() > 0) {
                    parts.append(curr);
                    curr = "";
                }
            } else {
                curr += ' ';
            }
        } else {
            curr += cmd[i];
        }
    }
    if (curr.length() > 0) {
        parts.append(curr);
    }

    return parts;
}
```

You can use it like this:

```c++
QStringList args = parseCommandString(cmd);
QString program = args.takeFirst();
QProcess::startDetached(program, args);
```

Teklad:
The \% is so the syntax highlighter on VIM doesn't make me go crazy(OCD ya know)... it doesn't really hurt anything and % does have special meaning under certain circumstances so I was just being safe mostly.

I don't mind getting the run command dialog within this setup, but the pull request will get fatter (I really think I'm getting good at big pull requests).  I'll implement your mini-parser shortly... is there any particular place you would like me to put it?

I think a RunCommand class which contains the dialog and run function would probably be best at this point to keep mainwindow clean-ish.  As clean as it can be anyways.

Teklad:
Nevermind that last bit... I think I found something we can all possibly agree on for format of the code.  I'm almost done implementing the run dialog bits... after that just need to cram your parser in there and we can call it complete. :)

Teklad:
Okay, I've pushed a slightly better run dialog with a save option... and included your parser in this latest commit.  Just need to ensure everything works thus far with some testing.

May need to make some changes to the run dialog... I'm not sure I like the popup behavior for typing the name of the command for the run menu, so any better ideas to make it less ugly are welcome. :)

Teklad:
I can revert it, but the idea was to have the bash location modifiable 
in preferences (possibly auto-located if the setting doesn't exist 
yet).  This allows for "terminal" syntax, so to speak... but it is a bit 
less portable (probably won't work on windows) than fd938b6.


On the other hand, having to specify each argument individually would be 
rather cumbersome to users that like to use argument heavy commands like 
ffmpeg, as an example, and make saving the commands a real annoyance 
code-wise (but it is doable with enough elbow grease).


With that said, I don't mind reverting to the previous commit, but it'll 
make the syntax for more complicated commands a little wonky for the 
moment until I can figure out a way to make it behave properly (thank 
you QT).  I'm not against rolling a custom "forking" library for this 
particular purpose if need be... or finding one that's easily ported to QT.


On your last point, you would be correct. Outside of a few minor changes 
to usability interface-wise, the code is more or less complete.  I 
should have the free time this Saturday to finish up on it, assuming I 
can find something noteworthy to help with running commands in a fashion 
that doesn't make the syntax really awkward.  The arguments method is on 
the table as well if no suitable solution is found, of course.  I will 
flesh it out to include options like working directory in the future, 
regardless of which method is used.


On a side note, it turns out QT does use PATH for the start function, so 
this is valid:

Process.start("bash", QStringList() << "-c" << arg);


Would you be against having a checkbox to "use interpreter to run 
commands" possibly?  This would allow for much more syntaxically 
complicated commands for users if they need it. It wouldn't have to be 
locked to bash specifically either, of course.

Teklad:
The \% is so the syntax highlighter on VIM doesn't make me go crazy(OCD 
ya know)... it doesn't really hurt anything and % does have special 
meaning under certain circumstances so I was just being safe mostly.

I don't mind getting the run command dialog within this setup, but the 
pull request will get fatter (I really think I'm getting good at big 
pull requests).  I'll implement your mini-parser shortly... is there any 
particular place you would like me to put it?

I think a RunCommand class which contains the dialog and run function 
would probably be best at this point to keep mainwindow clean-ish.  As 
clean as it can be anyways.

danieleds:
> On a side note, it turns out QT does use PATH for the start function

Perfect!

(I'm not sure if some of the other points of your message are still valid now that we have the parser)

> I'm not sure I like the popup behavior for typing the name of the command for the run menu

What do you mean?

Teklad:
You can disregard that... I was being nitpicky. My perfectionist nature
tends to get the best of me at times.

Anything else this PR needs?

danieleds:
> You can disregard that... I was being nitpicky. My perfectionist nature tends to get the best of me at times.

I like perfectionism :D Feel free to suggest anything, from structural changes to fix-that-space-at-the-end-of-the-line kind of things :P

> Anything else this PR needs?

I think we're good, I'll test it a little more in the next days and then merge.

danieleds:
Okay, three things:

 * If I save the command from the Run dialog, then I don't see it in the Run menu
 * I'd show the list of "special placeholders" in the Run dialog like you did in the other dialog
 * When I save a command from the Run dialog, it's not clear that something happened. We should maybe show a confirmation dialog ("your command has been saved, you can find it in the Run menu") or some alternative.

Teklad:
Finally got around to patching up the issues you described... I used a status label beside the save button that fades out after about 10 seconds.  It seemed a bit less clunky than force-feeding another dialog in the user's face, lol.

Give it a shot and see what you think.

src/ui/nqqrun.cpp
```diff
@@ -304,9 +320,27 @@ void RunDialog::slotSave()
...
+        m_status->setText(tr("Command saved..."));
+        QGraphicsOpacityEffect *eff = new QGraphicsOpacityEffect(m_status);
+        m_status->setGraphicsEffect(eff);
+        QPropertyAnimation *a = new QPropertyAnimation(eff, "opacity");
+        a->setDuration(10000);
+        a->setStartValue(5);
+        a->setEndValue(0);
+        a->setEasingCurve(QEasingCurve::Linear);
+        a->start(QPropertyAnimation::DeleteWhenStopped);
+        connect(a, SIGNAL(finished()), this, SLOT(slotHideStatus()));
```
danieleds: for uniformity, use the new syntax for signal/slots: https://wiki.qt.io/New_Signal_Slot_Syntax

Teklad:
Okay, all fixed up (I think).  Just double check to make sure everything works as intended. :)

danieleds:
Thanks, it looks fine! :)
Merging

PR #295, opened 2016-10-15
Remember a document's display language in sessions.

JuBan1:
With this patch sessions will also remember the set display (or syntax) language. 

This is especially useful for new'd documents which have no proper file extension. The display language won't revert back to plaintext on every launch.


Teklad:
I'll check this out later. I was having issues with session files (even
saved files) reverting to plaintext when restoring session. I would imagine
this patch fixes all those quirks judging by the description. :p

JuBan1:
Yep, that was because modified documents are saved inside the cache with a random extension (to prevent name collision), and the DocEngine gets confused when loading that file and trying to figure out the proper language based on the extension.

danieleds:
Merging!

PR #303, opened 2016-10-28
Added Line Height as a modifiable property. Closes #300.

### LINKED ISSUE #300: Customizable editor line-height
I prefer the line-height to be set a bit higher.
I figured it's set in `/usr/share/notepadqq/editor/styles/app.css` in the `.editor > .CodeMirror` part, so I just edited that file.
But can you make it customizable in the settings, like the text size?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38739982-customizable-editor-line-height?utm_campaign=plugin&utm_content=tracker%2F346321&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F346321&utm_medium=issues&utm_source=github).
</bountysource-plugin>

JuBan1:
Allows the user to change line height the same way they can change font family or size.

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -400,15 +400,16 @@ UiDriver.registerEventHandler("C_CMD_SET_THEME", function(msg, data, prevReturn)
 UiDriver.registerEventHandler("C_CMD_SET_FONT", function (msg, data, prevReturn) {
     var fontSize = (data.size != "" && data.size > 0) ? ("font-size:" + (+data.size) + "px;") : "";
     var fontFamily = data.family ? ("font-family:'" + ('' + data.family).replace("'", "\\'") + "';") : "";
-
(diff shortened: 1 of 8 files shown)


danieleds:
Nice!

I noticed a small glitch when, in the preview editor, you select some text and then you change the line height: the selection doesn't refresh.

Actually, the same thing happens for the font family and font size too.
If you're able to fix it in a reasonable time, could you add the fix to this PR?

JuBan1:
It appears calling `editor->setLanguage(editor->language());` also updates the text selection. That's why the other editors work fine but this one doesn't. I've patched that in (ignore the deleted lines in the last commit- they're remnants of the borked commit 5 minutes earlier).

PR #305, opened 2016-10-30
Adds Count button to Find dialog. Resolved #304

### LINKED ISSUE #304: Show matches count when searching
Hello! First I want to thank you for the program, it's amazing :)

A small addition I'd love to see is displaying the "matches count" when searching something from the _Search dialog_. Or even a _Count_ button would be nice!

JuBan1:
This adds a Count button to the Find and Replace dialog. Clicking it runs the current search through a counting routine.
When thinking about how to display the count to the user I wanted to avoid a MsgBox for something as trivial. Instead, the button's labels changes when clicked:
image

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -201,6 +201,36 @@ UiDriver.registerEventHandler("C_CMD_SET_TABS_VISIBLE", function(msg, data, prev
     editor.refresh();
 });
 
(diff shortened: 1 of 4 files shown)


JuBan1:
Thanks. It'll probably be a while until repositories are updated and it's usable without rolling your own version. 

That is if the PR gets merged.

PR #314, opened 2016-11-06
Add support for string literals in replace.

### LINKED ISSUE #223: Search and replace: special character are used as literal in replace
In the search and replace (any mode), if I try to replace the matched text with a special character (\r, \n, \t, ... basically every character that need escaping) the replaced value will be replaced by literals ("\t" and not a tabulation for instance).

This issue might be associated with #191.
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/35531829-search-and-replace-special-character-are-used-as-literal-in-replace?utm_campaign=plugin&utm_content=tracker%2F346321&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F346321&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Teklad:
This is a really small change which basically hooks into SearchString::unescape and allows use of string literals for replacement text.  I believe this fixes #223.


danieleds:
Perfect :)

Issue #293, opened 2016-10-10
notepadqq fails to build on ArchLinux


IceflowRE:
qtchooser was removed from Arch.

```
==> Beginne build()...
Info: creating stash file /tmp/yaourt-tmp-iceflower/aur-notepadqq/src/notepadqq/.qmake.stash
cd src/ui/ && ( test -e Makefile || /usr/bin/qmake-qt5 /tmp/yaourt-tmp-iceflower/aur-notepadqq/src/notepadqq/src/ui/ui.pro PREFIX=/usr -o Makefile ) && make -f Makefile 
Project MESSAGE: Release build
make[1]: Verzeichnis „/tmp/yaourt-tmp-iceflower/aur-notepadqq/src/notepadqq/src/ui“ wird betreten
(cd "/tmp/yaourt-tmp-iceflower/aur-notepadqq/src/notepadqq/src/ui" && rm -rf "../../out/release/lib/../appdata/editor" && cd "../editor" && make DESTDIR="../../out/release/lib/../appdata/editor")
make[2]: Verzeichnis „/tmp/yaourt-tmp-iceflower/aur-notepadqq/src/notepadqq/src/editor“ wird betreten
mkdir -p "../../out/release/lib/../appdata/editor"
cp -r classes styles app.js index.html init.js "../../out/release/lib/../appdata/editor"/
mkdir -p "../../out/release/lib/../appdata/editor"/libs
cp -r libs/jquery "../../out/release/lib/../appdata/editor"/libs/jquery/
# Call CodeMirror makefile passing DESTDIR as absolute path, so we have no problems with cd-ing to libs
make -C libs -f Makefile-codemirror DESTDIR="/tmp/yaourt-tmp-iceflower/aur-notepadqq/src/notepadqq/out/release/appdata/editor"/libs/codemirror
make[3]: Verzeichnis „/tmp/yaourt-tmp-iceflower/aur-notepadqq/src/notepadqq/src/editor/libs“ wird betreten
mkdir -p "/tmp/yaourt-tmp-iceflower/aur-notepadqq/src/notepadqq/out/release/appdata/editor/libs/codemirror"
cd codemirror && cp -r addon keymap lib mode theme LICENSE "/tmp/yaourt-tmp-iceflower/aur-notepadqq/src/notepadqq/out/release/appdata/editor/libs/codemirror"/
# Clean unnecessary files within "mode"
[...]

PR #325, opened 2016-11-19
replace qtchooser with lrelease. fixes issue #293

realchrisolin:
This should be self-explanatory. In lieu of a more elegant way of fixing this, I opted for the "sledgehammer approach".

### DIFF
--- a/configure
+++ b/configure
@@ -1,8 +1,8 @@
 #!/bin/sh
 # Don't set `-e'!
 
-test -n "$QMAKE"    || QMAKE="qtchooser -run-tool=qmake -qt=5"
-test -n "$LRELEASE" || LRELEASE="qtchooser -run-tool=lrelease -qt=5"
+test -n "$QMAKE"    || QMAKE="qmake-qt5"
+test -n "$LRELEASE" || LRELEASE="lrelease"
 test -n "$MAKE"     || MAKE="make"
 test -n "$CXX"      || CXX="c++"
 
--- a/configure
+++ b/configure
@@ -1,7 +1,7 @@
 #!/bin/sh
 # Don't set `-e'!
 
-test -n "$QMAKE"    || QMAKE="qmake-qt5"
+test -n "$QMAKE"    || QMAKE="qmake"
 test -n "$LRELEASE" || LRELEASE="lrelease"
 test -n "$MAKE"     || MAKE="make"
 test -n "$CXX"      || CXX="c++"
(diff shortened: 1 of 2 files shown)


danieleds:
Thanks.

`qtchooser` is used as a default command in `./configure`... Don't you get an error when you run `./configure` too?

realchrisolin:
strangely, no. I'll try reasoning out why and comment back.

realchrisolin:
It's because I used the notepadqq-git AUR package, which doesn't run ./configure:

build() {
        cd "${_pkgname}"

         qmake-qt5 PREFIX=/usr notepadqq.pro

         make

}

Would you prefer I fix it accordingly in this PR before merging?

danieleds:
> Would you prefer I fix it accordingly in this PR before merging?

Yeah, thanks!

realchrisolin:
Done. I've only tested this on Arch (successfully)and can't guarantee these changes won't break something else on other distributions. However, they shouldn't as long as qmake-qt5 and lrelease are available.

realchrisolin:
Also, the Travis CI build appeared to have failed because the `./configure` command didn't include the `--prefix /usr` argument, which is noted in the readme.

configure
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
 # Don't set `-e'!
 
-test -n "$QMAKE"    || QMAKE="qtchooser -run-tool=qmake -qt=5"
-test -n "$LRELEASE" || LRELEASE="qtchooser -run-tool=lrelease -qt=5"
+test -n "$QMAKE"    || QMAKE="qmake-qt5"
```
danieleds: Put just the standard `qmake` here... then, if an alias is needed, users can specify it by manually setting the environment variable.

realchrisolin: Done, sorry for the delay.

danieleds:
Sorry for the delay.
Travis fails because, on Travis, "qmake" is using Qt 4 and without lrelease we don't have a way to reliably use Qt 5 (qmake-qt5 is not always available with that name).
For now, I'd say that the best workaround is the one specified [here](https://github.com/notepadqq/notepadqq/issues/293#issuecomment-252586049) unless there is some problem with it. If that is the case, let me know!

hasufell:
The suggested workaround doesn't work for me. Even when setting QMAKE and LRELEASE qtchooser is invoked.

realchrisolin:
Out of curiosity (and to help others that might run into the same issue), what distribution and release version are you using?

hasufell:
I've packaged it for the souce distro exherbo: 
https://github.com/hasufell/hasufell-repository/blob/master/packages/app-editors/notepadqq/notepadqq-1.0.1.exheres-0
http://git.exherbo.org/summer/packages/app-editors/notepadqq/index.html

Issue #331, opened 2016-12-23
Edit | Blank operations | Tab to space off by one


TopView:
Hi, there is a little problem with converting tabs to spaces.  Here is how to see it:

1) Open editor and click `Settings` | `Preferences` | `Languages` and set Tab stop to 4, and leave  `Use spaces instead of tabs` unchecked.

2) Then enter the following text, typing a tab key, in place of each "\<tab\>":

```
1234567890
abc<tab>ab<tab>a<tab>x

```
Then select `Edit` | `Blank operations` | `TAB to Space` 

Result:

```
1234567890
abc    bc    c    x

```

Expected results:

```
1234567890
abc ab  c   x    

```

What it's doing is always inserting 4 spaces for each tab, no mater what.  But this isn't how tabs work.

What it needs to do is insert from 1 to 3 spaces for each tab, depending on the current column.  For example,

If you are on column 1, then replace the tab with 3 spaces.
If you are on column 2, then replace the tab with  2 spaces,
If you are on column 3, then replace the tab with  1 space.

For comparison, see how Notepad++ works which does this correctly.

Thanks guys.  Great editor!

*I'm on Linux, Debian Jessie 8.6, w/ notepadqq version 0.53.1 (I hope that's the latest notepadqq version)*

JuBan1:
You're right, of course. This is certainly not how tabs are supposed to be handled!

TopView:
Thanks for your work on this.  It's educational to me to look at your 
commit.  I like that you have a few comments in it to explain it a bit.

I might have named *space*, *spaces*, because that's what it becomes, 
one or more, rather than just one.  I found it misleading to call that 
function space.

Similarly, what about calling tabToSpaceHelper simply tabToSpaces 
(plural)?  (In other words, I don't find the word "helper" to be of any 
help in understanding what is going on.)

Looking forward to having this work properly.

I like tabs, and have coded with them for many years w/ Notepad++ in 
Windows, but now having moved to Linux (because Win 10 is such a mess), 
I find that simple things that I used to rely on are not as advanced in 
Linux, so I'm really glad to see you jumping right on this.

Thanks a bunch!

JuBan1:
You're welcome.

I stuck to the naming convention already in the code, although pretty much all of it should probably be renamed to something more sensible. I'll probably amend that after the festivities.

[...]

Issue #330, opened 2016-12-22
blank operations  space to tab


wangyinhu:
blank operations  space to tab

only replace space to tab, witch change indent

PR #332, opened 2016-12-23
Fixed the tabs-to-spaces and spaces-to-tabs 

JuBan1:
Some of the blank operations mindlessly replaced tabs with spaces and vice-versa. This update changes the behavior to be more in line with what is expected.

This isn't perfect since mixed whitespace isn't handled. For example
`__\t__` (with _ being space and \t being a tab) won't be unified into `\t\t` or `________` because the code *only* replaces either spaces or tabs. Still, this is a step in the right direction.

Bug report: #331, #330


danieleds:
Thanks!

PR #340, opened 2017-01-16
Periodic Backups to prevent data loss. Fixes #266

### LINKED ISSUE #266: Autosave Featue
Looking to implement Autosaving next. Here's what I thought up:

* Autosave creates a file backup of changed files every X minutes. The Y most recent backups will be kept; they're labeled by date and users can select any of them if they want to restore a file from backup. 
* Users can choose the backup interval, number of backups kept, how long backups are kept, and whether to enable or disable autosaves completely.
* Autosave stores the file backups in the cache directory (like sessions) so directories aren't cluttered with .bak files.

But there are a number of things I'm unsure about:

* Should unsaved changes be autosaved? Basically, should be autosave be made from the document currently loaded or the underlying file?

* Likewise, should "orphaned" (new'd documents that weren't opened from a file) documents be autosaved?

* It's difficult to determine when the autosave backups should be deleted. If a user simply closes a file and doesn't ever open it anymore then the autosave files would remain. Best solution I came up with is to put a time-limit on the duration of the autosave files. So they'll be deleted after they are older than Z days.

* The autosave cache could possibly grow quite sizable. An extreme example: user wants to keep 10 last backups of several 1MiB+ size files. That could easily amount to 10s to 100s of MiB depending on how many documents they work on. Maybe add a label to the settings menu of how much space the autosave cache currently uses up? Maybe even an option to limit to cache to X MiB?

Thoughts?

JuBan1:
So this was a long time in the pipe.

The autosave functionality periodically inspects all open MainWindows and saves their current session into an autosave directory inside Nqq's .config folder. Right now the autosave interval is set to 5 seconds. Autosave is clever enough not to re-save a window if nothing has changed since the last time it was saved.

This process is completely transparent to the user. It's only use is that, if Nqq wasn't properly shut down, all windows, tabs, and their content will be restored from the autosave directory. This is useful since unsaved changes would otherwise be lost (the "Remember tabs on exit" feature won't work if Nqq or the system crashed).

Right now the feature is referred to as "Autosaving" pretty much everywhere in the code but it's probably not a good name for it since it doesn't actually *do* autosaving. I couldn't think of a good, short, descriptive text for Nqq's settings menu so I left it at "Enable Autosaving" for the time being. I'm open to good ideas.

A few commented debug statements/etc are still in the code; I'll remove them if/when the code is about to be merged. Right now they're pretty useful for me.

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -34,6 +34,10 @@ UiDriver.registerEventHandler("C_FUN_IS_CLEAN", function(msg, data, prevReturn) 
     return isCleanOrForced(changeGeneration);
 });
 
+UiDriver.registerEventHandler("C_FUN_GET_HISTORY_GENERATION", function(msg, data, prevReturn) {
+    return editor.getHistoryGeneration();
+});
+
 UiDriver.registerEventHandler("C_CMD_SET_LANGUAGE", function(msg, data, prevReturn) {
     Languages.setLanguage(editor, data);
 });
--- a/src/ui/EditorNS/editor.cpp
+++ b/src/ui/EditorNS/editor.cpp
@@ -204,6 +204,11 @@ namespace EditorNS
         sendMessage("C_CMD_MARK_DIRTY");
     }
 
+    int Editor::getHistoryGeneration()
+    {
+        return sendMessageWithResult("C_FUN_GET_HISTORY_GENERATION", 0).toInt();
+    }
+
     QList<QMap<QString, QString>> Editor::languages()
     {
         QMap<QString, QVariant> languages =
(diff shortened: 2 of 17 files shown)


JuBan1:
PS: This PR needs a small addition to codemirror.js which can be found here: https://github.com/notepadqq/CodeMirror/pull/15

danieleds:
Hi,
how is this different (or better) than just automatically saving the session each _n_ seconds?

JuBan1:
In essence it does save the session of each window every n seconds. In
addition to that it does two more things: it can detect when Nqq was
improperly shut down and automatically restores the backup session; and it
is clever enough not to save the session of a window if its contents has
not changed (so it doesn't perform useless disk writes every N seconds even
if you are away from your system).

Its purpose is mainly to provide some crash protection for users who depend
on Nqq's ability to remember unsaved changes.

Issue #203, opened 2016-04-06
Indentation message crash


darealshinji:
Starting NQQ for the first timer after a boot often takes a few seconds. When I double-click a file to open two times or more often, and the file's indentation doesn't match up, the indentation message will appear more than once (see screenshot). Clicking either "Use spaces" or "Ignore" then will make the app hang-up.
![indent-crash](https://cloud.githubusercontent.com/assets/5700937/14316661/9578b388-fc04-11e5-8ac7-4f0433be2fcc.png)
##

danieleds:
Okay, I've been able to reproduce it.
Thanks for the report!

danieleds:
Easily reproducible with this shell command:

    (notepadqq <filename> &); (notepadqq <filename> &)

Where `<filename>` is the path of a file which triggers the banner.

If it doesn't work on the first run, close notepadqq and try it again a few times

danieleds:
Apparently the bug has been inadvertitely fixed

PR #346, opened 2017-02-05
Clear banner-related signal prior to moving tabs.

### LINKED ISSUE #344: Notepadqq crashes when...
I know the title isn't as self-explanatory as it should, but It seemed difficult for me to describe the way it crashed...

Here's the way to reproduce it:

1. Open nqq
2. Open a file
3. Open a second file NOT indented with your default indentation settings (Just make the "indentation warning" dialog appear)
4. Move file to "Other View"
5. Now, click "Use spaces/tabs" (Whatever option appears to you) and prepare your data to be forgot :(

Luckily, It seems Nqq made a .bck file from my lost data. 
Unluckily, it contains nothing new (A exact copy of my last file before the modifications made with Nqq)

P.S = I haven't tried to click "Close" instead of using recommended indentation...
P.S 2 = Console output just says "Segmentation Fault"

JuBan1:
If a tab shows a banner and is then moved, `on_editorAdded()` is called for that tab. It creates a connection to delete the banner when the user clicks it away. Since this is not a newly created tab it already had such a connection from the first time it was added. Thus, `banner` is deleted twice.

The banner code is kind of a mess and probably deserves some cleaning up. But until then this is a simple fix.

Fixes #344, see this issue for reproduction.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -1096,7 +1096,7 @@ void MainWindow::on_editorAdded(EditorTabWidget *tabWidget, int tab)
     Editor *editor = tabWidget->editor(tab);
     
     // If the tab is not newly opened but only transferred (e.g. with "Move to other View") it may
-    // have a banner attached to it. We need to connect previous signals to prevent
+    // have a banner attached to it. We need to disconnect previous signals to prevent
     // on_bannerRemoved() to be called twice (once for the current connection and once for the connection
     // created a few lines below).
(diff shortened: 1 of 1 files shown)


danieleds:
Nice catch.
Could this also be related to #203 ?

JuBan1:
I can't replicate #203 with or without this fix. Maybe it's been fixed by some other patch.

danieleds:
Probably.
Merging, thanks.

PR #348, opened 2017-02-07
Minor cleanup of Banner messages

JuBan1:
I took a closer look at the way banners are handled after that last PR. 

A banner's deletion even is propagated through a few signals and ultimately ends up in a function inside `MainWindow`. These signals aren't used for anything else, and a good chunk of the call chain can be removed with no consequences by moving the deletion to the `removeBanner` function.

That also removes the need for that last band-aid fix.


danieleds:
The Editor class shouldn't delete the banner: it didn't create it. It doesn't know what it is and if it needs to stay alive.

The responsibility of deleting it when the time comes is of the one who actually created it: that's why we notify MainWindow.

I don't see anything bad with your fix ;)

PR #349, opened 2017-02-07
Minor visual changes to Banners.

JuBan1:
Right now a banner has a relatively high minimum width which sometimes messes with tab widths, especially when using several views.

Here's a patch that makes the banner message flexible by allowing it to word-wrap (and moves banner buttons to the right side).

I kept this separate from the #348 since one is internal and this one is an arguable visual change.

Preview Image

### DIFF
--- a/src/ui/EditorNS/bannerbasicmessage.cpp
+++ b/src/ui/EditorNS/bannerbasicmessage.cpp
@@ -25,11 +25,14 @@ namespace EditorNS
         m_message = new QLabel(this);
 
         m_message->setStyleSheet(".QLabel { margin-left: 5px; margin-right: 10px; color: white } ");
-        m_message->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::MinimumExpanding);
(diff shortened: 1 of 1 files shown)


danieleds:
Nice! That was needed.

Just one thing, can you keep the buttons close to the text? On large screens they become too distant.

danieleds:
In addition, on my system the text gets cut when the window is too narrow:

screen2

Font size is standard. Does it do the same on your system? At a first look, I don't get why it shouldn't work.

JuBan1:
> Font size is standard. Does it do the same on your system? At a first look, I don't get why it shouldn't work.

It's because only QLabel's bounding box can be controlled. If the text is too long to be shown in that box it'll just go over it. I don't think there's an easy way to limit how often QLabel does a line wrap.

I've given it a good bit of optional extra room now so it should be fine even for rather long banner messages. Aligned the buttons to the left as well.

danieleds:
Perfect!!

PR #350, opened 2017-02-12
Misc. improvements for Preferences menu

JuBan1:
Here's a small batch of changes:

* Preferences menu now has an "Apply" button in addition to "Ok" and "Cancel". That way users can immediately see visual changes (like themes) without having to exit the menu all the time.
* The Shortcuts tab now has a "Reset Selected Shortcut" in addition to the "Reset All Shortcut" button.
* The Preferences menu remembers the last selected settings tab and will automatically switch to it when re-opened (this is not a saved setting. Only implemented as a static variable).
* Renamed the object names for the tabs from something like "page_2" to more sensible names.

PS: This PR's diff looks a bit misleading because I moved around a function or two. The changes made are pretty self-contained.

Sample

### DIFF
--- a/src/ui/frmpreferences.cpp
+++ b/src/ui/frmpreferences.cpp
@@ -11,7 +11,9 @@
 #include <QInputDialog>
 #include <QTableWidgetItem>
 #include <QSharedPointer>
+#include <QDialogButtonBox>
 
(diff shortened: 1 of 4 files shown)


JuBan1:
> Code style

Switching between projects with different code style rules is hard ~~~

PR #351, opened 2017-02-12
Make status bar interactive & a few fixes.

JuBan1:
This one's about the status bar:

* Turned the "File Format" and "Encoding" labels in the status bar into ClickableLabels. When clicked, they open the Format/Encoding QMenu that is also used in Nqq's menu bar.
* File encoding in the status bar wasn't properly updated. Fixed that by calling updateEditorUiInfo() when switching encoding.
* Fixed the object names of all QMenus (e.g. menuSe_ttings to menu_Settings) so they're all named after the same scheme.

Edit: Apparently Qt Creator also auto-updated a bunch of QAction labels but it doesn't seem like this had any impact on anything. Leave it? Revert it?

Sample

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -350,7 +350,7 @@ void MainWindow::createStatusBar()
     layout->addWidget(label);
     m_statusBar_textFormat = label;
     connect(dynamic_cast<ClickableLabel*>(label), &ClickableLabel::clicked, [this](){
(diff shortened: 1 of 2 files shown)


danieleds:
Hmm those additional changes by Qt Creator may actually be nice. Let's keep them.

danieleds:
Fix spaces at line 353 and we're ready!

JuBan1:
Done.

PR #352, opened 2017-02-13
Added Slovenian translation

urmatej:
**Adding** notepadqq_sl.ts and **updating** src/ui/ui.pro and src/ui/resources.qrc


danieleds:
Good, merging.

PR #354, opened 2017-02-14
Fix handling "Reload" actions if no file to reload exists.

JuBan1:
When clicking 'Reload from disk' or 'Reload file interpreted as' if no file exists, nothing should be done since there is no file to reload from.


danieleds:
Can you show a dialog in both situations saying something like "This document is not saved to a file, so it can't be reloaded"?

Or better... can you disable those actions unless the current editor is linked to a file?

JuBan1:
Yes, that's better. I'll leave the check in `reloadWithWarning` in since it can potentially be called from other places too.

Issue #246, opened 2016-08-25
Some themes are broken in preview editor


danieleds:
In the settings dialog, some of the themes in the preview editor are broken the first time one tries to load them:

![](http://pix.toile-libre.org/upload/original/1472075759.png)

After that, they're fine.

---

Originally reported by @JuBan1 at #241 
##

PR #356, opened 2017-02-15
Fix theme loading bug in preferences by using onload for new styleshe…

Teklad:
Resolves issue #246


danieleds:
Merged: 037e8858f4c1dee5745e1f2fcc7852fd9ac134cd

PR #357, opened 2017-02-15
First implementation of cmake for notepadqq.

Teklad:
I got some free time and created a cmake build system for notepadqq.  Looking for input on what everyone thinks thus far.  It's mainly to get rid of some annoyances with the current build system yelling about qmake and lrelease paths on certain distros. (Thanks Arch Linux)

Criticism is welcomed, of course.


JuBan1:
Personally I'd rather stick to the current system because I know how it works and it plays very nicely with Qt Creator. I find it hard to believe there's isn't a good solution to these problems with the current system.

I "fixed" my issues on Arch permanently by writing `lrelease "${@: -1}"` in a text file, calling it qtchooser, and moving it to /sbin/. Not what you'd call optimal buuuutt ...

danieleds:
If the problem is just about qtchooser, I believe we can support a --no-qtchooser configuration flag.

darealshinji:
I like CMake, but I don't think it should entirely replace the current build system.

Teklad:
Yea, we could do that... but I do prefer that it just be able to build without needing to set so many flags.  Right now I have to set --qmake and --lrelease just to run the configure script. Manually deleting the PROJECT_DIR/out directory causes any future builds to fail as well with this error:
```
/usr/bin/uic mainwindow.ui -o ../../out/build_data/ui_mainwindow.h
Could not create output file
make[1]: *** [Makefile:1109: ../../out/build_data/ui_mainwindow.h] Error 1
make[1]: Leaving directory '/home/justin/Projects/notepadqq/src/ui'
make: *** [Makefile:44: sub-src-ui-make_first] Error 2
```

I'm not sure why this isn't handled by the current build system.... is it a qmake issue or a ./configure script issue?  My thoughts are the directories shouldn't have issues being re-created automatically.  It's just a minor annoyance having to run mkdir -p ./out/build_data...  

If we could handle these edge cases more appropriately in the current qmake setup I wouldn't mind... and maybe use something besides qt-chooser to detect qmake/lrelease in the event qtchooser isn't found or returns an error.  Its mostly trivial issues but I love for things to run really smoothly if at all possible. :P

CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+  message("git found: ${GIT_EXECUTABLE}")
+  if(EXISTS "${CMAKE_SOURCE_DIR}/.git")
+      message("Initializing submodules.")
+      execute_process(COMMAND "${GIT_EXECUTABLE}" submodule init)
+      execute_process(COMMAND "${GIT_EXECUTABLE}" submodule update)
+  endif()
+endif()
+
+
+# Installation instructions
```
hasufell: You shouldn't hardcode things like that. This will make it hard for distributors to package notepadqq correctly. Especially things like `install(TARGETS ${PROJECT_NAME} DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/${PROJECT_NAME})` further down.

Instead, use [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html)

Teklad:
That's actually really handy information... I'm semi-new at CMake so I'm 
not familiar with a lot of the ins and outs of how to do things.  Is 
there a better way to call git as well?

darealshinji:
> I would return to CMake system, to easier build with newer Qt (Qt6 prefer using CMake).

Meson is a neat build system too.

Issue #369, opened 2017-02-26
No proper source archives


hasufell:
Because this project uses submodules, downloading via git tags is broken. Your github releases only offer precompiled stuff, but not actual source archives containing the submodule source code.

darealshinji:
I agree, there should be a full source tarball among the releases. Until that happens however such tarballs could be downloaded from the PPA.

Teklad:
It wouldn't be hard to make this part of the release process via an automated script of some sort... If someone wants to release nightly tarballs I'm not against it.

hasufell:
https://github.com/Kentzo/git-archive-all might be a start which works like `git archive` except that it includes submodules

PR #371, opened 2017-02-27
travis: release source tarballs too including all submodules

darealshinji:
Should close https://github.com/notepadqq/notepadqq/issues/369


danieleds:
I think the items you added to the "script" section should be moved to before_deploy.

danieleds:
In addition, you already are in a checked out branch (submodules included). I'm not familiar with travis conventions but it might be better to avoid cloning it again and just clean & tar the current workspace. This also ensures that the code we're packing is the same as the code we're testing.

darealshinji:
I've updated the file.

PR #373, opened 2017-03-07
Webenginev1

Teklad:
This is a QWebEngine implementation of Notepadqq... currently in alpha status, but putting this pull request here for code inspection and what not.


Teklad:
So the good news is... this latest variant introduces a very large speed improvement over the previous.  There's still quite a few loose ends to tie up... but things are progressing really well now.

I opted to go for "hooks" in UiDriver that can be placed in the appropriate function so we can fine tune all the calls we need.  This makes UiDriver a bit larger but it keeps actual editor modifications in app.js and communication in UiDriver.js.

Most of the recent changes have been to improve performance and stability.  Startup time is back down to reasonable levels and new tabs are created in ~50ms or less after Notepadqq is fully initialized.  There aren't too many calls that actually block at this point, so that's a big plus as well.

Certain things still need to be re-implemented(i.e. printing)... but for the most part the Editor is almost back up to speed in terms of functionality with where it was at before.... and according to @JuBan1 it seems to fix some odd issues regarding text dragging and stuff that was present in QWebKit.

JuBan1:
Issues remaining:

* White flash when opening new tab.
* Clean/Dirty recognition is screwed
* * Opening a new tab and closing it will ask whether changes are to be saved
* * Pressing CTRL+Z on an opened document erases all its contents.

Teklad:
Yea... the white flash is something I'm not sure how to fix... at least not unless we can force a draw before the widget is rendered.

Clean/dirty should be an easy fix... just need to handle it how it was originally handled, which for some reason I neglected when adding the hooks.

Teklad:
So QWebEngine in all it's glory decided to provide us with its own undo trigger... "yay"... Need to disable that :|

Teklad:
Well I think that wraps up the issues with clean state... maybe.

src/ui/include/EditorNS/jsproxy.h
```diff
@@ -0,0 +1,97 @@
...
+        QVariant getResult();
+        QVariant getCursor();
+        QVariant getTextLength();
+        QVariant getLineCount();
+        QVariant getSelections();
+        QVariant getSelectionsText();
+        QVariant getScrollPosition();
+        QVariant getLanguage();
+        QVariant getClean();
+        QVariant getDetectedIndent();
```
JuBan1: Instead of `QVariant m_detectIndent` and `QVariant getDetectIndent()`, can't we directly use `bool m_detectIndent` and `bool getDetectindent()`? I'd rather do the conversion from QVariant to the specific data type as close to the "source" as possible.

Teklad: Probably... in most cases it's not possible due to limitations in QWebChannel... but I doubt it would have trouble with bool... let me adjust that.

src/ui/include/EditorNS/editor.h
```diff
@@ -290,9 +257,11 @@ namespace EditorNS
         void setSelection(int fromLine, int fromCol, int toLine, int toCol);
 
         int lineCount();
+        int textLength();
```
JuBan1: While we're at it, should we use `getLineCount()` or just `lineCount()` for getters? Because right now we're rather inconsistent with that. @danieleds?

IMO, `getLineCount()` is better since it plays very nicely with autocomplete.

src/ui/include/EditorNS/jsproxy.h
Teklad: detectIndent passes other data outside of just bool, unfortunately.... but I can make m_clean a bool instead.

src/ui/include/EditorNS/editor.h
Teklad: I prefer get/set as well.... it just looks cleaner in general.

src/ui/EditorNS/editor.cpp
```diff
@@ -656,7 +708,7 @@ namespace EditorNS
     void Editor::print(QPrinter *printer)
     {
         sendMessage("C_CMD_DISPLAY_PRINT_STYLE");
-        m_webView->print(printer);
+        //m_webView->print(printer); // FIXME
```
JuBan1: There is now print support for WebEngineView:

[void QWebEnginePage::print(QPrinter *printer, FunctorOrLambda resultCallback);](http://doc-snapshots.qt.io/qt5-5.8/qwebenginepage.html#print)

However, it's only starting at Qt5.8 so we'd have a rather high Qt requirement if we're to use it.

Teklad: We could use a compiler IFDEF for that feature for the moment.

src/ui/include/EditorNS/jsproxy.h
JuBan1: Was just an example. `getDetectindent` can return a `QPair<bool, int>` for example.

My motivation is to limit passing around QVariants or QVariantMaps as much as possible by doing the conversion to "proper" data types as soon as possible. 

That way none will have to make the mistake I did and assume that DetectIndent's QVariant is a bool value.

Teklad:
So to add to the list of broken things... banners aren't currently working.  This shouldn't be too tough to fix, I don't think.

src/ui/include/EditorNS/jsproxy.h
Teklad: I'll see if it will perform the conversion without any hassle(hopefully).... I don't like passing around QVariant either as it makes things pretty muddy.

Teklad: I'll just remove the Q_PROPERTY stuff and use slots for setting the stuff so we can do conversion within JsProxy.... this will make things simpler in the long run.

src/ui/include/EditorNS/editor.h
danieleds: I agree with the get/set convention

src/ui/EditorNS/editor.cpp
```diff
@@ -148,9 +173,13 @@ namespace EditorNS
...
-            QVariantMap map = data.toMap();
-            emit currentLanguageChanged(map.value("id").toString(),
-                                        map.value("name").toString());
+            QVariantMap lang = data.toMap();
+            QString id = lang.value("id").toString();
+            QString name = lang.value("name").toString();
+            if (!m_customIndentationMode) {
+                setIndentationMode(id);
+            }
+            emit currentLanguageChanged(id, name);
```
danieleds: For some reason the toolbar now shows the internal language name instead of the friendly one (e.g. "csharp" instead of "C#")

danieleds:
Created project here: https://github.com/notepadqq/notepadqq/projects/1

src/ui/EditorNS/jsproxy.cpp
```diff
@@ -0,0 +1,116 @@
...
+QVariant JsToCppProxy::getResult()
+{
+    return m_result;
+}
+
+QVariant JsToCppProxy::getCursor()
+{
+    return m_cursor;
+}
+
```
danieleds: I'm not a fan of these hardcoded methods. Can't we use something like a generic map instead of multiple fields like these?

So that we can remove all these methods from here and move them in Editor, like this (pseudocode):

```
Editor::getResult()
 {
     return proxy->values.get("result");
 }
 
Editor::getCursor()
 {
     return proxy->values.get("cursor");
 }

Editor::getField(string field)
 {
    return proxy->values.get(field);
 }
```

That's important for future extensibility with extensions.

src/ui/EditorNS/editor.cpp
```diff
@@ -206,10 +236,27 @@ namespace EditorNS
 
     QList<QMap<QString, QString>> Editor::languages()
     {
+        // TODO: Cache this data.
+        QJSEngine engine;
+        QFileInfo fileInfo(Notepadqq::editorPath());
+        QString fileName = fileInfo.absolutePath() + "/classes/Languages.js";
```
danieleds: Why not do it in the old way just the first time, and then cache it for all the following times?
It would have the advantage to remove the dependency on qml and to avoid introducing assumptions about the javascript code (which should remain an abstraction).

Teklad: This method actually increases startup time by a margin... waiting for the entire web browser to be loaded before getting the languages doesn't make sense (Values that never change shouldn't require UI to be loaded to be retrieved).

It just seemed like a cleaner way to do it... I'm not against changing it, but in essence this method isn't crippled by requiring a widget to be initialized.

src/ui/EditorNS/jsproxy.cpp
Teklad: Think you can figure out a way to do the type conversion within the proxy still?   I would much prefer your method, but I don't want type conversion happening within the editor if we can help it.

src/ui/EditorNS/editor.cpp
Teklad: If we could find a way to parse the Languages.js file on build and create a readable config file for languages... this would be the optimal method... It would probably be faster than full JS parsing as well.  Alternatively, we could maintain the file manually, but that adds developer overhead.

src/ui/EditorNS/jsproxy.cpp
Teklad: Could we do something like this?

bool JsProxy::value(&T, "cursor");

It would be a bool function that returned false on failure to set the variable data... or true if we successfully grabbed the information we needed.

This would still allow the conversion to happen within JsProxy, but it would be flexible in that it would accept any type as the first argument and get the data without too much hassle.

Teklad:
There's a few remaining bugs/features and I'm sure some more performance improvements could be made... but it should be in a good enough state at this point to where it can be merged into its own branch.... that way I can start pushing smaller patches instead of one large fat PR like I'm doing now, lol.

It's stable enough that it doesn't crash or do anything which is detrimental to editor functions... though I would like to add a few more checks here and there to ensure if something goes wrong it does so gracefully.... just to be safe.

danieleds:
Is "material" theme working for you? A lot of themes seems to be broken here.

danieleds:
In addition, the preview editor in the settings dialog doesn't display syntax highlighting anymore.

notepadqq.pro
```diff
@@ -1,4 +1,4 @@
 TEMPLATE = subdirs
-SUBDIRS = src/ui \
-    src/ui-tests
+SUBDIRS = src/ui
+#    src/ui-tests
```
danieleds: Why has this been commented?

src/ui/mainwindow.cpp
```diff
@@ -1675,13 +1673,14 @@ void MainWindow::on_documentLoaded(EditorTabWidget *tabWidget, int tab, bool was
 
     if (!wasAlreadyOpened) {
         if (m_settings.General.getWarnForDifferentIndentation()) {
-            checkIndentationMode(editor);
+//            checkIndentationMode(editor);
```
danieleds: Why this change?

src/ui/include/EditorNS/editor.h
```diff
@@ -181,19 +147,22 @@ namespace EditorNS
...
-        QList<QMap<QString, QString> > languages();
+        static QList<QMap<QString, QString> > languages();
 
         /**
          * @brief Set the language to use for the editor.
          *        It automatically adjusts tab settings from
          *        the default configuration for the specified language.
          * @param language Language id
          */
+        Q_INVOKABLE QString getLanguage(const QString& val = "id");
```
danieleds: The doc is not coherent with the method

notepadqq.pro
Teklad: It doesn't compile on my end with that line enabled.

src/ui/mainwindow.cpp
Teklad: This one fires off before the editor is ready.... even with waitAsyncLoad

src/ui/include/EditorNS/editor.h
Teklad: Will fix shortly. :P

Teklad:
Been meaning to revert the local storage change.... I realized these issues earlier as well..... it would be a better change for another time.

JuBan1:
> In addition, the preview editor in the settings dialog doesn't display syntax highlighting anymore.

I think this is just a side-effect of the localStorage change. You have to change the theme once, maybe restart Nqq and it works. But sinceTeklad's gonna revert that it'll solve itself soon.

> Is "material" theme working for you? A lot of themes seems to be broken here.

Can confirm. Is borked.

notepadqq.pro
JuBan1: Does this actually break anything? I still get differentIndendation warnings in tabs.

Teklad:
QWebEngine broke my previous fix for the preview theme window... So that's going to go back to being odd...  Not much I can do about it until I brush up on some JS goodness.  For now we'll revert it and leave it be until we get this merged... I would like to have it fixed before pushing this to master branch though... Regressions annoy me to death.

Teklad:
Notepadqq.pro causes a compilation error on my end with that line enabled.  I'll post the error when I get to my computer later

JuBan1:
ui-tests just failed because its modules weren't updated https://github.com/Teklad/notepadqq/pull/7

Teklad:
Think someone can figure out why Travis builds are failing? Lol.

JuBan1:
> WARNING: Project ERROR: Unknown module(s) in QT: webchannel webenginewidgets

Poor Travis doesn't have WebEngine. Try https://github.com/Teklad/notepadqq/pull/8. I don't know jack about Travis but this seems easy enough.

danieleds:
@Teklad can you git revert the commit d076a9aa99951764a8b538cacc8e0d4871c6b785? I understand that the banner will be broken but I'd like to discuss that changeset later. I'll reopen the issue.

Teklad:
@danieleds Just the indentation banner portion?  The rest pertains to JsProxy I believe.

danieleds:
Forget it, don't revert.
The issues are:

 * `connect(editor, &Editor::fileLoaded, this, &MainWindow::checkIndentationMode);`
    the slot should be called `on_fileLoaded`, which is a method which then calls `checkIndentationMode(editor)`

* the current implementation of `checkIndentationMode` is inconsistent with the old behaviour, in fact you have left an empty "if" where you removed the old call to `checkIndentationMode()`. I suggest you add those checks within `on_fileLoaded`.

Teklad:
I seriously need to start remembering to commit in smaller changesets so I can stop manually reverting.

src/ui/include/EditorNS/editor.h
```diff
@@ -257,6 +234,7 @@ namespace EditorNS
          *        that needs to be used when the document gets saved.
          * @param codec
          */
+        QTextCodec *codec() const;
```
danieleds: Doc is incoherent

danieleds:
@Teklad I'll fix Travis by using Docker. We can't do anything else because qtwebengine isn't in Ubuntu Trusty.

Did you see my last comment? https://github.com/notepadqq/notepadqq/pull/373#issuecomment-285493624

src/ui/include/EditorNS/editor.h
```diff
@@ -265,17 +243,18 @@ namespace EditorNS
...
-
-        QList<Selection> selections();
-
+ 
         /**
          * @brief Returns the currently selected texts.
          * @return
          */
-        Q_INVOKABLE QStringList selectedTexts();
+        Q_INVOKABLE void setSelectionsText(const QStringList &texts, selectMode mode);
```
danieleds: Doc is incoherent

Teklad:
I did not... I'll let you handle travis though.... it's just making me angry, lol.... Let me revert my commit for indentationMode and fix it appropriately.

src/ui/include/EditorNS/editor.h
Teklad: While we're on the subject... would you be against prepending "get"  to these functions so the API is a little more consistent?  Right now its kind of back and forth depending on which part of Notepadqq's source you're looking at... so I'd like to have a standard for this.

danieleds: I'm strongly ok with prepending "get".
In the next PR though :P

Teklad: Yea... that's what I meant... just going to get this ready for merging at this point... no more changes unless requested, lol.

Teklad:
Holy cow almost 100 commits in this PR?  I really lose track of how much I program sometimes. ;_;

Teklad:
Well we're at the final stretch finally in terms of getting QWebEngine working... this was a pretty big change compared to a lot of previous PRs.... I didn't expect QWebEngine to be so vastly different from QWebView in terms of how it needs to be used.

danieleds:
@Teklad can you manually replace .travis with this exact content?

```
language: cpp
dist: trusty
sudo: required
os:
  - linux
env:
  - NQQ_VERSION=${TRAVIS_TAG:-${TRAVIS_COMMIT}}
before_install:
  - sudo add-apt-repository ppa:beineri/opt-qt58-trusty -y
  - sudo apt-get -qq update
  - sudo apt-get install -y qt58base qt58webengine qt58svg qt58tools; source /opt/qt58/bin/qt58-env.sh;
script:
  - "./configure --qmake qmake --lrelease lrelease && make"
  - "./src/ui-tests/ui-tests"
before_deploy:
  - "mv out/release out/notepadqq-${NQQ_VERSION}-linux-x64 && tar -zcf notepadqq-${NQQ_VERSION}-linux-x64.tar.gz -C out notepadqq-${NQQ_VERSION}-linux-x64"
deploy:
  provider: releases
  api_key:
    secure: OCEuXR1a2dDHKKQpnOeSoeWZeL0clf0eVUZt5QbrnkyvL3/iUHXjYbdVBEGrbuldffLy9YOC63viqh/r7Ak+7k/suG1BxwGczbvtMBvRyUOOI48YX90Ss+JUcpvJVNLoirDqA1hPMho5Rmydu+DFRlpjADfg3KtZZZA+8Xz8Uek=
  file: "notepadqq-${NQQ_VERSION}-linux-x64.tar.gz"
  skip_cleanup: true
  on:
    repo: notepadqq/notepadqq
    tags: true
```

src/editor/classes/UiDriver.js
```diff
@@ -1,31 +1,160 @@
...
+    var handlers = [];
+
+    var socket;
+    var channel;
+    var msgQueue = [];
+
+    var _this = this;
+
+    function usingQtWebChannel() {
+        return (typeof cpp_ui_driver === 'undefined')
```
danieleds: Is this function obsolete? And its uses too.

src/editor/index.html
```diff
@@ -155,7 +156,6 @@
     <script src="libs/codemirror/addon/show-invisibles/show-invisibles.js"></script>
 
     <script src="init.js"></script>
```
danieleds: init.js should be moved just before UiDriver.js

src/editor/classes/UiDriver.js
```diff
@@ -1,31 +1,160 @@
...
+    }
+
+    // Hook for when we load a file/change content
+    this.onSetValue = function(editor) {
+        var detectedIndent = this.detectIndentationMode(editor);
+        if (detectedIndent !== undefined) {
+            this.proxy.setValue("detectedIndent", detectedIndent);
+        }
+        editor.clearHistory();
+        this.proxy.sendEditorEvent("J_EVT_FILE_LOADED", 0);
```
danieleds: "J_EVT_FILE_LOADED" is misleading: I suggest "J_EVT_DOCUMENT_LOADED".
Same thing for the associated events and slots.

src/ui/mainwindow.cpp
```diff
@@ -1672,15 +1670,20 @@ void MainWindow::on_documentLoaded(EditorTabWidget *tabWidget, int tab, bool was
...
         updateRecentDocsInMenu();
     }
+}
 
-    if (!wasAlreadyOpened) {
-        if (m_settings.General.getWarnForDifferentIndentation()) {
-            checkIndentationMode(editor);
+// TODO: Make editor aware if a file is already loaded so
+// indentation banner only appears for new files.
+void MainWindow::on_fileLoaded(bool wasAlreadyOpened)
```
danieleds: regression?

src/editor/index.html
Teklad: It won't really matter where it's moved to at this point... everything is asynchronous (and I mean everything).... I'll go ahead and move it though, lol.

src/editor/classes/UiDriver.js
Teklad: Yes it is... its part of QWebKit's old compatibility.... should be able to get rid of it.

src/editor/index.html
danieleds: It's for semantic reasons. Loading and execution order of js code is *not* asynchronous, and "init.js" is meant to be loaded first :)

src/ui/mainwindow.cpp
Teklad: Unfortunately yes.... we're currently pretty heavily tied into how DocEngine functions... any fix I can come up with "quickly" will be hacky at best... so its best to leave that alone for another PR.

danieleds: Ok, let's open an issue then so we don't forget about it

Teklad:
Should we advertise testing the webengine version?  It would probably help greatly in finding bugs before the final release. (Assuming a few users participate)

danieleds:
We could maybe package a portable beta version. Not sure yet.

src/ui/include/EditorNS/jsproxy.h
```diff
@@ -0,0 +1,56 @@
...
+        bool getValue(const QString&, QString&);
+        bool getValue(const QString&, QStringList&);
+        QVariant getRawValue(const QString&);
+        // Functions to allow the proxy to set data on the CPP side.
+        void setResult(QVariant data);
+        void sendMsg(const QString& msg, const QVariant& data);
+        Q_PROPERTY(QVariant result READ getResult WRITE setResult NOTIFY replyReady);
+    public slots:
+        Q_INVOKABLE void sendEditorEvent(QString msg, QVariant data);
+        Q_INVOKABLE void setValue(QString name, QVariant data);
```
danieleds: Can you provide docs for all these methods? Explaining in depth what they do.

And for the signals too.

A general description at the beginning of the class explaining what it does and how it should be used would be great too.

src/ui/mainwindow.cpp
```diff
@@ -1118,6 +1116,7 @@ void MainWindow::on_editorAdded(EditorTabWidget *tabWidget, int tab)
     // created a few lines below).
     disconnect(editor, &Editor::bannerRemoved, 0, 0);
     
+    connect(editor, &Editor::documentLoaded, this, &MainWindow::on_fileLoaded);
```
danieleds: shouldn't this be `on_documentLoaded`?

src/editor/app.js
```diff
@@ -22,12 +27,14 @@ function isCleanOrForced(generation) {
 UiDriver.registerEventHandler("C_CMD_MARK_CLEAN", function(msg, data, prevReturn) {
     forceDirty = false;
     changeGeneration = editor.changeGeneration(true);
-    UiDriver.sendMessage("J_EVT_CLEAN_CHANGED", isCleanOrForced(changeGeneration));
+    UiDriver.proxy.clean = isCleanOrForced(changeGeneration);
+    UiDriver.proxy.sendEditorEvent("J_EVT_CLEAN_CHANGED", isCleanOrForced(changeGeneration));
```
JuBan1: Right now using `JsToCppProxy::getValue("some key")` does not work for clean events and maybe others. 

The problem is that lines such as this one send the `"J_EVT_CLEAN_CHANGED"` signal which is saved in the JsProxy's hash map. However, when querying whether the editor is clean, we generally use `"clean"` (see `Editor::isClean`).

As to fix that.. the most basic fix would be to just send the `"clean"` event alongside the  `"J_EVT_CLEAN_CHANGED"` event. But maybe it would be a good idea to split this into a `sendEditorEvent()` and a `setEditorData()` to distinguish between events and just setting data? Or maybe JsProxy should just react on ` `"J_EVT_CLEAN_CHANGED"`` events and set `"clean"` to the proper value? Then we wouldn't have to send any signals like `clean` or `language` anymore.

Right now it's a bit of a mess me thinks.

src/editor/classes/UiDriver.js
```diff
@@ -1,31 +1,153 @@
...
+        this.onCursorActivity(editor);
+        this.onChange(editor);
+        this.onScroll(editor);
+        this.onOptionChange(editor);
+        this.proxy.sendEditorEvent("J_EVT_READY", 0);
+        editor.refresh();
+    }
+    // END HOOKS
+}
+
```
danieleds: `detectIndentationMode`, `onLanguageChange`, `onSetValue`, `onCursorActivity`, `onChange` and `onScroll` shouldn't be here.

`onOptionChange` probably neither.

This file should only contain abstract communication functions and shouldn't be aware of any details about how our Editor works.

Let's discuss a better place to put these functions.

Teklad:
We already have a signal with that name I think... As for the other
issues... I'm not sure where to put the UiDriver event hooks... I mean we
could cram it all back in app.js but I don't know how good of an idea that
would be.

The rest of the stuff is trivial and shouldn't be too bad.  Sorry about the
code mess it was an interesting experience getting it to work to say the
least

danieleds:
Ahah no problem, I just want to be sure to merge the best code that we can get.

Instead of `on_fileLoaded` then maybe `on_editorValueUpdated` would be better.

Teklad:
Just keep in mind this won't be in master for a bit so we'll have time to
fix it up quite a bit before release ;). I do want it to be in a bit better
shape before its merged at all though. Any clue where to put the JS event
hooks?

danieleds:
That's for sure!

Not sure yet about the hooks. Move them within a "EditorStatusEvents.js" static class for now. This is a temporary solution.

In addition you seem to use `changeGeneration` within the `onChange` hook, but `changeGeneration` is undefined there.

Teklad:
Its defined by the time its called... Might be best to pass it as an
argument though just to be on the safe side... Gonna use some fancy JS if
I'm making my own class though.  Lots of new stuff we can use in QWeEngine.
:)

src/editor/app.js
Teklad: Its actually just there as a backwards compatibility with the old QWebKit events in Editor.  I need to spend some time actually reworking how events are handled in general.... and yes, setting the data when the event is sent would be a good call as well... I doubt a few extra signal/slots is gonna kill JsProxy anyhow.

Teklad:
I turned UiDriver and the event thingy you asked for into javascript classes for better control of instancing...  I "think" everything still works... I had the event hooks plugged in there pretty good, lol...

Might be a good idea to make the proxy itself its own class as well in the future.  I got 2 more days of work then I can get back to finishing this full force.

My current list consists of:
* A lot of JS clean up.
* Documenting functions
* Making events push the data we're gonna use instead of pushing the data and and an event separately.

These few things should clear up a lot of the current mess... I'm also probably going to put the editor.on() calls in the constructor of the event handler class so we don't pollute app.js more than necessary.

Hopefully everything goes smoothly so we can get this merged and get to the "bug hunting" phase...  I'm sure there's quite a few lingering nasties that hasn't been caught from the switch... but they'll be hard to root out without heavy testing.

danieleds:
Themes are broken again

src/ui/EditorNS/jsproxy.cpp
```diff
@@ -39,6 +39,10 @@ void JsToCppProxy::sendEditorEvent(QString msg, QVariant data)
     if (msg == "J_EVT_READY" && m_ready == false) {
         m_ready = true;
         pushQueuedMessages();
+    }else if(msg == "J_EVT_CLEAN_CHANGED") {
+        m_values["clean"] = data;
+    }else if(msg == "J_EVT_CURRENT_LANGUAGE_CHANGED") {
+        m_values["language"] = data;
```
danieleds: J_EVT_CLEAN_CHANGED and J_EVT_CURRENT_LANGUAGE_CHANGED don't belong here because we don't want to know about all possible events here. J_EVT_READY was only a special case.

There's still room for improvements but I actually prefer how it was done before... you can revert this commit (fcd8dd022729615d94f7d81eb22748e9f15b19e6)

src/ui/EditorNS/jsproxy.cpp
```diff
@@ -0,0 +1,115 @@
...
+void JsToCppProxy::sendEditorEvent(QString msg, QVariant data)
+{
+    if (msg == "J_EVT_READY" && m_ready == false) {
+        m_ready = true;
+        pushQueuedMessages();
+    }else if(msg == "J_EVT_CLEAN_CHANGED") {
+        m_values["clean"] = data;
+    }else if(msg == "J_EVT_CURRENT_LANGUAGE_CHANGED") {
+        m_values["language"] = data;
+    }
```
danieleds: J_EVT_CLEAN_CHANGED and J_EVT_CURRENT_LANGUAGE_CHANGED don't belong here because we don't want to know about all possible events here. J_EVT_READY was only a special case.

There's still room for improvements but I actually prefer how it was done before... you can revert this commit (fcd8dd022729615d94f7d81eb22748e9f15b19e6)

danieleds:
I would also like to see, if possible, a UML diagram explaining how the communication works right now. I'm not sure I totally get it.

JuBan1:
> Themes are broken again

This is something to do with applying themes at start-up. Re-applying themes through the Preferences setting makes them look good again.

You said 'again'. Was that behavior fixed for you in a previous commit? (I haven't paid attention to that).

danieleds:
> You said 'again'. Was that behavior fixed for you in a previous commit? (I haven't paid attention to that).

They seemed to be working here: ed6ce731bdffd98a1ccb0982e9ab99dc9ab72121

danieleds:
git bisect tells me the bug was reintroduced in f5bae20974267d54422460f1a37bf46cca511cbd:

```
f5bae20974267d54422460f1a37bf46cca511cbd is the first bad commit
commit f5bae20974267d54422460f1a37bf46cca511cbd
Author: Teklad <jkinnaird1991@gmail.com>
Date:   Thu Mar 9 17:11:57 2017 -0600

    Address some more change requests.

:040000 040000 03a672a20c164b46d8b65454f169ed6442e697bb ca4a3fcf9edff112d9ccf2a98d4a00940bbaf449 M	src
```

JuBan1:
Thanks- that makes it very obvious why.

Turns out moving `init.js` up that far doesn't work. `codemirror.css` needs to be loaded before `init.js`.

danieleds:
Sorry, didn't see your PR :/

danieleds:
Now I'm finally going to merge the most huge PR of history :)

I'll do the merge into a `webengine-beta` branch. From there we can focus on smaller tweaks and fixes.
Please open a different PR for each improvement.

Issue #390, opened 2017-03-12
QtWebEngine: Imrpove handling of events and data across the Js/C++ boundary


JuBan1:
With the revert of 03393a5 for webengine-beta we're back at the topic of handling signals and value updates. This should probably be somewhat high up on the TODO list.

danieleds:
For this issue I'd like to forget about the code and discuss what would be the best way to handle this communication.

I'd like to start with a UML diagram of the current status of the architecture.

Teklad:
I'm at work... And I suck with UML diagrams. But the communication
currently works like this.

CppToJS:
- editor invokes jsproxy via sendmsg
- jsproxy emits a signal to communicate with QWebChannel
- QWebChannel catches the signal and invokes received message on the JS
side.
- Depending on the message and data, JavaScript will set a data value in
JsProxy and emits a signal notifying of a data value change.

JsToCpp:
- JS editor event handler hooks to editor.on functions and pushes data as
it changes back to jsproxy via set value and sendEditorEvent...

It might be better to extend sendEditorEvent to where we can pass an
m_value key to modify if its needed.. That would keep from redundant extra
calls to setValue and keep our CPP code clean.

danieleds:
Here are some sequence diagrams about what happening right now, if I've got them correctly:

![c sends a blocking message to javascript](https://cloud.githubusercontent.com/assets/402652/23834345/70c9a332-0755-11e7-9201-109ffaaf4c93.png)
![c sends a command message to javascript](https://cloud.githubusercontent.com/assets/402652/23834346/70cbe494-0755-11e7-82bd-5fbcb5c81f2b.png)
![c wants to get a value from javascript](https://cloud.githubusercontent.com/assets/402652/23834347/70d2fdd8-0755-11e7-9604-7cf74ce03c2c.png)
![javascript sends a message to c](https://cloud.githubusercontent.com/assets/402652/23834348/70d83eb0-0755-11e7-963a-9aa90c6dc66a.png)

The last one is the only one which bugs me. I don't like the fact that the poor app.js is sending values to JsProxy even in case no one will ever care about them.

What about an async API on the c++ side?

Teklad:
Well it mainly sends stuff that's acted upon immediately... technically if we were to fix our signals/slots to actually use the data rather than query Editor every time, everything would go directly where it needs to go without requiring storage.

[...]

PR #391, opened 2017-03-13
QtWebEngine: Signal/Slot enhancements

Teklad:
This is the first batch of changes related to #390 ... this one deals with improving how we send signals from the editor, as well as segmenting out refreshEditorUiInfo into smaller pieces where we can.

Notable Changes:
* refreshEditorUiCursorInfo has been modified and placed in on_cursorActivity directly.  We send a struct with the relevant data to this slot and act on it there.
* Manually requesting cursor info for the UI can be done via editor->on_cursorInfoRequest
* refreshEditorUiInfoAll was added temporarily... but I may merge all the calls back into refreshEditorUiInfo so that can be our one-and-done function for updating all UI (eg. new documents, tab change)
* Various signals and slots to transport the data as needed.

Another thing I did was pull some of the stuff out of on_proxyMessageReceived and placed them into less ugly generate[event]Signal()... which creates the struct object containing the data and emits the proper signal.

I made various different JS modifications to help with performance... onCursorActivity was firing fast enough that it was slowing down selection rendering on my end... so I placed a small setTimeout() with a 20ms delay which handles fast changes a lot better.

Misc. modifications include Javascript EditorEventHandler::cursorActivityObject, which generates the Json object to send back to C++.... as well as storing m_clean within the editor when the signal passes through.  On top of that, I set up J_EVT_CLEAN_CHANGED to only fire when the clean state actually changes.

All of the above appear to have helped with oddities regarding clean state of the editor. (I was still getting random "Would you like to save" dialogs when loading sessions and closing a tab.

Overall, I think this is quite a bit better for event handling... and gets rid of the need for JsProxy to store most of its values.

This should compliment asynchronous calls really well.


src/ui/editortabwidget.cpp
```diff
@@ -24,6 +24,7 @@ EditorTabWidget::EditorTabWidget(QWidget *parent) :
 
 EditorTabWidget::~EditorTabWidget()
 {
+    this->blockSignals(true);
```
Teklad: We don't need to receive signals from this when its destructor is being called.... it causes issues regarding pointer validity(somehow)... and causes a crash.

Teklad:
@JuBan1 can you give this a lookover and tell me what you think about the changeset?  I tried to keep it fairly minimal and direct...

JuBan1:
I like the direction this is going in. Grouping related data into structs like `UiCursorInfo` is good, and JS being able to directly call slots to submit data is better than having to use `"C_CMD_STRING_IDENTIFIERS"`.

I'm not sure if we need the extra stuff inside `onCursorActivity`. Does the timer actually improve anything? We've never had trouble with moving the cursor, and adding a new time-dependent component to that mix if we don't have to sounds like a recipe for subtle tiny bugs later down the road.

Minor bug I spotted:  `refreshEditorUiInfoAll` doesn't update the character/line count in the status bar.

Teklad:
It mainly keeps the cursor event from hosing javascript... running that 
javascript in quick succession causes slowdowns on my end (My processor 
isn't the greatest)... so it lowers hardware power requirements just a hair.


When we want accurate live data... we'll be able to use callbacks for 
that... but since it was simply updating the GUI... being ultra-realtime 
like an FPS shooter seemed a little excessive.

src/ui/EditorNS/editor.cpp
```diff
@@ -152,43 +151,65 @@ namespace EditorNS
...
+        //     and getLineCount?  I don't think they're used outside of
+        //     refreshEditorCursorUiInfo.
+        UiCursorInfo info;
+//        info.charCount = v["charCount"].toInt();
+//        info.lineCount = v["lineCount"].toInt();
+        info.cursorLine = v["cursorLine"].toInt();
+        info.cursorColumn = v["cursorColumn"].toInt();
+        info.selectionCharCount = v["selectionCharCount"].toInt();
+        info.selectionLineCount = v["selectionLineCount"].toInt();
+        emit cursorActivity(info);
```
JuBan1: So, nothing would speak against saving a copy of `UiCursorInfo info` as a member variable and simply returning it via a `UiCursorInfo Editor::getCursorInfo()`, right? We're doing it wit `isClean` right now, and it sounds like a great idea to have the best of both worlds: non-blocking getters that return data synchronously.

JuBan1:
If you see an actual gain from it then that's fine. I'm just a bit weary of premature optimization.

Teklad:
If we were going to store the values like in your previous comment.... removing the timeout would be a very feasible thing to do.  (I think my slowness had to do with text mangling when getting selections previously, anyhow)

It does make sense to store the values that are actually useful.  I think the cursor info is only used for like one function though....  Unless I missed a call somewhere.  I do, however, think that the cost of storing would be less than the cost of querying each time, performance wise.

If @danieleds approves of caching the values I'm all for it...  The less CppToJS "get" calls we have to make, the better.

My main concern is high-traffic values, such as is-clean... however, a lot of these changes more or less remove the need for querying the editor for the data in most cases.

JuBan1:
Another one: `getLineCount()` returns 0 all the time. Using the `Go To Line` functionality doesn't work right now. Can't enter any line numbers because it chooses the result of `getLineCount()` as the upper limit.

>  I think the cursor info is only used for like one function though

We should keep new functionality in mind. Addons/Extensions/whatnot may want to do something with it.

Teklad:
Ah... that's one of the values I haven't updated with the signal/slot mechanisms yet...  I'm going to go ahead and remove the timer... storing the data in the editor as the data passes through seems like a good idea. 

Also, the line count is already sitting right at the bottom of our editor in the status bar... with a bit of magic we could yank that value out as an alternative method to getting the line count... but it sounds hacky, lol.

Teklad:
There you go... This seem a little better?

JuBan1:
After testing responsiveness with and without the timer I can say I see and feel the version using the timer be more responsive when drag-selecting code.

So, uhmm, I guess we'll keep it?

Teklad:
Dang it! Lol... okay... reverting that one.

danieleds:
Did you guys test the performance of the timer with a lot of tabs open?

JuBan1:
Still got it open Here's the output with 10 tabs opened:

Tab opened in 4msec
Tab opened in 3msec
Tab opened in 13msec
Tab opened in 12msec
Tab opened in 2msec
Tab opened in 3msec
Tab opened in 2msec
Tab opened in 2msec
Tab opened in 2msec
Tab opened in 3msec
Started in 303msec

However, with so many tabs its very clear that Nqq isnt yet usable when the startup timer ends (because the webengine pages hvent finished loading yet). It takes pretty much exactly another second for the last of the tabs to actually be loaded. I guess there's still some blocking call somewhere that prevents the pages from being loaded simultaneously.

Still, even with that extra second it's faster than the current branch.

JuBan1:
Edit: derp, for some reason my head went "timer? startup timer!" instead of recognizing the context of this discussion.

Regarding the cursor timer: I don't see any difference between having 2 tabs or 10 tabs open. Still more responsive than without the timer.

danieleds:
> Edit: derp, for some reason my head went "timer? startup timer!" instead of recognizing the context of this discussion.

Don't worry, I'm tired and I totally misunderstood what the timer was doing, so my question didn't make much sense either, lol.
Anyway, looks like a very good improvement.

src/editor/classes/EditorStatusEvents.js
```diff
@@ -48,29 +78,23 @@ class EditorEventHandler {
...
-        proxy.setValue("selections", out);
-        proxy.setValue("selectionsText", editor.getSelections("\n"));
-        var cur = editor.getCursor();
-        proxy.setValue("cursor", [cur.line, cur.ch]);
+        clearTimeout(this.cursorActivityTimer);
+        // We put this in a small timer so we don't flood the queue
+        // during selection and fast cursor movement.
+        this.cursorActivityTimer = setTimeout(function() {
+            proxy.sendEditorEvent("J_EVT_CURSOR_ACTIVITY", this.cursorActivityObject(editor));
+        }.bind(this), 20);
```
danieleds: Trick: if you use arrow functions, you don't need to bind(this).

E.g.
```
... = setTimeout(() => {
   proxy.sendEditorEvent...
}, 20);
```

Teklad: I'm javascript retarded... the fact that I've managed to do as much as I have(eg. CodeMirror plugin I wrote forever ago) is a miracle... I'll change this real quick.

Teklad:
I think I'm mostly done with this PR unless we need some more changes, lol.

src/ui/include/EditorNS/editor.h
```diff
@@ -285,6 +304,10 @@ namespace EditorNS
         bool m_bom = false;
         bool m_customIndentationMode = false;
         bool m_alreadyLoaded = false;
+        bool m_clean = true;
+        
+        UiChangeInfo docInfo;
+        UiCursorInfo cursorInfo;
```
JuBan1: Should use the `m_name` convention here.

src/ui/include/EditorNS/editor.h
```diff
@@ -296,38 +319,90 @@ namespace EditorNS
...
         void on_proxyMessageReceived(QString msg, QVariant data);
-        void on_cursorActivity() { emit cursorActivity(); }
-        void on_languageChange();
 
     signals:
+        //FIXME: Is the below signal still relevant?
+        /**
+         * @brief The editor received a message from Javascript.
+         */
         void messageReceived(QString msg, QVariant data);
```
JuBan1: As far as I can see this signal is indeed not used and not useful anymore.

Teklad:
All fixed, lol.

JuBan1:
Okay, I think I'm done nitpicking.

Teklad:
Lol, okay... back to implementing Xdg Icon Spec using standard C++ libraries. >.> (I get really bored when I'm not poking at notepadqq)

Teklad:
@danieleds ... Are there anymore changes I need to make prior to this getting merged?  Next up is asynchronous calls once these changes are approved.

danieleds:
@Teklad Sorry, busy days. Btw Travis is failing.

Teklad:
I think that's because a .travis.yml change slipped in when I merged @JuBan1's crash on exit fix, lol.

Teklad:
Nevermind that... its the stupid ui-tests failing again.

src/ui/EditorNS/editor.cpp
```diff
@@ -152,43 +151,57 @@ namespace EditorNS
...
-    {
-        QVariantMap data = m_jsProxy->getRawValue("language").toMap();
-        QString id = data.value("id").toString();
-        QString lang = data.value("lang").toMap().value("name").toString();
-        emit currentLanguageChanged(id, lang);
-    }
-
     void Editor::on_proxyMessageReceived(QString msg, QVariant data)
     {
         emit messageReceived(msg, data);
```
JuBan1: `messageReceived` has been removed. This causes travil to error out.

Teklad:
Travis is having a heart attack I think.  Or maybe my interwebs is being stupid... but I can't access it, lol.

Teklad:
Travis is broken again. :(

JuBan1:
Not our code that fails though. Travis had issues grabbing Qt from the repo.

danieleds:
Damn. That repo fails a lot. I've restarted the build.

Teklad:
Finally passing... Travis is really getting on my nerves, lol.

danieleds:
Sorry, I can't merge yet. The changeset is still a bit large and I'm still finding issues here and there.

I'm not quite convinced about how the code is organized in editor.cpp. Why `on_cursorInfoRequest()` has a slot-like naming? It appears to be treated like a method. Same with `on_documentInfoRequest()`.

Why is `m_docInfo` a variable of type `UiChangeInfo` which has fields `charCount` and `lineCount`? Something seems wrong with the naming. Reading the code I would expect `m_docInfo` to contain information like clean-ness, but that is stored in `m_clean`.

Reading the code it's not very clear what the methods `generateChangeActivitySignal()`, `generateCursorActivitySignal()`, ... are supposed to do either. It's clear that they send a signal but not how, when and why they are called. It would be better to still emit all the signals in `on_proxyMessageReceived` and creating utility methods just to populate the appropriate data structures to attach to the event.

As a side note, the whole javascript part (even that of the QtWebKit version of nqq) is a huge mess of spaghetti code, but that's mostly my fault. I'd like to give it more love before the final QtWebEngine merges into master, otherwise it's going to be a challange each time we're going to put our hands there, especially now that that code is growing.

Teklad:
The plan was to make it where we could eventually connect to the "on_request" stuff in mainwindow and call it passively... I don't think I ever finished it... naming things isn't exactly my strong suit either, lol.

I'm not really too comfortable about having on_proxyMessageReceived being a conglomerate of all the signals crammed into one place (More JS text signals = more mess in one place).

The structs are setup in such a way that only the NEEDED data is passed... rather than force-feeding everything.

isClean is handled separately from everything else... but it could be put into docInfo and returned via m_docInfo.clean.

danieleds:
> I'm not really too comfortable about having on_proxyMessageReceived being a conglomerate of all the signals crammed into one place (More JS text signals = more mess in one place).

What I'm saying is, instead of this:
```c++
    } else if(msg == "J_EVT_CONTENT_CHANGED") {
        generateChangeActivitySignal(data.toMap());
    } else if(msg == "J_EVT_CLEAN_CHANGED") {
```
is better to do this:
```c++
    } else if(msg == "J_EVT_CONTENT_CHANGED") {
        emit contentChanged(buildContentChangedEventData(data.toMap(), /*cache*/ = true));
    } else if(msg == "J_EVT_CLEAN_CHANGED") {
```

Teklad:
Okay, I did some general clean up and renaming of a bunch of different things.... I think it makes a bit more sense now... though I should really try to comment within my own code more, lol.

Agreed on the JS... it needs a pretty extensive overhaul if we're gonna do this right... app.js looks like a cluster**** (you know the word!)

I'd like to get the whole JSON language file we talked about implemented as well... and we could touch on loading CodeMirror modes on the fly rather than having them all defined in our index.html file.... which would get rid of all the CodeMirror.defineSimpleMode errors.

I think with these changes the code should make more sense.  Just to touch on the changes:
* DocInfo is now ContentInfo and m_docInfo is m_contentInfo
* m_clean was moved into ContentInfo as "clean"... since it pertains to the content of the document.
* New LanguageInfo struct just to keep things consistent with our signals.
* All generate[] functions have been changed to build[]EventData functions, signals are emitted through on_proxyMessageReceived and uses the utility build functions like you suggested.
* All previously mentioned functions feature a bool cache parameter that defaults to true, false causes it to bypass being placed in m_contentInfo and m_cursorInfo respectively.
* Moved on_cursorInfoRequest() to public and renamed to  requestCursorInfo(), same goes for on_documentInfoRequest, with the exception this was changed to requestContentInfo to keep things more consistent. (I'm still not sure why we use both document/content, as they're nearly the same thing.)

That just about covers it.  Unless there's some other semantic issues I think it should be good now.  Sorry about the mess,  I understand our code base is already pretty messy. It'll be a project in its own playing clean up later on.

danieleds:
Perfect!

I prefer "content" over "document" because, even if at the end it is just a matter of notation, they're conceptually different things. We can have e.g. two documents holding the same content, or we can have a document with no content, etc. A "document" is an abstract concept over the concept of "content". A "document" *has* content, but it is *not* the content. We handle documents at the level of `DocEngine`, while an editor only knows about content.

Merging

PR #394, opened 2017-03-13
QtWebEngine: Fix crash-on-exit.

JuBan1:
There was a SIGSEV that popped up in the experimental branch. I fixed it and broke editor signaling with it. When I fixed editor signaling I made the SIGSEV come back.

So here's the fix for the fix for the fix for the SIGSEV.

I should go into gardening, not programming.

------

Technical details:

I'm not quite sure why this happens but here's what I believe: When `TopEditorContainer` is destroyed, it first destructs its members and then calls its parent `QObject` destructor. That one makes sure all children get destroyed as well, that includes all `EditorTabWidgets` and the tabs therein. During that process, a number of signals are emitted that are being caught and acted upon by the now half-destructed `TopEditorContainer`. That kills the program.

Manually disconnecting the affected signals at destruction leads to a SIGPIPE instead (first time I ever managed that error!), probably because of some meta-type magic done by Qt. So instead, I added a switch to stop handling the problematic signal when the `TopEditorContainer` is being destructed.

I think it'll suffice for now, the whole tab stuff is due for an overhaul anyways.

### DIFF
--- a/src/ui/include/topeditorcontainer.h
+++ b/src/ui/include/topeditorcontainer.h
@@ -17,6 +17,8 @@ class TopEditorContainer : public QSplitter
     Q_OBJECT
 public:
     explicit TopEditorContainer(QWidget *parent = 0);
+    ~TopEditorContainer();
+
     EditorTabWidget *addTabWidget();
     EditorTabWidget *tabWidget(int index);
     EditorTabWidget *currentTabWidget();
@@ -49,6 +51,7 @@ public:
 private:
     EditorTabWidget *m_currentTabWidget;
     int m_currentTabIndex = -1;
+    bool m_shuttingDown = false;
 
 signals:
     /**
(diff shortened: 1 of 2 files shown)


Teklad:
YOU BROKE TRAVIS!  Lol... kidding, but this will do fine until we can get QWebEngine integration complete...  The next big project is Editor/Tab overhaul... I'm just @danieleds will be falling out of his chair with that PR. xD

JuBan1:
At least this doesn't seem to be my fault. Way to be an ass, Travis.

src/ui/include/topeditorcontainer.h
```diff
@@ -49,6 +51,7 @@ class TopEditorContainer : public QSplitter
 private:
     EditorTabWidget *m_currentTabWidget;
     int m_currentTabIndex = -1;
+    bool m_shuttingDown = false;
```
danieleds: Add a big comment summing up your "Technical details" paragraph?

JuBan1: Done

PR #398, opened 2017-03-15
Add parsing of %directory% in Run command

aryasenna:
The parsing of `%directory%` placeholders is advertised in the run menu, but the code wasn't actually there.
image

The current directory is retrieved using `QFileInfo` since `QUrl `does not provide clean directory path without "file://" scheme. Also to avoid the use of hard-coded regex.

P.S. `%path%` parsing is there but it's not advertised, I leave it to the maintainer to decide if it's should be removed.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -2027,10 +2027,13 @@ void MainWindow::runCommand()
     Editor *editor = currentEditor();
 
     QUrl url = currentEditor()->fileName();
-    QStringList selection = editor->selectedTexts();
+    QString path = url.path(QUrl::FullyEncoded);
+    QFileInfo fi(path);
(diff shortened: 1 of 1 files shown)


danieleds:
Thanks. While you're at it, can you change it so that this is what we get and what we advertise?

|Placeholder|Description|Example|
|----------|--------------------|-----------------------|
| `%url%` | Full URL of the currently active file. |`file:///home/my/path/file.txt` |
| `%path%` | Full path of the currently active file. | `/home/my/path/file.txt`|
| `%directory%` |Directory of the currently active file. |`/home/my/path/`|
| `%filename%` | Name of the currently active file. | `file.txt`|
| `%selection%` | *(unchanged)* | *(unchanged)*|

PR #401, opened 2017-03-18
Add show whole lines option to search results

MichalKopczynski:
I have recently switched from notepad++ to notepadqq. The only thing that disturbs me in notepadqq is that results of Find in Files option display just 50 characters before and 50 characters after match.

This makes searching through files which have long lines quite difficult. 
Especially in case of log files, lines may be mostly similar to each other, with just part of the line containing some specific information.  It is likely that this information might be cut out from search preview. As a result one have to click in each found line to see it in main window:
screenshot from 2017-03-17 23-45-40
This seems unnecessary as find results could show whole lines:
screenshot from 2017-03-17 23-44-34
However, as I understand some users may still prefer to have in preview only part of line which is close to match.
Therefore I propose to make enabling of this feature configurable in Preferences.
screenshot from 2017-03-17 23-43-03

Also, refactored buildResult function so that it uses linePosition to determine start and end of line. This should improve performance of that function.

The only issue left which could be fixed in next iteration is lack of horizontal scroll bar in search results. This could be helpful if lines are longer then Find result window width.

### DIFF
--- a/src/ui/Search/searchinfilesworker.cpp
+++ b/src/ui/Search/searchinfilesworker.cpp
@@ -1,6 +1,7 @@
 #include "include/Search/searchinfilesworker.h"
 #include "include/Search/searchstring.h"
 #include "include/docengine.h"
+#include "include/nqqsettings.h"
 #include <QDirIterator>
 #include <QMessageBox>
 
@@ -104,6 +105,8 @@ FileSearchResult::FileResult SearchInFilesWorker::searchPlainText(const QString 
     fileResult.fileName = fileName;
     Qt::CaseSensitivity caseSense = m_searchOptions.MatchCase ? Qt::CaseSensitive : Qt::CaseInsensitive; 
     QVector<int> linePosition = getLinePositions(content);
+    NqqSettings& s = NqqSettings::getInstance();
+    bool showWholeLines = s.Search.getShowWholeLines();
 
     const int totalLines = linePosition.length();
     const int matchLength = m_string.length();
@@ -117,7 +120,7 @@ FileSearchResult::FileResult SearchInFilesWorker::searchPlainText(const QString 
                 if (linePosition[i] > column) {
                     line = i-1;
                     if (hasResult) {
-                        fileResult.results.append(buildResult(line, column - linePosition[line], column, content, matchLength));
+                        fileResult.results.append(buildResult(line, linePosition, column, content, matchLength, showWholeLines));
                     }
                     break;
                 }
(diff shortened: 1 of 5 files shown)


danieleds:
Hi! Thanks for the contribution.

If I remember correctly, that 50 chars limit was just a matter of performance in case of files with very long lines.

I agree that it is an arbitrary limit which should be changed to avoid the issue you're describing. But I'm not sure about adding a dedicated setting for that... I'd prefer to find a better default, if possible.

Off the top of my head I'm also wondering what would users see if the matching pattern is in the middle of a very long line.

Alternative suggestions?

MichalKopczynski:
Hi! Thanks for response.

Users would see whole line with match highlighted. To see end of line which didn't fit to the screen (match could be there also) they would have to scroll horizontally to see it.
That is, after horizontal scroll bar is added - I will try to find time to add it next week - seems a little tricky at first to add it to QTreeView.
After it is added then behavior would be exactly like in notepad++ and at least for me it would be best option to leave it as default.

Regarding performance, we are displaying more text but instead we are not using time to shrink the lines. I would expect that performance improves, maybe at some cost of memory usage (but even that might not be the case). I will check if there is some performance degradation.

Some alternative suggestion that comes to my mind is to extend 50 chars limit. But in my opinion it's not possible to find universal value. There are source files with 100 chars and log files which have 300+ chars. And users are also using different screens.

Another one would be to get the width of the find result area and shrink only the lines which are longer then that. Lines could be cut one or both sides. But this seems overcomplicated...

In general, I am not sure if all users would agree on that but at least I would expect the lines to be displayed in the same way in search preview as in main window.

Teklad:
I'm the one who rewrote most of "Find in Files"... I recall there being some severe performance issue that caused me to do that(or maybe it was heavy memory allocation).  Between that and the threading there was a very significant performance boost for those functions.

QTreeView, for reasons I'm not aware of, has trouble de-allocating memory, even after the results are cleared.  If you can find and fix the cause of that... along with  horizontal scrolling, I'm definitely open to it... so long as its implemented decently.

Its been a while so I may have already fixed the memory leak issue... but that's been almost a year ago, lol.

JuBan1:
Since I'll have to make alterations to the search functionality in the near future, I think it makes sense to postpone this and do it all in one swing if that's alright with you.

I've added this as a feature to #406, I think it's roughly what you wanted:

image

MichalKopczynski:
That's fine with me.
Btw. find in open files option was also one thing that I missed. Glad it will be added :)

JuBan1:
I'll close this since it's finally included by de15c09's new search tool.

PR #402, opened 2017-03-19
Webengine async

### LINKED ISSUE #389: QtWebEngine: syntax highlighting doesn't work in Editor preview in preferences dialog

Teklad:
Putting this here so progress on the async implementation can be tracked.  The goal is to make calls asynchronous where we can... and supply a blocking fallback where we can't.

Currently, sessions uses the fallback getLanguage... mainly because I'm not familiar enough with that part of Notepadqq to make it asynchronous. (And it might be pretty messy to do so.)

I'm going to try and keep asynchronous to what's necessary for the moment just to be safe.... Most of the data is transported through JsProxy on the fly anyways, which alleviates a lot of the issues.

The current commits fix #389, I believe, which is a big plus since it's been a little evasive.

### DIFF
--- a/src/ui/Search/frmsearchreplace.cpp
+++ b/src/ui/Search/frmsearchreplace.cpp
@@ -528,22 +528,23 @@ void frmSearchReplace::on_radSearchWithSpecialChars_toggled(bool checked)
         manualSizeAdjust();
     }
 }
-
+#include <QDebug>
 void frmSearchReplace::on_searchStringEdited(const QString &/*text*/)
 {
     NqqSettings& s = NqqSettings::getInstance();
 
     if (s.Search.getSearchAsIType()) {
         if (ui->actionFind->isChecked()) {
-            Editor *editor = currentEditor();
+            currentEditor()->getSelections([&](const QList<Editor::Selection> selections) {
+                Editor *editor = currentEditor();
+                if (selections.length() > 0) {
+                    qDebug() << "Test" << selections[0].to.line;
+                    editor->setCursorPosition(
+                        std::min(selections[0].from, selections[0].to));
+                }
+                findFromUI(true);
 
-            QList<Editor::Selection> selections = editor->getSelections();
-            if (selections.length() > 0) {
-                editor->setCursorPosition(
-                            std::min(selections[0].from, selections[0].to));
-            }
-
-            findFromUI(true);
+            });
         }
     }
 }
--- a/src/ui/Search/frmsearchreplace.cpp
+++ b/src/ui/Search/frmsearchreplace.cpp
@@ -528,7 +528,7 @@ void frmSearchReplace::on_radSearchWithSpecialChars_toggled(bool checked)
         manualSizeAdjust();
     }
 }
-#include <QDebug>
+
 void frmSearchReplace::on_searchStringEdited(const QString &/*text*/)
 {
     NqqSettings& s = NqqSettings::getInstance();
--- a/src/ui/Search/frmsearchreplace.cpp
+++ b/src/ui/Search/frmsearchreplace.cpp
@@ -535,10 +535,9 @@ void frmSearchReplace::on_searchStringEdited(const QString &/*text*/)
 
     if (s.Search.getSearchAsIType()) {
         if (ui->actionFind->isChecked()) {
-            currentEditor()->getSelections([&](const QList<Editor::Selection> selections) {
-                Editor *editor = currentEditor();
+            Editor *editor = currentEditor();
+            editor->getSelections([&, editor](const QList<Editor::Selection> selections) {
                 if (selections.length() > 0) {
-                    qDebug() << "Test" << selections[0].to.line;
                     editor->setCursorPosition(
                         std::min(selections[0].from, selections[0].to));
                 }
--- a/src/ui/include/EditorNS/editor.h
+++ b/src/ui/include/EditorNS/editor.h
@@ -14,6 +14,7 @@
 #include <QMutex>
 #include <QMutexLocker>
 #include <QEventLoop>
+#include <QJsonDocument>
 namespace EditorNS
 {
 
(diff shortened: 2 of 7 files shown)


Teklad:
Welp, I'm done I guess, lol... Editor::[get]Value needs to remain synchronous until we're in the mood to rewrite docengine and sessions to be asynchronous (That's a PR in itself).

I can't really think of anything else that would benefit from asynchronous data for the moment.

src/ui/include/EditorNS/editor.h
```diff
@@ -280,8 +284,7 @@ namespace EditorNS
          * @brief Returns the currently selected texts.
          * @return
          */
-        QStringList getSelectedTexts();
-        QList<Selection> getSelections();
+        QList<Selection> convertToSelections(const QVariant& sels);
```
danieleds: Please be careful to keep documentation in sync when you move functions around.

danieleds: In addition, I think this method should be private?

src/ui/Search/frmsearchreplace.cpp
```diff
@@ -535,15 +535,16 @@ void frmSearchReplace::on_searchStringEdited(const QString &/*text*/)
...
-
-            QList<Editor::Selection> selections = editor->getSelections();
-            if (selections.length() > 0) {
-                editor->setCursorPosition(
-                            std::min(selections[0].from, selections[0].to));
-            }
-
-            findFromUI(true);
+            currentEditor()->getSelections([&](const QList<Editor::Selection> selections) {
+                Editor *editor = currentEditor();
```
danieleds: Can you move the assignment to `*editor` outside the callback and pass it in as a closure?

src/ui/Search/frmsearchreplace.cpp
```diff
@@ -535,15 +535,16 @@ void frmSearchReplace::on_searchStringEdited(const QString &/*text*/)
...
+                if (selections.length() > 0) {
+                    qDebug() << "Test" << selections[0].to.line;
```
danieleds: Sneaky debug line

src/ui/include/EditorNS/editor.h
```diff
@@ -307,6 +309,49 @@ namespace EditorNS
...
+         *                  Must accept QList<Editor::Selection> type as parameter.
+         */
+        template<typename T>
+        void getSelections(T callback)
+        {
+            sendMessageWithCallback("C_FUN_GET_SELECTIONS", 
+                    [&, callback](const QVariant& v) {
+                callback(convertToSelections(v));
+            }); 
+        }
```
danieleds: I'd like to find a clean way to move these out of the .h

src/ui/mainwindow.cpp
```diff
@@ -1520,14 +1523,13 @@ void MainWindow::on_editorMouseWheel(EditorTabWidget *tabWidget, int tab, QWheel
...
-
-    for (int i = 0; i < sel.length(); i++) {
-        sel.replace(i, func(sel.at(i)));
-    }
+    currentEditor()->getSelectedTexts([&](QStringList sel) {
+        for (int i = 0; i < sel.length(); i++) {
+            sel.replace(i, func(sel.at(i)));
+        }
+        currentEditor()->setSelectionsText(sel, Editor::selectMode_selected);          
+    });
```
danieleds: Can you assign currentEditor() to a variable outside the callback?

src/ui/mainwindow.cpp
```diff
@@ -1997,23 +1999,33 @@ void MainWindow::runCommand()
...
+
+    // Check for selection before performing asynchronous call.
+    if(cmd.contains("\%selection\%")) {
+        editor->getSelectedTexts([cmd, doRun](const QStringList& selection) mutable {
+            if(!selection.isEmpty() && !selection.first().isEmpty()) {
+                cmd.replace("\%selection\%",selection.first());
+            }
+            doRun(cmd);
+        });
+    }else {
```
danieleds: Fix code style (yeah I know, it was already a mixture of everything)

src/ui/mainwindow.cpp
```diff
@@ -2030,48 +2042,6 @@ void MainWindow::on_actionPrint_Now_triggered()
...
-{
-    QUrl fileName = currentEditor()->fileName();
-    if (!fileName.isEmpty()) {
-        QStringList args;
-        args << fileName.toString(QUrl::None);
-        QProcess::startDetached("google-chrome", args);
-    }
-}
-*/
-QStringList MainWindow::currentWordOrSelections()
```
danieleds: Wait, that was actually useful. We lost it along the way.
Can you just fix and keep `currentWordOrSelections()` even if it's not used for now? I'd like to provide it as a Run placeholder later.

Teklad: I could reimplement it and keep it for now, yea... shouldn't be an issue.... I haven't the slightest clue if it works, however.

Teklad:
A "clean" way to move template functions outside of the .h.  Templates are notorious for not liking to work correctly when you split the definition from the implementation.... mainly compiler error issues.

Teklad:
Gonna start getting these moved out of the header then I'll rework currentWordOrSelections to be an editor function.

Teklad:
It really doesn't want to be separated, lol.... Compiler errors all around complaining to high heaven about it not being declared.

Teklad:
Does this work for you @danieleds?  sendMessageWithCallback is a template function... but the rest accept a strict type of std::function... so it will throw an error if you try to use it incorrectly.

This has the following benefits:
* Strict typing when accessing specific asynchronous calls(i.e. can't request QString from getSelections).
* Less overhead than using std::function twice.

The only limitation I can think of is if you capture more than two references and/or pointers you'll cause a heap allocation(the std::function overhead I was talking about)... which shouldn't really be an issue under any normal usage scenarios.

src/ui/include/EditorNS/editor.h
JuBan1: I think that's near impossible for such a generic `T`. But we could move the function definition outside the class so we only have function declarations there.

danieleds:
Sorry, been busy with work.
Looks very good to me!
Merging :)

Issue #378, opened 2017-03-07
QtWebEngine: debug warnings


danieleds:
I get the following warnings a lot at runtime:

    js: Uncaught TypeError: CodeMirror.defineSimpleMode is not a function

and

    Registered new object after initialization, existing clients won't be notified!

Teklad:
The 2nd message is internal to QT... it really can't be helped for the moment.... the former one is actually an issue with CodeMirror itself.... we have a few languages that are trying to access methods that no longer exist within CodeMirror.

Issue #265, opened 2016-09-10
Editor focus is not correct on start


danieleds:
This is evident after @JuBan1 's PR #249 which added session support (but the issue isn't caused by that).

I'm trying to understand what's going on.
##

Teklad:
Putting currentEditor()->setFocus() at the end of loadSession forced it to focus properly for me.  Of course there may be a better place to put it.

The way it was setup appears to favor focusing the very last tab instead, which is impossible when its not displayed on initial startup. :P

danieleds:
Yes, that's a different issue, I noticed that too. Your fix is correct.

However the problem discussed here is different: see here https://github.com/notepadqq/notepadqq/pull/249#issuecomment-245045728

The problem, I believe, is located among TopEditorContainer, EditorTabWidget and maybe the JS script (unlikely). We shouldn't need to apply the focus manually to fix things. It should just work.

severach:
I'd like to expand this with tabs.

Open 4 tabs in noteqpadqq.
Select the first tab.
Close notepadqq.
Open notepadqq.
Press Up or Down. These keys do nothing because the editor does not have the focus.
What has the focus? Press left/right. The tab bar has the focus.
Press tab a few times and the focus lands in the editor.

There's more.
Select the first tab.
Close and open notepadqq.
Click on each tab with the mouse. Focus is outside the editor in tabs 1,2,3. The editor has the focus in the last tab 4.
Click through each tab setting the focus to the editor.
Click through each tab again and the focus will still be in the editor.

Issue #388, opened 2017-03-11
QtWebEngine: Possible memory leak


JuBan1:
Right now the QtWebEngine branch leaks a decent amount of memory.

To replicate:
* Open your favorite system monitor and watch notepadqq-bin's memory usage.
* Open and close tabs a bunch of times.
* Watch memory grow and not shrink again.

After 60 open/closed tabs memory consumption grew from ~120MiB to ~320MiB for me. That's around 3MiB per opened tab.

Not sure where it originates from. The `CustomQWebView`, `JsProxy` and `Editor` objects are properly destroyed when a tab is closed. Valgrind crashes the process for some reason, so I can't really trace it.

This may not be a memory leak but some lazy block deallocation from somewhere. Not sure.

Teklad:
There's similar behavior when you extensively use the file search results
widget... Once you clear the results the memory never gets fully
reclaimed... I'm almost certain this is normal QT behavior (it does bother
me though).

My most recent patch to said widget made it capable of reclaiming "most" of
the memory... But I think QT has some internal caching that it does. At
least I hope so.

On Mar 11, 2017 1:36 PM, "JuBan1" <notifications@github.com> wrote:

Right now the QtWebEngine branch leaks a decent amount of memory.

To replicate:

   - Open your favorite system monitor and watch notepadqq-bin's memory
   usage.
   - Open and close tabs a bunch of times.
   - Watch memory grow and not shrink again.

After 60 open/closed tabs memory consumption grew from ~120MiB to ~320MiB
for me. That's around 3MiB per opened tab.

Not sure where it originates from. The CustomQWebView, JsProxy and Editor
objects are properly destroyed when a tab is closed. Valgrind crashes the
process for some reason, so I can't really trace it.

This may not be a memory leak but some lazy block deallocation from
somewhere. Not sure.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/notepadqq/notepadqq/issues/388>, or mute the thread

Teklad:
It could also be a bug regarding how we handle raw editor pointer... But
you would have to test master Notepadqq for the same leak if that were the
case.

JuBan1:
Master doesn't exhibit the same behavior. The destructor of the closed tab's Editor is properly called, too.

Teklad:
In any case, this should resolve itself once we rework the editor assuming
its in there somewhere.  We won't be having to recreate so many widgets
like we're doing now.

[...]

PR #403, opened 2017-03-20
Webengine JavaScript refactor

Teklad:
So.... this PR looks a lot fatter than it really is... it partially implements the json language file bit we were talking about.

At some point or another... I accidentally fixed that notorious preview gutter bug while I was working on this as well.

There's still some work to be done... but the overall performance of this is very good!  This also happens to get rid of the QJSEngine requirement from the initial QWebEngine PR.

I'll most likely start cleaning up the javascript in this PR as well.  Its a mess right now and could really use some TLC.  Feel free to review the code when you guys get a chance.  It's still a work in progress though, lol.


src/ui/include/mainwindow.h
```diff
@@ -256,16 +254,13 @@ private slots:
     int                 saveAs(EditorTabWidget *tabWidget, int tab, bool copy);
     QUrl                getSaveDialogDefaultFileName(EditorTabWidget *tabWidget, int tab);
     void                setupLanguagesMenu();
-    void                transformSelectedText(std::function<QString (const QString &)> func);
+    void                transformSelectedText(QString (*func)(const QString&));
```
JuBan1: Why that change? Isn't a `std::function` all around more convenient and more compatible?

JuBan1:
Is this PR supposed to also have all the changes from #402 in it?

JuBan1:
Here's a question for my understanding (because I only take a glance at some of the changes at it's hard to keep up):

Right now we've two ways of getting data: the async calls you're implementing right now and using signals to set `Editor`'s members like `m_contentInfo`. Why can't we only use signaling to handle all getter functions? What's special about `getSelections` that it can't be done via signals and members?

src/ui/include/mainwindow.h
Teklad: It didn't make sense to have the overhead for something that doesn't need any captures to begin with.

Teklad:
That works great for data that's accessed rather passively... but not so much for on-demand requests.  As an example, if we got 10 functions that use getSelectionsText... that's either 10 signals + lambdas. or one signal... a bunch of functors and a message transport handler.

... Though the idea of just sending a "request" and intercepting the reply does sound nice, lol.... it's basically what we're doing with the async setup.... just.. uglier?

Teklad:
The biggest blocker for that change is "getValue"... for the record... the way its tied into sessions, mainwindow, docengine, etc... makes it a pain in the behind to get working right.

Teklad:
I could change all the get[GetFunctionName] thingies to request... and provide a small flexible API for handling callbacks there in one place.  It's probably doable now that I think about it.  So I'm probably gonna try... because segmented API annoys the crap out of me.

Teklad:
I can't really think of a way to implement that for the others without blocking the current running thread, lol... Asynchronous programming is really tedious.

Teklad:
Here's an example of what the code would look like with signals/slots... assuming you did an in-place lambda.

```cpp
            auto conn = std::make_shared<QMetaObject::Connection>();
            *conn = connect(editor, &Editor::selectionReply, [this, conn, editor] (QList<Editor::Selection> selections) {
                if (selections.length() > 0) {
                    editor->setCursorPosition(
                        std::min(selections[0].from, selections[0].to));
                }
                findFromUI(true);
                QObject::disconnect(*conn);
            });
```

JuBan1:
> That works great for data that's accessed rather passively... but not so much for on-demand requests. As an example, if we got 10 functions that use getSelectionsText... that's either 10 signals + lambdas. or one signal... a bunch of functors and a message transport handler.

I think you misunderstood. What I mean is that, right now, `Editor` already catches a whole bunch of signals from JavaScript. Like `J_EVT_CLEAN_CHANGED` causes `m_contentInfo.clean` to be set. Then we can super easily access it via `Editor::isClean()`. 

Why is it we don't do that for stuff like `getSelections()`? Just have `Editor` catch a signal if selection changes (we kind of already do that with `J_EVT_CURSOR_ACTIVITY`) and store it. Then we'd need to lambdas or any other magic inside the getter function.

Sorry if that's just me being daft :P

> Here's an example of what the code would look like with signals/slots...

Kill it with fire!

Teklad:
Lol... oh I get you.  I think @danieleds said something about not 
storing values that aren't frequently accessed or something along those 
lines... though I don't think it would be an issue with "selections" as 
it's just an integer.  I think he was mainly refering to getValue, 
getSelectionsText, etc....

Teklad:
@JuBan1, how's this look for getSelections?  :)

JuBan1:
I think it's good- if it works :p The less we have to worry about async operations in all of the code using `Editor` the better.

Teklad:
Seems to work fine... getSelections is only used one place in the code 
so it wasn't hard to test. :P... Speaking of which I just figured out 
how to iterate over a QVariant without performing conversions... this 
has been a real annoyance for me for a while, lol.

Teklad:
So the proxy is probably going to go back to just sending messages/signals... as it stands it doesn't really need to store any values, because that's being handled by Editor as the data passes through.

So long as I tread carefully I should be able to implement a very clean interface for all of this and reduce the need to rely on javascript callbacks in general.

Teklad:
EVIL!  Unwinding that javascript spaghetti took forever!  There might be a couple of functions in the wrong class... but aside from that, I think its a bit cleaner.... or at least more reliable.

This changeset really got out of hand, lol...

Teklad:
@JuBan1, I'm going to go out on a limb and assume @danieleds is having a few of his busy weekdays.  Care to go over the Javascript portion and see if there's any code I could move around and/or more correct class name/filename for certain things?

src/editor/classes/Helpers.js
```diff
@@ -0,0 +1,162 @@
...
+            to keep track of the extra string length outselves. 
+            tabToSpaceCounter holds the number of extra spaces we've added.
+        */
+        var trueOffset = offset + this.tabToSpaceCounter;
+
+        var numSpaces = tabSize - (trueOffset % tabSize);
+
+        // Since the original tab is replaced by a space we only need numSpaces-1 
+        // new spaces.
+        tabToSpaceCounter += numSpaces-1;
```
JuBan1: `this.` missing?

src/editor/classes/MessageHandler.js
```diff
@@ -0,0 +1,457 @@
...
+
+    C_CMD_TRIM_LEADING_SPACE(data) 
+    {
+        editLines(function (x) { return x.replace(/^\s+/, ""); });
+    }
+
+    C_CMD_TAB_TO_SPACE(data) 
+    {
+        editLines(function (x) {
+            tabToSpaceCounter = 0
```
JuBan1: Does this still work since `tabToSpaceCounter` is now in `App.helpers`?

src/editor/classes/MessageHandler.js
```diff
@@ -0,0 +1,457 @@
...
+
+    C_CMD_SPACE_TO_TAB_ALL(data) 
+    {
+        editLines(function (x) {
+            App.helpers.spaceToTabCounter = 0
+            var tabSz = editor.getOption("tabSize")
+
+            return x.replace(/ +/g, function(match, offset) {
+                return App.helpers.spaceToTab(match, offset, tabSz)
+            });
```
JuBan1: While I like the idea of moving code logic into another class, I think this is a bad place to do it. There are really tricky interactions here where the algorithm won't work if a user forgets to call `App.helpers.spaceToTabCounter = 0` or similar.

Previously all code was right next to each other so the interactions were easy to see and follow but now that code is more encapsulated it becomes very hard to follow code execution.

Can we move all the code to `App.helpers` so we only cal something along the lines of `editLines(App.helpers.spaceToTab)` or similar? 

Same goes for the other XtoY functions.

Teklad: How about we forget about "Helpers" all together and make it a plain functions file since they're all utility functions... this would fix all three problems I think.

src/editor/classes/CodeMirrorInit.js
```diff
@@ -0,0 +1,59 @@
...
+                    return;
+                }
+            }
+
+            if (cm.options.indentWithTabs)
+                cm.execCommand("insertTab");
+            else
+                cm.execCommand("insertSoftTab");
+        },
+        "Shift-Tab": function (cm) {
```
JuBan1: So, button presses are pretty much events too. Might be a nice idea to move these two functions into EditorEventHook.js as well?

JuBan1:
Take my input with a grain of salt. My JavaScript is not great :)

Teklad:
I'm trying to think of the best way to go about those helper functions in particular...  Having them in a class does break code logic a bit.  I could write a shortcut "functions" file that calls the appropriate App.helpers.Function... but in all reality I think that's just nesting, lol.

src/editor/classes/MessageHandler.js
JuBan1: My main point was that tightly coupled code should either be near each other or should be very clear to reduce code spaghetti. So if you think that solves it, give it a try.

Teklad:
I seriously need to learn how to base a branch on a branch using git... merge conflicts are evil, lol.

Teklad:
a.k.a. I didn't mean to include #402's changeset into this.

Teklad:
Gotta remember to never do that again, lol.  I think we're good, lol... cherry-pick is so useful. :)

Teklad:
I think that should do it... we'll just keep the helpers as normal functions.  They don't really have a place in a class when it comes right down to it... not unless we inlined them into the MessageHandler prototype.

Teklad:
This is going pretty smoothly now... Would anyone object to having the JS-To-Cpp transport(CommunicationsManager) having functions like "pushContentChangedEvent()"... or should that go in a different place?  Maybe directly in App?

Then it would be like "App.pushContentChangedEvent(data)"

JuBan1:
Sounds like `CommunicationsManager` would be the right place for these functions.

Teklad:
Yea... just wanting a uniform interface for these... I don't like seeing "J_EVT_BLAH_BLAH" sprinkled around everywhere, lol.

src/ui/EditorNS/editor.cpp
```diff
@@ -383,9 +383,14 @@ namespace EditorNS
                 }
             }
         }
-        // If we make it here we need to make sure we're loaded.
-        waitAsyncLoad();
+        // We wait for the page to finish loading before doing this one.
         // First non-blank line
+        connect(this, &Editor::editorReady, this,
+                &Editor::detectLanguageFromContent);
```
JuBan1: What is `setLanguageFromFilename` is called at some point after the `editorReady` signal was sent? Wouldn't that mean `detectLanguageFromContent` doesn't get executed?

Also, it's a bit weird for a function called `setLanguageFromFileName` to deduce the language from the doc's *contents*. IMO it should either be called differently, split into two functions, or simply removed. There are only like five languages with a non-empty `firstNonBlankLine`. Is it actually worth keeping it?

src/ui/EditorNS/editor.cpp
```diff
@@ -267,26 +267,48 @@ namespace EditorNS
...
+   
+    void Editor::initLanguageCache()
+    {
+        QFileInfo fileInfo(Notepadqq::editorPath());
+        QString fileName = fileInfo.absolutePath() + "/classes/Languages.json";
+        QFile scriptFile(fileName);
+        scriptFile.open(QIODevice::ReadOnly | QIODevice::Text);
+        QJsonDocument json = QJsonDocument::fromJson(scriptFile.readAll());
+        scriptFile.close();
+        for (auto it = json.object().constBegin(); it != json.object().constEnd(); ++it) {
```
JuBan1: Make it a `const auto&` to prevent copies =)

src/ui/EditorNS/editor.cpp
Teklad: I'm still debating on how to do this cleanly... I might put a C_FUN_DETECT_LANGUAGE in the handlers to deduce the language instead.

Teklad:
setLanguageFromFileName is getting called before setValue, so this breaks the entire detection process.

JuBan1:
I vote for outright removing the "detect based on contents" thing as it's used only by very few languages.

Teklad:
Yea... I mean technically with the changes we could allow adding extra extensions/names for filetypes(i.e. QT's ui files).

I'm keen on removing it myself... but we'll probably have to wait until we have an alternative method implemented.  I'm just assuming that @danieleds would say something along those lines, lmao.

Teklad:
My biggest issue with all of our "get" methods is if they are called to early it will throw a javascript error because something isn't defined quite yet... Javascript doesn't do waiting around for a return value too well... unless we can find a way to force runJavaScript to return from a promise.

The entire setup is based around the fact that javascript needs to be ready... but waiting on that ready signal has unfortunate side effects...  The only real way to fix it would be to rewrite chunks of Notepadqq to operate on a callback mechanism, which is just a ton of work.... it'll have to be done eventually, but that's honestly another project all by itself.

JuBan1:
So where does this translate into actual issues at the moment?

Teklad:
As far as I can tell... the ACTUAL issues with the QWebEngine port are darn near next to none now... since this pull request fixes the preferences preview and reworks the Javascript to be a little more sane.

This PR fixes #378, #389,  and #390. as well as the gutter overlap bug.  The only real thing we NEED to look into would be #388 and #405.  Other things such as utilizing localStorage can be done another time.

#388 probably isn't going to be fixable for the moment, which isn't too big of a deal considering we're planning to overhaul how editor/tabs work anyhow.

#405 is probably pretty simple to fix, though I haven't the slightest clue how to do it, lol.

Teklad:
Also, only three functions are asynchronous as of now.... If I can find a way to make JsProxy behave in its own thread we can block in the thread and wait for the result.

I'd basically be implementing a JsProxy::promise type feature, probably using something similar to this:
http://en.cppreference.com/w/cpp/thread/promise

src/editor/init.js
```diff
@@ -7,15 +7,8 @@
...
-    var link = document.createElement("link");
-    link.href = path;
-    link.type = "text/css";
-    link.rel = "stylesheet";
-    link.media = "screen,print";
-
-    document.getElementsByTagName("head")[0].appendChild(link);
-    return link;
+function addStylesheet(sheet){
+	document.getElementById('pagestyle').setAttribute('href', sheet);
```
danieleds: Why the change? This completely alters the behavior of the function, but the function is still used in the same way as before. Is this intended?

Also, if the new behavior is actually intended, this should probably not be called "addStylesheet" anymore ;)

Teklad: Yea... swapStylesheet is more appropriate... just haven't got around to name-changing, lol.

JuBan1: `setThemeStyleSheet` or something like that.

JuBan1: @danieleds the behavior is altered but it still has the same effect- just better. It used to create a new link element  every time the theme was changed. By appending it to the document it would overwrite the previously created link element.

What @Teklad did was to create one named link element and just set its source to whichever theme.css was selected.

Teklad:
So I've been thinking... since MessageHandler.js contains the message handlers... should we have a separate Editor class as a tie-in?  C_FUN_SET_VALUE would call App.editor.setValue()... etc... Not the best example, but I think you get my point.

Teklad:
Someone care to test the latest commit to see if it actually fixes #265 for everyone?  You should be able to go to any tab and start typing now.

Teklad:
@danieleds, I'm gonna need you to merge my CodeMirror PR so I can add a cleaner way to update the background.... "optionChange" wasn't added until after 5.22.0 :)

Teklad:
I'm also going to have to modify the configure script because codemirror now uses npm.

JuBan1:
Alright, testing it. Here's what I got:

* debug warnings mostly gone, gutter overlap gone, annoying white flash gone. 

* There's still one warning left: `js: Err[CommunicationsManager.js: 15 ]: Uncaught TypeError: Cannot read property 'sendEditorEvent' of undefined`

* #265 is not fixed. In fact, it got worse. Key combinations don't work (like CTRL+Z) and the Ui info (file name in titlebar, and file size in status bar) don't show the right file. Steps to reproduce: open two files, move one to second view. Now try to click into the two editors alternatingly. You'll see the Ui doesn't update. In addition, key combos only work for the left editor. Seems Nqq doesn't recognize the switch of currentEditor through clicks anymore.

JuBan1:
There's also what seems to be a slightly longer delay when opening Nqq. Here's a gif:



Once the Ui opens, I get a black content pane for about a second or so. That didn't happen before, at least not quite as noticeably. It's not a big deal (since doesn't happen when opening new tabs) but I thought it was noteworthy.

Teklad:
I made a few rendering changes to fix the flashing which causes a black background as a side effect.  It shouldn't be too big a deal though. 

Haven't the slightest clue why hotkeys are broken... The container code was pretty fragile. I'm sure something simple we did broke it to pieces

JuBan1:
Found the cause of tab switching not working. in `editor.cpp` inside `on_proxyMessageReceived`, the `if if(msg == "J_EVT_GOT_FOCUS") emit gotFocus()` is never ever triggered.

So JS seems to not emit the right event. I'll leave it  to you to fix that since you know your way around in the new JavaScript files :p

danieleds:
Wait. Why is npm required to build notepadqq??

danieleds:
There are too many unrelated changes in this PR. Can we bring it back to the essential refactoring, leaving out the various fixes for now?

Especially upgrading CodeMirror is delicate. I'd like to do that in a completely separate PR.

Teklad:
One of the events I need kinda requires CodeMirror be up to date... And 
CodeMirror added npm as a build dependency.  I can revert the changes 
but I don't feel a timer is the most reliable method for updating the 
background of the widget to match the theme.

Teklad:
I'll just revert all the configure related changes and use the timer method for the moment...  I'll stash it away in a patch for the next PR (CodeMirror is going to have to be updated in order to get the optionChange event I need.)

I didn't run into any issues regarding the update though, for the record.  There's a PR on notepadqq/CodeMirror with the updated and merged CodeMirror branch when you find time.

Teklad:
Okay... reverted all the CodeMirror modification stuff... just keep in mind I'm going to need it updated in order to provide a more elegant solution for the flashing bug, lol.

Teklad:
@JuBan1, this should fix the focus issue... try it again and see if that also fixes the other issues. :)

JuBan1:
Still not 100% fixed. Open two documents, move one to other view, then close and reopen editor. The left view's document is active and you can type in it. But if you click on the right view's document, you can type but the Ui isn't updated (and key combs still apply to the left doc). You can have to either click on the tab bar or first click on the left and then on the right document again.

JuBan1:
Current hypothesis of what causes it:

When loading a document, DocEngine also calls `editor->setFocus()` is used which calls `sendMessage("C_CMD_SET_FOCUS")`. Since JS isn't ready at this point, the message is moved into the queue and executed later. Meaning every document created at start-up has a `C_CMD_SET_FOCUS` in the pipe that is executed once JS is ready. Maybe that screws with things.

Honestly, I'm not sure. There's some weird stuff going on.

Teklad:
I think the focus event isn't firing on the JS side when swapping between views... it might be better to capture the widget's focus instead.... but that is gonna be pretty hairy with how QWebEngine is set up.

Teklad:
Interesting... the focus event doesn't fire until both views have been clicked twice...

JuBan1:
Does that mean the JS only sends focus events once both views were clicked twice?

Teklad:
Correct... because we automatically focus them on start... so technically they are ALL focused at start... which is the issue... it breaks event handling in JS.

JuBan1:
So can we *not* focus any of them at the start, then specifically send a focus event to the one document that we want focused?

Teklad:
That would be the best course of action... the problem is I think currentChanged fires off on tabs as soon as we start... so we'd need to find a way to hook that up AFTER loading is fnished.

JuBan1:
For now I suggest a simple patch instead of anything big: add something like a `bool m_blockFocusEvents;` to .. uhh.. TopEditorContainer? and simply return in `currentChanged` when that value is `true`. Add some documentation and we can fix the whole mess when the tab widget stuff gets modernized.

Teklad:
I have to agree... just one more thing to add to the list of reasons why 
we should overhaul all that stuff after WebEngine is finished. :P

Teklad:
Maybe re-iterating through the open editors and unfocusing all but currentEditor would work for now... just need it to not be stupid, lol.

JuBan1:
Right now in `Sessions` the load function calls `tabW->clearFocus();` for each loaded editor, I think that was part of fixing the focus issues with QWebKit. But I don't think `clearFocus();` sends anything to the JS since it's the normal `QWidget::clearFocus` implementation. So a possibility is to overwrite it in EditorTabWidget to also send a JS command to unfocus.

Teklad:
@JuBan1, I think I came up with a somewhat elegant solution that isn't too hacky... try this out and see if it works for you now.

src/ui/editortabwidget.cpp
```diff
@@ -197,6 +216,7 @@ void EditorTabWidget::tabRemoved(int)
 
 Editor *EditorTabWidget::currentEditor()
 {
+    setFocusProxy(editor(currentIndex()));
```
JuBan1: Is there a reason this is in `currentEditor()` rather than `setFocus()`?

JuBan1:
Nice! Yes, that seems to take care of everything.

src/ui/editortabwidget.cpp
Teklad: Actually that's left over... I should be able to remove that entirely, lol.

JuBan1: I just removed it and it broke things again. Guess it's needed at least to some degree.

Teklad: Give me a few... making some small changes which should make it better.

JuBan1: Removing that line and adding a `setFocusProxy(editor);` to `EditorTabWidget::setFocus()` works, however.

Teklad:
Give that a try.. and tell me what's broken if it doesn't work. :)

Teklad:
Also, I'm pretty sure this is the last bug that needs fixed before merging... I'm more or less done with the refactoring (i.e. I can make minor adjustments later)

JuBan1:
Heh! It's still there but it doesn't happen all the time. Steps:

* Open two tabs, move one to other side.
* Click inside the left view.
* Close and reopen Nqq a few times. In roughly 1 in 5 cases the left view will have the blinking caret while the right view has focus (meaning key combs go right but typing goes left).

There's also another bug that snuck in:

* Open two new tabs (don't open files! just new tabs) write something into them so they get remembered.
* Close and re-open Nqq. 

You'll find the first of the two tabs is not opened again. Also works if you open a non-file document and a file document. The non-file document won't be loaded if it sits left of the file document. If it's to the right of the file document it'll be loaded.

Teklad:
Lol wut?  I don't understand how that even snuck in... I haven't even touched sessions until just now... (Removing the line I added does not fix it)

JuBan1:
Another bug I can reproduce very consistently:

* Open a file document, then another non-file document (and write something into it).
* Make sure the non-file document has focus and is to the right of the file document.
* Close and open Nqq
* After opening, the non-file document is open (as it should) but the editor doesn't show. It just stays invisible. I have to click on another tab (like the file document) and back to make it show up.

This does *not* happen when it's the file document that has focus. It also doesn't happen when there is *only* a non-file document open. I can't test it when the non-file document is left of the file document because the non-file document isn't loaded properly in this case (as per the other bug).

Crazy stuff.

JuBan1:
Got it. It's DocEngine's load code:

```
// If there was only a new empty tab opened, remove it
                if (tabWidget->count() == 2) {
                    Editor *victim = tabWidget->editor(0);
                    if (victim->fileName().isEmpty() && victim->isClean()) {
                        tabWidget->removeTab(0);
                        tabIndex--;
                    }
                }
```

This crap stops it. First it properly loads the non-file document, then loads the file document. So `tabWidget->count()` is 2. And `victim->fileName()` is obviously empty because it's a non-file document. And `victim->isClean()` is also true because of the changes to the JS code.

Teklad:
So basically the JS changes borked docengine's assumptions?

JuBan1:
Yes, because the way JS works right now is that a freshly loaded document is clean at first (because `bool ContentInfo::clean = true;`), then it's properly set to clean/dirty a short while after when the JS is loaded.

Problem is we can't just set `ContentInfo::clean = false;` because then it would break the *other* way around (meaning if the victim tab was actually empty it wouldn't remove it).

Teklad:
Lol... this PR is making Notepadqq's older code shine really brightly(and not in a good way)

JuBan1:
Maybe add a `bool Editor::m_isBlank` member variable that is initially set to true when the editor was created via `addNewDocument()` (which is used when creating an empty document) and set to false if `loadDocuments()` was used.

Teklad:
Try that one... should fix it I think.

JuBan1:
Now it doesn't work the other way around because at the time where DocEngine calls `isLoadedDocument` it will always be `false` (because it's only set to `true` when `J_EVT_DOCUMENT_LOADED` is emitted).

Steps to reproduce:

* Close all tabs so you only have an empty document left. Then open any file.
What should happen is that the empty document is removed/replaced by the file. What happens is that the empty document stays open while a new document is also created.

JuBan1:
Hmm, can you run by me why `m_webView->page()->setBackgroundColor(Qt::transparent);` causes the tab to not show up? Apparently `updateBackground()` doesn't do its job if the document is a non-file doc. But why?

Teklad:
I've not figured it out yet... but for some reason if that gets set at just the right time it breaks QWebEngine rendering until you change tabs (i.e. it's probably a QWebEngine rendering bug).

As for the dirty/clean issue... could you modify Sessions to load each file individually instead so you can set the clean status before that check comes to pass?  Or we could write a new function for docengine that lets you apply some functions during load.

Teklad:
Also, I'm not sure what you mean by non-file doc?

Teklad:
ALSO!  The reason the focus wasn't correct 1/5 times was because APPARENTLY QT thought it would be a GREAT idea to make m_webView->load() cause focus to be given as a feature. -_-... The fix was simply adding setEnabled before and after load()

JuBan1:
We don't really have good terminology for it but by "file document" I mean a document that was opened by loading a file, and a "non-file" document is one that was initially an empty document.

Basically, if you CTRL+S a document and it asks you where to save it's a non-file document.

Oh, or in terms of code, if `editor->fileName` is empty it's a non-file document.

JuBan1:
> could you modify Sessions to load each file individually instead so you can set the clean status before that check comes to pass?

It does that already, doesn't it? 

The way I see it, the problem is that can't immediately call `isClean()` on a newly created `Editor` because it's not set yet. And that's because the new JS stuff works asynchronously. So if I wanted to modify the session load code to work with that I would have to listen for the editor to emit its `JS_DOCUMENT_READY` signal or whatever it was called. That'd turn the load code into an utter mess.

So I think either we need a blocking way to retrieve the value of `isClean()` (aka one that waits until the JS is ready) or we use a workaround like the one I posted a bit further above (with a `m_isBlank`). First one is probably a bad idea because it would revert much of the faster startup speed of Nqq. 

I can implement the second alternative if you want - just not today. It's still early enough to nag you with bug reports but too late to do any coding :P

JuBan1:
> I've not figured it out yet... but for some reason if that gets set at just the right time it breaks QWebEngine rendering until you change tabs (i.e. it's probably a QWebEngine rendering bug).

That it only happens with non-file documents makes me think it's something on our end. For some reason it's okay if a file document has focus. Maybe there's another hidden bug like the one with `isClean()` not being set right off the bat.

Teklad:
Maybe, lol... putting an asynchronous thing into a mostly synchronous program is a brain tumor waiting to happen, after all.... what really needs to happen is docengine needs to just load files and let something else handle the pruning... something that doesn't have to be aggressive/fast.

JuBan1:
Pretty much. In fact I'd say DocEngine has no business closing any tabs at all. Especially not in the code that is supposed to *load* tabs.

I think a central piece of code is needed that runs early in the constructor of `MainWindow` and basically does this:
```
if(there is a session to load)
    load_the_session();
if(nqq was started with commandline parameters telling it to open a specific file)
    open_that_file();
if(there is no open document yet)
   open_an_empty_document();
```

But that's probably for another PR or else @danieleds will have a stroke.

Teklad:
DocEngine has no business interacting with the editor period... it should just emit a signal "fileLoaded" with the data and let the caller that requested it handle it. Your example sums it up pretty well, actually.

And yes... Gonna be hard to get a PR merged if @danieleds is doing backflips in his grave. xD

Teklad:
Anyways... try to see if this last one fixes the focus bug the rest of the way... I opened nqq about 10 times without incident... but my computer isn't the fastest either. 

We'll focus on fixing docengine to behave better in the next PR of webengine related changes.

Teklad:
@danieleds, could you code review this when you get a chance?  I don't really wanna add anything else to this PR since it's already pretty large, and the tasks at this point are starting to get outside of the actual PR's scope.

danieleds:
Yep, reviewing.

danieleds:
Oh, a recap of all the changes would help.

danieleds:
The diff says some files in the submodule has been updated. Can you check why? Can you undo it for now?

src/ui/EditorNS/editor.cpp
```diff
@@ -23,53 +24,61 @@ namespace EditorNS
...
         QWidget(parent)
     {
-
-        QString themeName = NqqSettings::getInstance().Appearance.getColorScheme();
-        if (themeName == "") {
-            themeName = "default";
-        }
-
-        fullConstructor(themeFromName(themeName));
+        fullConstructor();
```
danieleds: I think the content of fullConstructor() can now be moved within the real constructor.

src/ui/EditorNS/editor.cpp
```diff
@@ -23,53 +24,61 @@ namespace EditorNS
...
+        vbox->setContentsMargins(0, 0, 0, 0);
+        vbox->setSpacing(0);
+        vbox->addWidget(m_webView, 1);
+        setLayout(vbox);
+
         connect(m_webView->page(),
-                &QWebEnginePage::loadFinished,
+                &QWebEnginePage::loadStarted,
                 this,
                 &Editor::on_javaScriptWindowObjectCleared);
```
danieleds: This event was called loadFinished and was connected to on_javaScriptWindowObjectCleared.
Now it is called loadStarted and it is still connected to on_javaScriptWindowObjectCleared.

Are we sure the semantic is still correct?

src/ui/EditorNS/editor.cpp
```diff
@@ -125,6 +134,20 @@ namespace EditorNS
...
 
+    void Editor::on_loadFinished()
+    {
+        updateBackground();
+    }
+
+    void Editor::updateBackground()
+    {
+        QString js = "rgb2hex($('.cm-s-'+editor.getOption('theme')).css('background-color'));";
+        m_webView->page()->runJavaScript(js, [&] (const QVariant& v){
```
danieleds: I'd prefer asking javascript to give us the background color by calling an already defined js function, instead of sending that string which is out of context.

src/ui/EditorNS/editor.cpp
```diff
@@ -268,117 +305,184 @@ namespace EditorNS
...
     }
 
-    QList<QMap<QString, QString>> Editor::languages()
+    void Editor::initLanguageCache()
     {
-        // TODO: Cache this data.
-        QJSEngine engine;
         QFileInfo fileInfo(Notepadqq::editorPath());
-        QString fileName = fileInfo.absolutePath() + "/classes/Languages.js";
+        QString fileName = fileInfo.absolutePath() + "/classes/Languages.json";
```
danieleds: Languages.json isn't a class anymore: it should probably be in `fileInfo.absolutePath() + "/Languages.json";`

Teklad:
It was because I had updated CodeMirror prior and reverted it at your request, lol.

src/ui/EditorNS/editor.cpp
```diff
@@ -268,117 +305,184 @@ namespace EditorNS
...
-            QMap<QString, QVariant> mode = lang.value().toMap();
-
-            QMap<QString, QString> newMode;
-            newMode.insert("id", lang.key());
-            newMode.insert("name", mode.value("name").toString());
-            newMode.insert("mime", mode.value("mime").toString());
-            newMode.insert("mode", mode.value("mode").toString());
-            langs.append(newMode);
+    QVariant Editor::getLanguageVariantData(const QString& langId)
+    {
```
danieleds: Is this method used? In case it is, can we give it a better name?

src/ui/EditorNS/editor.cpp
```diff
@@ -268,117 +305,184 @@ namespace EditorNS
...
             setIndentationMode(language);
+        }
+        QString mode;
+        if (m_info.language.mime.isEmpty()) {
+            mode = m_info.language.mode;
+        } else {
+            mode = m_info.language.mime;
+        }
+        sendMessage("C_CMD_SET_LANGUAGE", mode);
+        emit languageChanged(m_info.language);
```
danieleds: When someone receives this signal, is it guaranteed that CodeMirror did already update the language?

src/ui/EditorNS/editor.cpp
```diff
@@ -268,117 +305,184 @@ namespace EditorNS
...
+                if (l.fileExtensions.contains(ext, Qt::CaseInsensitive)) {
+                    setLanguage(l.id);
+                    return;
+                }
+            }
+        }
+    }
+/*
+    void Editor::detectLanguageFromContent()
+    {
```
danieleds: Why was this commented?
What happens now with script files without an extension but with a shebang?

src/ui/include/EditorNS/editor.h
```diff
@@ -267,29 +431,99 @@ namespace EditorNS
...
 
+        /**
+         * @brief Set the indentation mode for the current document.
+         * If transport is set to false, the data isn't sent to javascript.
+         * @param useTabs
+         * @param size
+         * @param custom
+         * @param transport
+         */
+        void setIndentationMode(const bool useTabs, const int size,
```
danieleds: Why shouldn't the data be sent to javascript?

src/ui/EditorNS/editor.cpp
```diff
@@ -678,52 +778,16 @@ namespace EditorNS
         sendMessage("C_CMD_SET_TABS_VISIBLE", visible);
     }
 
-    Editor::IndentationMode Editor::detectDocumentIndentation(bool *found)
-    {
```
danieleds: Is there a similar equivalent method now?

src/ui/EditorNS/editor.cpp
```diff
@@ -755,6 +819,5 @@ namespace EditorNS
             }
         });
     }
-
-
+QVector<Editor::LanguageData> Editor::m_langCache;
```
danieleds: weird position

src/ui/include/EditorNS/editor.h
```diff
@@ -317,47 +550,72 @@ namespace EditorNS
...
+
+
+        /**
+         * @brief  Workaround. Using isClean() right after creating an Editor always returns true.
+         *         This value gets set to 'true' if the Editor starts out as a newly created empty
+         *         document, and is 'false' if the document was loaded from any file.
+         * @return
+         */
+        bool isInitiallyBlank() const { return m_initiallyBlank; }
+        void setInitiallyBlank(bool blank) { m_initiallyBlank = blank; }
```
danieleds: Hmm, can you elaborate?

src/ui/EditorNS/editor.cpp
JuBan1: Some context is here: https://github.com/notepadqq/notepadqq/pull/403#pullrequestreview-28507410 (the review plus the 2nd and 3rd comment right below it)

src/ui/include/EditorNS/editor.h
Teklad: @JuBan1, care to comment on this one?  I think this was your fix for docengine being stupid

src/ui/include/EditorNS/editor.h
Teklad: It was to make it a multi-purpose function where we could initialize setting without needing a secondary function... though I could probably just write an initDefaults type function for this purpose for when it needs to be extended in the future.

src/ui/EditorNS/editor.cpp
Teklad: It may be unused now that I have the LanguageData struct... I'll have to double check to verify before removing.

src/ui/EditorNS/editor.cpp
Teklad: Agreed.

src/ui/include/EditorNS/editor.h
JuBan1: Right now `DocEngine` closes an empty tab when opening a new tab if that empty tab was the only open tab. That is reasonably but goes wrong phenomenally with the new asynchronous Javascript. Reason for that is that when `DocEngine` loads several documents at once (like it does when restoring tabs at startup) it'll check whether an `Editor` is empty right after creating it. At that point the JavaScript hasn't completely initialized yet and `isClear()` returns its default value of `true`.

So a way was needed to determine whether an `Editor` was created as an empty document (in which case it can be closed by `DocEngine`) or not. `isInitiallyBlank` does that.

src/ui/EditorNS/editor.cpp
Teklad: Probably a good idea... might create a struct that pulls all the theme data, if you think it would be useful... otherwise we'll just turn the background portion into its own function.

src/ui/EditorNS/editor.cpp
Teklad: QWebChannel prefers to be loaded as soon as possible, this is opposite of how QWebKit did things, so it is correct.

src/ui/include/EditorNS/editor.h
JuBan1: PS: This stuff can be properly cleaned up when `DocEngine` and friends get their chance of being reworked a little. The `loadDocuments()` function is hardly in the business of *closing* documents at its leasure. But that's reserved for another PR; `isInitiallyClean()` is a patch until the root of the problem is removed.

src/ui/EditorNS/editor.cpp
Teklad: Agreed again. :)

src/ui/EditorNS/editor.cpp
Teklad: We could change it to "C_FUN_SET_LANGUAGE" and have it return true on success... then use sendMessageWithCallback to ensure its set.

Teklad: Actually, scratch that... the editor.on("option") event in the new CodeMirror will solve this issue as well.

src/ui/EditorNS/editor.cpp
Teklad: This was literally only used for banner displays, which now operates on a signal rather than a function call.  We can always re-implement it later if the need arises (over-engineering is the devil).

src/ui/include/EditorNS/editor.h
Teklad: @JuBan1, the great thing about the new design I have planned for that includes extremely useful things, such as being able to queue page loading to the background, and emitting a signal telling it to create a tab when it's done (This should give sessions much better integration). 

Another useful thing we could do to speed up the initial load (another PR in itself), would be to thread DocEngine once we get all the GUI cruft out of it, which would stop it from blocking the UI load during session initialization.

Teklad:
I implemented a more reliable method which will signal when specified options are changed.  This gives a guarantee that the data is valid before the signal is emitted.

src/ui/EditorNS/editor.cpp
danieleds: :+1:

src/ui/EditorNS/editor.cpp
danieleds: :+1:

src/ui/EditorNS/editor.cpp
Teklad: I figured out a way to do this... I'm going to push momentarily with my language cache fixes.  Turns out we were copying the language structs about 800 times over between mainwindow and frmpreferences.

I "think" I got everything working.

danieleds:
> I implemented a more reliable method which will signal when specified options are changed

Wait wait, I can't review if more stuff is added in the meantime :sweat_smile:
Does that directly fix an issue of those I highlighted?

Teklad:
It made it where I could do the language detection by content without breaking my neck, so yes. :P

Teklad:
I could save the language detection fix for the next PR, if you like.... I don't mind, lol.

Teklad:
Oh... you were referring to a different thing... that was to get rid of the timer crap in updateBackground... its extremely unreliable to do things and "hope" its updated by the time you call upon it, lol.  It applies to the actual JS-side of things anyways.

I'm completely done poking at stuff now.  Promise!

src/ui/include/EditorNS/editor.h
danieleds: I'd like to avoid any possible way of letting the state of the js part be not in sync with that of the c++ part. Why would we want to do that?

src/ui/include/EditorNS/editor.h
danieleds: So the real fix will be to make DocEngine asynchronous too, if I got it right? And then we will be able to remove these methods?

Can we solve it by calling a synchronous version of `isClean()` for now?

danieleds:
lol, ok! :P

src/ui/include/EditorNS/editor.h
Teklad: To initialize values before the javascript side is ready, mostly.

src/ui/include/EditorNS/editor.h
Teklad: The real fix will be to get DocEngine's fingers out of the editor, lol.  It doesn't necessarily have to be asynchronous in order to do that.  It just doesn't need to be directly connected to the editor and tabs like it is now.

Teklad: Also, DocEngine calls isClean too soon, which is why it obliterates the 1st tab thinking it's a "New, unmodified document", and waitAsyncLoad kills performance... so until we can fix this proper, @JuBan1's solution is probably the best we're gonna get.

Teklad: Or we could just remove DocEngine's auto-close function until we fix it... that would work too. :)

Teklad: I just tested this in loadSession, which seems to circumvent the issue for now with minimal interference:
```cpp
            tabW->editor(0)->setFileName(QUrl("tmp"));
            const bool success = docEngine->loadDocumentSilent(loadUrl, tabW);
            tabW->editor(0)->setFileName(QUrl());
```

That'll keep the hack contained to a single location until we can fix it proper.  I could just add a FIXME here so we don't forget.

Teklad:
There we go... I think that covers pretty much everything you had highlighted.  The work-around I applied for the DocEngine destroying the first tab is contained in one place so we can easily remove it later once we fix DocEngine to not poke at tabs anymore.  

That'll probably be after all this WebEngine mess though, since reworking that stuff is another big project on its own

danieleds:
Perfect!

I'd like to give it just another look before merging.

src/editor/classes/AppManager.js
```diff
@@ -0,0 +1,121 @@
+//"use strict";
```
danieleds: I think this should be uncommented to be consistent with the other files.

src/editor/classes/AppManager.js
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * @brief This is our controller class.  All calls are made through here
+ *        so we have a consistent interface for javascript communication.
+ * 
+ * Some brief examples of how to use this:
+ * App.proxy.sendEvent(Message, Data);
+ * App.eventHook.onChange();
+ *
```
danieleds: Add more information... for example why and when should I call `App.eventHook.onChange()` or `App.proxy.sendEvent(Message, Data);`.

You can use realistic examples to give the idea.

src/editor/classes/AppManager.js
```diff
@@ -0,0 +1,121 @@
...
+        let themeName = getQueryValue("themeName");
+        if (themePath !== "") {
+            this.setEditorStyleSheet(themePath);
+            this.defaultTheme = themeName;
+        }
+        $(document).ready(() => {this.initializeCodeMirror()});
+    }
+
+    setEditorStyleSheet(sheet)
+    {
```
danieleds: It seems that this method doesn't just set a stylesheet, but actually changes the theme of CodeMirror (adding a stylesheet as a side effect).

A better name would be setTheme, with a comment explaining that the associated stylesheet will be loaded at runtime.

src/editor/classes/CommunicationsManager.js
```diff
@@ -0,0 +1,93 @@
...
+        this.sendEvent("J_EVT_CLEAN_CHANGED", data);
+    }
+
+    this.contentChangedEvent = data => {
+        this.sendEvent("J_EVT_CONTENT_CHANGED", data);
+    }
+
+    this.cursorActivityEvent = data => {
+        this.sendEvent("J_EVT_CURSOR_ACTIVITY", data);
+    }
```
danieleds: These functions are here just for convenience, I guess?
But they're not actually something which is *needed* for communication, so they shouldn't be in CommunicationsManager (except `setReady()`).

Teklad: They're there to keep from unexpected typos of J_EVT_XX_X.  Spelling errors are a soft fail (No output).  This guarantees an error will be thrown.  But they are also directly linked with communication.  I can't think of anywhere else they would make sense... We could put them directly in App, but I'm hesitant to do that because App is supposed to be an interface for all this.

src/editor/classes/AppManager.js
Teklad: Actually, it only looks that way.  It swaps the css file then updates the C++ ui.  It then requests the current theme from CodeMirror and sends a message to C++ so we can keep the background data in sync.  So it's literally setting the stylesheet on both sides of the fence, so to speak.

danieleds: Nevermind, I totally misread...

src/editor/classes/MessageHandler.js
danieleds: @Teklad

Teklad: This comment is no longer relevant since I moved helpers to a plain functions file.  It all works as-is

src/ui/editortabwidget.cpp
```diff
@@ -39,6 +40,18 @@ int EditorTabWidget::addEditorTab(bool setFocus, const QString &title)
     return this->rawAddEditorTab(setFocus, title, 0, 0);
 }
 
+bool EditorTabWidget::eventFilter(QObject* object, QEvent* ev)
+{
+    if(object == tabBar() && ev->type() == QEvent::FocusIn) {
```
danieleds: This fix could have been done in a different PR so that it would have been usable in the master branch too... Please don't mix different stuff in the same PR. There are a lot of reasons for not doing that.

I'll turn a blind eye for this time ;)

danieleds:
Good news, I've managed to review almost everything and it looks fine.
I only need to go through editor.cpp asap.

I've tagged you in a comment brought up by @JuBan1, and then I want to think about those functions in CommunicationsManager.

After that... merge time!

src/ui/editortabwidget.cpp
Teklad: Status bar updates weren't operating properly... so it seemed necessary considering the JS was failing to be called because of it.... I'll try to be a little less heavy-handed with my code in the future...  I'm highly aware that most of my PRs are extensive in scope, lol.

Teklad:
@danieleds I already replied to that comment you tagged me in... went a different route with that so it should be fine there.  I'm just ready to get this merged so I can move on to the next painful thing I have to smack around to get WebEngine working. -.-

src/ui/include/EditorNS/editor.h
```diff
@@ -33,36 +36,32 @@ namespace EditorNS
...
+         * @return
+         */
         static QSharedPointer<Editor> getNewEditor(QWidget *parent = 0);
         static Editor *getNewEditorUnmanagedPtr(QWidget *parent);
-
         static void invalidateEditorBuffer();
 
+        //FIXME: Possibly un-necessary?
+        /*
         struct LanguageGreater {
```
danieleds: who sorts languages now?

danieleds:
Good to know! :D

Okay then, let's also forget about that stuff in CommunicationsManager for now.

src/ui/include/EditorNS/editor.h
Teklad: Do they actually need to be sorted?  From what I can tell QMap is sorting everything anyways... exactly like the sorting function was doing, lol.

danieleds:
Hmm, segfault on startup at languagecache.cpp#94... I need to investigate.

Teklad:
@danieleds, that's new to me... I tested the cache heavily without any oddities.

Teklad:
Care to share the full gdb output on that?

danieleds:
```
GNU gdb (Ubuntu 7.11.90.20161005-0ubuntu1) 7.11.90.20161005-git
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from out/debug/lib/notepadqq-bin...done.
(gdb) run
Starting program: /home/daniele/Progetti/qt/notepadqq/out/debug/lib/notepadqq-bin 
warning: the debug information found in "/lib64/ld-2.24.so" does not match "/lib64/ld-linux-x86-64.so.2" (CRC mismatch).

[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Start-time benchmark started.
WARNING: Notepadqq is running in DEBUG mode.
[New Thread 0x7fffe2439700 (LWP 16202)]
[New Thread 0x7fffe180e700 (LWP 16203)]
[0401/232102:ERROR:browser_main_loop.cc(217)] Running without the SUID sandbox! See https://chromium.googlesource.com/chromium/src/+/master/docs/linux_suid_sandbox_development.md for more information on developing with the sandbox on.
[New Thread 0x7fffd97f7700 (LWP 16206)]
[New Thread 0x7fffd8ff6700 (LWP 16211)]
[New Thread 0x7fffcffff700 (LWP 16212)]
[New Thread 0x7fffcf7fe700 (LWP 16213)]
[New Thread 0x7fffceffd700 (LWP 16214)]
[New Thread 0x7fffce7fc700 (LWP 16215)]
[New Thread 0x7fffcdffb700 (LWP 16216)]
[New Thread 0x7fffcd7fa700 (LWP 16217)]
[New Thread 0x7fffccff9700 (LWP 16218)]
[New Thread 0x7fffabfff700 (LWP 16219)]
[New Thread 0x7fffab7fe700 (LWP 16220)]
[New Thread 0x7fffaaffd700 (LWP 16221)]
[New Thread 0x7fffaa7fc700 (LWP 16222)]
[New Thread 0x7fffa9ffb700 (LWP 16223)]
[New Thread 0x7fffa8ff9700 (LWP 16225)]
[New Thread 0x7fffa97fa700 (LWP 16224)]
[New Thread 0x7fff93fff700 (LWP 16226)]
[New Thread 0x7fff937fe700 (LWP 16227)]
[New Thread 0x7fff929ad700 (LWP 16228)]
[New Thread 0x7fff921ac700 (LWP 16229)]
Extension support is not installed.
[New Thread 0x7fff90d87700 (LWP 16234)]

Thread 1 "notepadqq-bin" received signal SIGSEGV, Segmentation fault.
0x00007ffff5904b20 in QJsonObject::keyAt(int) const () from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Core.so.5



(gdb) bt
#0  0x00007ffff5904b20 in QJsonObject::keyAt(int) const () from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Core.so.5
#1  0x00005555555e9bac in QJsonObject::const_iterator::key (this=0x7fffffffc790)
    at /home/daniele/Qt/5.7/gcc_64/include/QtCore/qjsonobject.h:168
#2  0x00005555555e8d03 in EditorNS::LanguageCache::LanguageCache (
    this=0x5555558e0190 <EditorNS::LanguageCache::getInstance()::instance>) at ../ui/EditorNS/languagecache.cpp:94
#3  0x00005555555e99c8 in EditorNS::LanguageCache::getInstance () at ../ui/EditorNS/languagecache.cpp:161
#4  0x00005555555daadb in EditorNS::Editor::getLanguage (this=0x555555abbea0) at ../ui/EditorNS/editor.cpp:301
#5  0x0000555555589616 in MainWindow::refreshEditorUiInfoAll (this=0x555555d65450, editor=0x555555abbea0)
    at ../ui/mainwindow.cpp:1150
#6  0x0000555555588bc2 in MainWindow::on_currentEditorChanged (this=0x555555d65450, tabWidget=0x555555d56dd0, tab=0)
    at ../ui/mainwindow.cpp:1081
#7  0x00005555555b86fd in QtPrivate::FunctorCall<QtPrivate::IndexesList<0, 1>, QtPrivate::List<EditorTabWidget*, int>, void, void (MainWindow::*)(EditorTabWidget*, int)>::call (f=
    (void (MainWindow::*)(MainWindow * const, EditorTabWidget *, int)) 0x555555588b86 <MainWindow::on_currentEditorChanged(EditorTabWidget*, int)>, o=0x555555d65450, arg=0x7fffffffcb90)
    at /home/daniele/Qt/5.7/gcc_64/include/QtCore/qobjectdefs_impl.h:507
#8  0x00005555555b71aa in QtPrivate::FunctionPointer<void (MainWindow::*)(EditorTabWidget*, int)>::call<QtPrivate::List<EditorTabWidget*, int>, void> (f=
    (void (MainWindow::*)(MainWindow * const, EditorTabWidget *, int)) 0x555555588b86 <MainWindow::on_currentEditorChanged(EditorTabWidget*, int)>, o=0x555555d65450, arg=0x7fffffffcb90)
    at /home/daniele/Qt/5.7/gcc_64/include/QtCore/qobjectdefs_impl.h:526
#9  0x00005555555b4141 in QtPrivate::QSlotObject<void (MainWindow::*)(EditorTabWidget*, int), QtPrivate::List<EditorTabWidget*, int>, void>::impl (which=1, this_=0x555555e43da0, r=0x555555d65450, a=0x7fffffffcb90, ret=0x0)
    at /home/daniele/Qt/5.7/gcc_64/include/QtCore/qobject_impl.h:149
#10 0x00007ffff594877d in QMetaObject::activate(QObject*, int, int, void**) ()
   from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Core.so.5
#11 0x0000555555629a93 in TopEditorContainer::currentEditorChanged (this=0x555555d57410, _t1=0x555555d56dd0, _t2=0)
    at ../../out/build_data/moc_topeditorcontainer.cpp:316
#12 0x00005555555bb0b9 in TopEditorContainer::on_currentTabChanged (this=0x555555d57410, index=0)
    at ../ui/topeditorcontainer.cpp:103
#13 0x00005555555bc848 in QtPrivate::FunctorCall<QtPrivate::IndexesList<0>, QtPrivate::List<int>, void, void (TopEditorContainer::*)(int)>::call (f=
    (void (TopEditorContainer::*)(TopEditorContainer * const, int)) 0x5555555baff6 <TopEditorContainer::on_currentTabChanged(int)>, o=0x555555d57410, arg=0x7fffffffcdd0) at /home/daniele/Qt/5.7/gcc_64/include/QtCore/qobjectdefs_impl.h:507
#14 0x00005555555bc6cd in QtPrivate::FunctionPointer<void (TopEditorContainer::*)(int)>::call<QtPrivate::List<int>, void> (f=
    (void (TopEditorContainer::*)(TopEditorContainer * const, int)) 0x5555555baff6 <TopEditorContainer::on_currentTabChanged(int)>, o=0x555555d57410, arg=0x7fffffffcdd0) at /home/daniele/Qt/5.7/gcc_64/include/QtCore/qobjectdefs_impl.h:526
#15 0x00005555555bc265 in QtPrivate::QSlotObject<void (TopEditorContainer::*)(int), QtPrivate::List<int>, void>::impl (
    which=1, this_=0x555555d6d440, r=0x555555d57410, a=0x7fffffffcdd0, ret=0x0)
    at /home/daniele/Qt/5.7/gcc_64/include/QtCore/qobject_impl.h:149
#16 0x00007ffff594877d in QMetaObject::activate(QObject*, int, int, void**) ()
   from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Core.so.5
#17 0x00007ffff73cc4ee in QTabWidget::currentChanged(int) () from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Widgets.so.5
#18 0x00007ffff73ce5e7 in ?? () from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Widgets.so.5
#19 0x00007ffff5948056 in QMetaObject::activate(QObject*, int, int, void**) ()
   from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Core.so.5
#20 0x00007ffff73c052e in QTabBar::currentChanged(int) () from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Widgets.so.5
#21 0x00007ffff73c53b0 in QTabBar::setCurrentIndex(int) () from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Widgets.so.5
#22 0x00007ffff73ca36b in QTabBar::insertTab(int, QIcon const&, QString const&) ()
   from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Widgets.so.5
#23 0x00007ffff73cd0ac in QTabWidget::insertTab(int, QWidget*, QIcon const&, QString const&) ()
   from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Widgets.so.5
#24 0x00007ffff73cd154 in QTabWidget::insertTab(int, QWidget*, QString const&) ()
   from /home/daniele/Qt/5.7/gcc_64/lib/libQt5Widgets.so.5
#25 0x00005555555bd3f8 in EditorTabWidget::rawAddEditorTab (this=0x555555d56dd0, setFocus=false, title=..., 
    source=0x0, sourceTabIndex=0) at ../ui/editortabwidget.cpp:132
#26 0x00005555555bcccc in EditorTabWidget::addEditorTab (this=0x555555d56dd0, setFocus=false, title=...)
    at ../ui/editortabwidget.cpp:40
#27 0x00005555555c1b11 in DocEngine::loadDocuments (this=0x555555d34040, fileNames=..., tabWidget=0x555555d56dd0, 
    reload=false, codec=0x0, bom=false, rememberLastSelectedDir=false) at ../ui/docengine.cpp:164
#28 0x00005555555c16c0 in DocEngine::loadDocumentSilent (this=0x555555d34040, fileName=..., tabWidget=0x555555d56dd0)
    at ../ui/docengine.cpp:110
#29 0x000055555561d3de in Sessions::loadSession (docEngine=0x555555d34040, editorContainer=0x555555d57410, 
    sessionPath=...) at ../ui/Sessions/sessions.cpp:384
#30 0x0000555555581c52 in MainWindow::MainWindow (this=0x555555d65450, workingDirectory=..., arguments=..., parent=0x0)
    at ../ui/mainwindow.cpp:138
#31 0x0000555555582186 in MainWindow::MainWindow (this=0x555555d65450, arguments=..., parent=0x0)
    at ../ui/mainwindow.cpp:191
#32 0x000055555557eca3 in main (argc=1, argv=0x7fffffffde28) at ../ui/main.cpp:122
(gdb) 
```

src/ui/include/EditorNS/editor.h
danieleds: Perfect! If that's the case, then we can remove it.

danieleds:
Quite interestingly, it doesn't happen if you replace the iterator with this (optimizable) code:

```c++
    for (QString key : json.object().keys()) {
        QJsonObject mode = json.object().value(key).toObject();
        Language newMode;
        newMode.id = key;
        newMode.name = mode.value("name").toString();
        newMode.mime = mode.value("mime").toString();
        newMode.mode = mode.value("mode").toString();
        newMode.fileNames = mode.value("fileNames")
            .toVariant().toStringList();
        newMode.fileExtensions = mode.value("fileExtensions")
            .toVariant().toStringList();
        newMode.firstNonBlankLine = mode.value("firstNonBlankLine")
            .toVariant().toStringList();
        m_languages.append(std::move(newMode));
    }
``

Teklad:
I'm gonna guess it's a bug in the iterator for QT 5.7... which isn't a big deal, but it's annoying, nonetheless.... one sec and I'll modify to use your code.

danieleds:
I believe that too. Qt bugs are starting to get annoying.

Teklad:
Try this commit and see if it fixes it for you.

Teklad:
There's a lot of black magic under the hood of QT that I'm not very fond of... It doesn't bother me to write a little extra code to make something work the way I want it... I don't need QT implying "optimizations" for me.

I started programming before we had fancy smart pointers and such, after all. :P  Oh the days of malloc() and free() were simpler times.

danieleds:
Yup, working.
But does the assignment of `json.object().keys()` to a variable actually increase performance? I thought the range expression in the range based loop was only evaluated once.

Teklad:
Is it?  It might not increase performance in that case... but that "eval once" optimization might not apply to every compiler under the sun, so it doesn't hurt anything at least.

Teklad:
I think it actually slows it down a bit with it as a variable (odd)... removed it. :)

danieleds:
Actually, the semantic would be completely different in that case. I think that behavior must be well defined somewhere, as it's not a decision that should be left to the compiler implementation.

Actually, you can find it in the official specification, between page 141 and 142: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf

(just nitpicking :P)

Teklad:
The performance is very decent, language initialization is roughly 1/100th of a second(on my hardware) on average... and lookups are on a scale I can't really measure accurately, as they should be.

The only way to make it faster at this point would be to write a JSON parser from scratch and read it directly... but I'm not that hellbent on juicing something so simple, lol.

Teklad:
Also, initialization doesn't happen until the first call of getInstance(), which keeps it self-contained, much like nqqsettings

danieleds:
Yeah it's not a performance-critical portion of code.

I was pointing it out just because of your old comment above that line, which was fine for the old loop but now it wasn't true anymore :P

Can you delete that commented stuff which was used to sort languages?

Teklad:
Done.

Teklad:
Before we close this PR... should probably touch on what else needs to be done to complete QWebEngine... we're getting reasonably close, I think.  I'm just not sure if there are anymore barriers that we're unaware of as of yet.

danieleds:
I still have problems with performance.

I guess that now is a good time for little fixes (I was some glitches somewhere) and smart performance optimizations.

Other than that, it seems that the difficult part is over.

Issue #406, opened 2017-03-21
Advanced Search & Replace design mockup


JuBan1:
I've wanted to do this for a long time but now I've finally finished the mockup.

The idea is to implement a Search & Replace dialog more akin to Qt Creator's and roughly as powerful. I've not code yet but I've put together a shoddy mockup of the interface.

Here's a full shot:

![full image](http://i.imgur.com/ooPnzwr.png)

With expanded search history:

![search history](http://i.imgur.com/CWghuhV.png)

With expanded scope combobox:

![](http://i.imgur.com/Ov31nc3.png)

When conducting a search (or picking one from the search history), the bottom part is replaced by the search results. I intend to slightly modify them as well. For now they're the same as we have right now:

![](http://i.imgur.com/HMDUz9Y.png)

Clicking the "Replace Options" button in the header bar will toggle a bar for replace actions:

![](http://i.imgur.com/sXzoxnz.png)

The UI is just hacked together at the moment, hence a few visual inconsistencies.

Thoughts and suggestions? Before I start coding I want to get the UI right and make we're all happy with it.

JuBan1:
PS: some more info:

* This is intended to replace the "Find in files..." search option we have right now. The normal search and replace dialog is still useful because it's quicker to use than this one. (Like QtCreator does it too).

* Search history is not meant to be saved across restarts. Will just be wiped at exit.

* The "<" and ">" buttons are "Find Previous" and "Find Next", respectively.

JuBan1:
Obligatory edit for spamming you guys with messages: @Teklad @danieleds

danieleds:
Just one quick comment: it looks awesome.

Teklad:
I thought this email said ham. :(.  Looks very good though!

PR #410, opened 2017-03-25
[Do Not Merge] Preview of Advanced Search & Replace dialog

JuBan1:
This is the first step to getting the new search & replace system merged. This PR contains the finished user interface with most ui interactions enabled but with only very limited functionality (no searching yet, etc).

My goal is to get the ui right before starting on functionality so I'd like you to try the PR @danieleds @Teklad . Do a search, click around a bit and see if anything should be changed, if any functionality should be added, or if anything is confusing to use.

(For the purpose of testing, the search dialog will always open when Nqq is started.)

P.S. I'm also interested in whether the ui looks good on different monitor resolutions. I only have a 1440p display here. If anyone has something smaller or larger to see if everything looks good, that'd by great.


danieleds:
That beautiful awesome UI which you showed in #406 becomes this when using the obscene Ubuntu default theme:

screenshot

:(

Teklad:
Technically speaking we could pull values from the active codemirror theme and apply it to the search UI.  That would make everything pretty sexy I thinks.

danieleds:
Interesting idea!

JuBan1:
Yay, some responses! :)

While pulling values from the theme is possibility I'm not sure if it's a good idea:

*  It's inconsistent with other parts of the Ui unless we turn *everything* into the CodeMirror color scheme.
* I wouldn't be surprised if many CodeMirror themes would produce a terrible-looking Ui (since they come in pretty much all colors).
* IIRC commonly all program GUIs should respect the DE's color scheme.

An alternative would be to offer a "dark theme" and a "default theme" for the Ui.

Teklad:
Its hard to tell a GTK/QT theme "Hey, don't be stupid on X program"... In my experience the best thing to do is simply throw out the theme for something that behaves better.

There's a really large number of "shoddy" themes available for GTK and QT, after all. Lol.

danieleds:
There is one thing that I would like to highlight right now, as I'm not used yet with this UI.

The first times that I tried it, it was not immediately intuitive to me what I should have pressed to search what I wanted to search.

I mean, at first sight the focus (in graphic terms) seems to be on that "New Search" dropdown, while I *feel* it should have been on the "Search" textbox.

As a user, the first thing I want to do is write the search string. Let me do that as fast as possible. Second thing, I want to say *where* I should search (current doc? file system?). Third thing, in case I want to search in files, I want to specify *the folder* in which to search, and then some pattern to filter the files.

Notice how the current dialog elegantly adheres to these rules:

screen1

You can see how the user attention is captured by that first text field. The user opens the dialog and he can start writing the search string without even thinking. Then he realizes *where* he wants to search and clicks over the "Search in files" tabs.
The search string is preserved, but the dialog becomes this:

screen2

These are very subtle design choices which I believe can make the difference in the usability of the program, because then the program follows the user's way of thinking, instead of the other way around.

Comments / ideas?

JuBan1:
I get what you mean- after using the QtCreator search for a while I have no idea anymore whether I found it intuitive or not at the beginning.

So, first question is how to draw the user's focus on the essential stuff. First things that came to mind are:

* Create a stronger visual separation between the titlebar and search panel. (Done that by adding a separating line)
* Move the search field to the top of the panel where users would generally look first.
* Give focus (the Qt focus) to the search field once the user opens the search widget or switches to a "New Search"

Changes here: 
image

Second, how to make the flow of input more intuitive.

* Putting the search field first is a good start.
* Move the directory field up one notch.
* Now users can tab through all fields pretty much by order of importance (what -> where -> where specifically -> which files)
* Also, I changed the text for search scopes a little (e.g. from "Scope: Current Document" to "Search in Current Document") which may be more intuitive.

image

Those are very minor adjustments but I think they improve things without making me have to redo larger parts of the UI (which is a plus). Larger changes are okay too if we come up with something concrete.

danieleds:
That's a whole lot better!!

I'll write as soon as I find other observations.

JuBan1:
Here's a relatively big update. In addition to the ui changes discussed above, here's a summary of changes:

* Searching still only returns dummy results. The dummy searches take a few seconds to finish to illustrate how searching a larger directory would look like. Searches show their progress and can be canceled at any time.
* Lots of buttons now work: 
    * Previous/Next result to iterate through individual matches,
    * Redo Search opens the search panel and fills in the previous search's parameters (search term, etc),
    * Copy Selected Contents copies all selected matches to the clipboard. 

With this the ui is almost completely functional. What's left are all the interactions with other systems (actually searching, replacing, going to result location) and a few minor things like shortcuts and proper focus sequence.

Give it another try if you'd like.

Next steps will be to turn the code from the pile of junk it is now into something nice and to integrate file system search. Search in open documents will probably follow once we've got WebEngine all figured out.

Teklad:
We might consider having a dedicated thread for all the search workers: searchinfilesworker/searchworker/extras.

Creating threads over and over can have a bit of overhead, so it might not hurt as something to think about for the long-term of things.

JuBan1:
Not 100% sure how the worker class will end up looking like yet. I'd like to re-use lots of SearchInFilesWorker but changes need to be done to support the new interface and functionality (like searching in open documents).

I want to experiment with splitting the workload up into multiple workers to see if it'll have a positive impact on overall search time. Maybe keeping thread objects alive makes some difference for 8ish workers but I doubt there would be any noticeable effect when just one thread is started.

JuBan1:
Here's there next large update.

Searching the file system is now enabled and on nearly the same level as our old Find In Files feature. Replacing doesn't yet work. Searching in opened documents neither. Almost everything else is working although there are still a few rough edges.

I'm quite happy with progress so far. When searching large directories the UI gives you feedback on how much it already searched but there is no popup so you're free to do other stuff (and even start more searches) in the mean time.

Here's a sample of the UI while a search is underway and of the result tree.

image

image

JuBan1:
Phew, alright. After a few weeks of hiatus I've made good progress the last days.

Filesystem Search & Replace is now done with all features that were planned. S&R for documents will follow but I would like input and some extra eyes to make sure everything works as intended @danieleds @Teklad.


I've also done some result verification and performance testing, and the results look good. The following was taken from a search through a 4.7GiB folder with 123,000 files:

Version | Time | Results Found
------------ | ------------- | -------------
Current Master | 10500ms | 4484 hits
Advsearch Branch | 10300ms | 4484 hits

PR #414, opened 2017-04-02
Webengine misc cleanup.

Teklad:
Just some minor cleanup, removing unused headers... mild tweaks to code to use a few more C++11 constructs.  And a small bug fixes where I forgot to include "real" files in the DocEngine work-around.

On a side note, we probably need to find a better way to do on_proxyMessageReceived... its starting to look pretty ugly.  Might write a "function map", which would route the message to the appropriate function.  This would be similar to how we do it on the JS side.

Another side note: I also came up with a way to delay resizing of the QWebEngineView until the resize is finished (buttery smooth resize)... but its a little hacky so I'm going to try a few different things to see what works best there.  

Mainly getting the ACTUAL state of the mouse after a resize event has completed is darn near impossible using your typical QT functions.  QApplication::mouseButtons() returns the state when the event started, unfortunately.

We could use QAbstractEventDispatcher, but we'd need to write separate bits of code for each individual OS(Windows/Mac/Linux).

Setting up an app-wide eventFilter and keeping track of "is_resizing" would probably be a good way to go, though I'm not 100% on that.

### DIFF
--- a/src/ui/EditorNS/editor.cpp
+++ b/src/ui/EditorNS/editor.cpp
@@ -1,7 +1,6 @@
 #include <QDir>
 #include <QMessageBox>
 #include <QRegularExpression>
-#include <QStringBuilder>
 #include <QUrlQuery>
 #include <QVBoxLayout>
 #include <QWebEngineSettings>
@@ -58,7 +57,7 @@ namespace EditorNS
         query.addQueryItem("themePath", theme.path);
         query.addQueryItem("themeName", theme.name);
 
-        QUrl url = QUrl("file://" % Notepadqq::editorPath());
+        QUrl url = QUrl("file://" + Notepadqq::editorPath());
         url.setQuery(query);
         m_webView->setEnabled(false);
         m_webView->page()->load(url);
@@ -646,8 +645,8 @@ namespace EditorNS
             return theme;
 
         auto editorPath = QFileInfo(Notepadqq::editorPath());
-        auto bundledThemesDir = QDir(editorPath.absolutePath() % "/libs/codemirror/theme/");
-        auto themeFile = bundledThemesDir.filePath(name % ".css");
+        auto bundledThemesDir = QDir(editorPath.absolutePath() + "/libs/codemirror/theme/");
+        auto themeFile = bundledThemesDir.filePath(name + ".css");
         if (QFile(themeFile).exists()) {
             theme.name = name;
             theme.path = themeFile;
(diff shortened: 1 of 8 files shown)


Teklad:
@danieleds, you alive with us? :)

src/ui/EditorNS/editor.cpp
```diff
@@ -656,8 +646,8 @@ namespace EditorNS
             return theme;
 
         auto editorPath = QFileInfo(Notepadqq::editorPath());
-        auto bundledThemesDir = QDir(editorPath.absolutePath() + "/libs/codemirror/theme/");
-        auto themeFile = bundledThemesDir.filePath(name + ".css");
+        auto bundledThemesDir = QDir(editorPath.absolutePath() % "/libs/codemirror/theme/");
+        auto themeFile = bundledThemesDir.filePath(name % ".css");
```
danieleds: Same as before. Is this giving a visible performance improvement?

src/ui/main.cpp
```diff
@@ -101,7 +101,7 @@ int main(int argc, char *argv[])
 
     // There are no other instances: start a new server.
     a.startServer();
-    Editor::addEditorToBuffer();
+    //Editor::addEditorToBuffer();
```
danieleds: mistake? Or was it making performance worse?

danieleds:
Still alive, sorry for the long wait!

src/ui/main.cpp
Teklad: It no longer has any effect on performance and just eats more memory/increases startup time.

src/ui/EditorNS/editor.cpp
JuBan1: I don't think using `%` will give any improvements here. `QStringBuilder` delays construction of the `QString` until all concatenations a processed. But with just one concatenation there's nothing to delay.

Teklad: You're more than likely right... not sure what I was thinking here.

Teklad:
Alright... reverted the QStringBuilder change since it probably doesn't have any performance improvement.  The last bit was more-so to test and see if it made any difference related to startup/runtime performance originally.

The consensus I got on it was it no longer gives any performance increase at all... and actually causes notepadqq to eat more memory, which isn't desirable at this point.

I'm not really sure what else I can do for QWebEngine at this point outside of minor pokes and prods as they are requested.  It's (mostly) feature complete... and any remaining bugs are those that probably shouldn't be fixed in this scope. (DocEngine and Editor overhauls are way past due)

danieleds:
Perfect!

PR #427, opened 2017-04-23
Update configure script

Teklad:
Fixes having to use --qmake and --lrelease flags on systems that don't have qtchooser.

### DIFF
--- a/src/ui/mainwindow.ui
+++ b/src/ui/mainwindow.ui
@@ -21,7 +21,7 @@
    <bool>true</bool>
   </property>
   <property name="unifiedTitleAndToolBarOnMac">
-   <bool>true</bool>
+   <bool>false</bool>
   </property>
   <widget class="QWidget" name="centralWidget">
    <layout class="QVBoxLayout" name="verticalLayout">


Teklad:
this can be applied to master as well without issues, I think.

Teklad:
Okay... I'm done with this now.  I got rid of a lot of the hard-coded cruft in favor of more dynamic variables at build time... so this configure script should run on more systems now. :)

configure
```diff
@@ -125,63 +137,48 @@ check c++ "$CXX" 1
...
-test -d "$cmdir" || errorExit "$cmerror" 1
-# check if directory is empty
-test "$(ls -A $cmdir)" || errorExit "$cmerror" 1
-
-if [ -f "Makefile" ]; then
-    make distclean 2>/dev/null 1>/dev/null || true
-fi
-
+test -d "$cmdir" && test "$(ls -A $cmdir)" || \
+    errorExit "error: CodeMirror submodule not available at \`$cmdir'! Make sure you downloaded the source code using git as specified in README.md." 1
```
danieleds: prefer if/fi

configure
```diff
@@ -125,63 +137,48 @@ check c++ "$CXX" 1
...
 
+# Generate our makefile
+test -f "Makefile" && make distclean >/dev/null 2>&1 || true
```
danieleds: prefer if/fi

configure
```diff
@@ -125,63 +137,48 @@ check c++ "$CXX" 1
 rm -f test test.cpp
 cat << EOF > test.cpp
 #include <iostream>
-
-int main()
-{
-  std::cout << "Hello World!" << std::endl;
-  return 0;
-}
+int main() { std::cout << "A!"; return 0; }
```
danieleds: no need to change this, use the old code, it's immediate to recognize it's just an hello world.

Teklad:
There ya go, lol.  I suppose making a clean looking shell script can be a bit of a pain. :p

danieleds:
There is no such thing as a clean shell script :P

danieleds:
If that's ok for you I'm going to merge only the first few lines, as those seems to be the only changes related to this PR ("*Fixes having to use --qmake and --lrelease flags on systems that don't have qtchooser.*")

Some of the other changes are interesting (e.g. the --debug flag) but it's better to do that in another PR so that the history of the repo doesn't become a mess.

All the other stuff is better suited for a "refactor" PR.

Teklad:
Gimme one sec and I'll revert all and apply the top portion as a patch... then I'll open a separate PR for the debug flag (its useful)

Teklad:
This should be more to the point.  I'll add a debug flag in the next PR... then we can refactor at a later time. (Before webengine becomes master)

Teklad:
All done.

Issue #433, opened 2017-04-25
%directory% in run command not giving directory


qcerasmus:
**I'm submitting a
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/notepadqq/notepadqq/blob/master/CONTRIBUTING.md#question
```
When creating a "Run" command you can use "%directory%" as an argument.

**Current behavior**
%directory% does not get replaced with the directory of the current tab. It literally stays "%directory%".

**Expected behavior**
Change the "%directory%" of the command to the current tab directory.

**Minimal reproduction of the problem with instructions**
Add a new run command like "firefox %directory%". When you run this command, firefox will try to open: "http://www.%directory%.com/"

**Please tell us about your environment:**
<!-- Operating system, Qt version, Notepadqq version, ... -->

PR #434, opened 2017-04-25
%directory% in run command.

qcerasmus:
fixes https://github.com/notepadqq/notepadqq/issues/433.


JuBan1:
Hey,

I see a potential issue with files with a path like ".../some_file_dir/file". Since all instances of `fileName` are replaced with `""`, directories with `filename` in its name will be mangled. Also, you can get the file path without the `"file://"` by using `toLocalFile()`

Something like this should work: 

`QString directory = QFileInfo( url.toLocalFile() ).absolutePath();`

qcerasmus:
Updated to the safer way of doing it.

danieleds:
Thanks! There already was a pending PR (but the author didn't reply to the feedback): #398 
Can you follow those guidelines?

qcerasmus:
I've made the changes to use the keywords that were suggested.
The translation files will need to be updated now as well.

PR #443, opened 2017-05-10
add handel-c langage https://en.wikipedia.org/wiki/Handel-C

FAIG2014:
This pull request works with https://github.com/codemirror/CodeMirror/pull/4737

they are working together to add support of handel-c: https://en.wikipedia.org/wiki/Handel-C

### DIFF
--- a/src/editor/classes/Languages.js
+++ b/src/editor/classes/Languages.js
@@ -53,6 +53,13 @@ var Languages = new function() {
             fileExtensions: ["c", "h", "i", "xbm", "xpm"]
         },
 
+        "handel-c": {
+            name: "handel-C",
+            mode: "clike",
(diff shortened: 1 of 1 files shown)


JuBan1:
Sorry it took so long to get back to this. Merged.

danieleds:
Do this language actually work with the version of CodeMirror that we currently ship?

JuBan1:
Derp, yea, I thought it was just using the clike mode but it's actually looking for the handel-c "submode". We could add it to our CM fork, or revert the change.

danieleds:
Probably better if we revert for now.
We need a big change to switch to an upgraded version of CodeMirror and I'd like to do that as cleanly as possible.

Issue #452, opened 2017-05-23
src/editor/libs/codemirror/ is empty in downloadable source release


ericjs:
**I'm submitting a ...**  (check one with "x")
```
[ x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/notepadqq/notepadqq/blob/master/CONTRIBUTING.md#question
```

**Current behavior**
Building from source downloaded from https://github.com/notepadqq/notepadqq/archive/v1.0.1.tar.gz
fails because src/editor/libs/codemirror/ is empty.

**Expected behavior**
src/editor/libs/codemirror/ is not empty and the build does not fail

**Minimal reproduction of the problem with instructions**
Download from link above, extract, cd to the right directory and run
./configure --prefix /usr --qmake /usr/bin/qmake-qt5 --lrelease /usr/bin/lrelease-qt5
(successful)
the command fails with 
error: CodeMirror submodule not available at `src/editor/libs/codemirror'! Make sure you downloaded the source code using git as specified in README.md

Note there are no such instructions in the README.md. And surely this is not intended anyway with this downloadable source package?

**Please tell us about your environment:**
Chakra Linux, QT 5.8, Notepadqq 1.0.1

ericjs:
Also your Makefile-codemirror tries to
cd "$(DESTDIR)"/mode/m4/

I tried manually adding a few recent codemirror releases and they have no m4 directory in their mode directory, so this fails.

ericjs:
It case it wasn't clear, this issue was about the latest release src from https://github.com/notepadqq/notepadqq/releases not being buildable without manual intervention (and taking  some guesses). I made it work on my local machine, and would have submitted it to my distro's communuty repository, but not in that state (i.e. with my own hacks, without even any comment from the npqq team as to the correct version of codemirror to use).

JuBan1:
Sorry for not getting a reply for so long. The project is on a low burn for the time being.

Deployment isn't exactly my specialty but I'll try to clear things up as best I can:

> Note there are no such instructions in the README.md. And surely this is not intended anyway with this downloadable source package?

The instructions are to download the source from git. The README provides the following commands:

> $ git clone --recursive https://github.com/notepadqq/notepadqq.git
    $ cd notepadqq
    notepadqq$ ./configure --prefix /usr
    notepadqq$ make

[...]

PR #471, opened 2017-08-17
Allow configure to init CodeMirror if downloaded by source.

JuBan1:
Possible solution for #452. 

This here PR allows configure to clone our CodeMirror repo directly. Usually it would only do so through `git submodule init/update` which isn't available when Nqq was downloaded as a .zip archive. 

The CM branch being pulled is hardcoded (`d790fc3`) because I didn't find a way to pull that info off github. But we haven't changed it in a year and I doubt we'll do so in the foreseeable future. Maybe never if we get away from QWebKit and CM.

I don't feel comfortable making this change without a second pair of eyes though. My bash-fu isn't good.

### DIFF
--- a/configure
+++ b/configure
@@ -171,17 +171,30 @@ if [ $(echo $?) -eq 0 ] ; then
     done
 fi
 
-# If we're in a git repo and git is installed, update the submodules.
-# If ".git" is not found, assume that the complete source
-# code is already available.
+# Test if git is installed
 which git 2>/dev/null 1>/dev/null
-if [ -d ".git" ] && [ $(echo $?) -eq 0 ]; then
-    git submodule init
-    git submodule update
+if [ $(echo $?) -eq 0 ]; then
+
+    # If we're in a git repo, init submodules.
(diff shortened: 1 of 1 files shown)


danieleds:
I'm not much comfortable with hardcoding the CM branch. We *will* forget about it.

I think a better (but annoying) way is to just detect if the submodules are missing and abort the process with instructions on how properly download it using git, maybe offering to do that automatically.

We could also explore git subtrees: https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree

I'm not sure right now what impact this would have on our repo.

danieleds:
@JuBan1 I have included the squashed CM tag in the Notepadqq repo, as a git subtree.

The .zip from github now contains the whole source ;)

JuBan1:
With this we won't need the stuff in `configure` dealing with submodule initialization and checking for CM's presence, do we?

danieleds:
Yeah, we can simplify that too. I'll take care of it.

PR #475, opened 2017-09-04
Revert "Vector icon: translate text to path"

danieleds:
Reverts notepadqq/notepadqq#474

The font is different in the "path" version.

Here you can see the PNG version (left) compared to the old SVG (right):

nqq_orig

And here is the same PNG version (left) compared to the new SVG (right):

nqq_new

In addition, the SVGs in `support_files/icons/hicolor/scalable/apps` and `src/ui/icons/notepadqq/scalable` weren't updated. 
If I remember correctly, their hash is somewhat different from that of images/notepadqq.svg because images/notepadqq.svg is an "inkscape svg" and the others are "pure svg" or something like that.

@JuBan1 @darealshinji

### DIFF
--- a/images/notepadqq.svg
+++ b/images/notepadqq.svg
@@ -225,30 +225,18 @@
        x="28.000006"
        y="920.36218"
        ry="0" />
-    <g
-       id="g4232"
-       transform="translate(-0.6570115,134.84626)">
-      <g
-         transform="translate(-3.1819805,-2.8284271)"
-         id="text4303-0-9-2"
-         style="font-style:normal;font-variant:normal;font-weight:normal;font-stretch:normal;font-size:40px;line-height:125%;font-family:sans-serif;-inkscape-font-specification:sans-serif;letter-spacing:0px;word-spacing:0px;fill:#e0e0e0;fill-opacity:1;stroke:none;stroke-width:1px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1">
-        <path
-           inkscape:connector-curvature="0"
-           id="path4226"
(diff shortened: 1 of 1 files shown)


darealshinji:
Did you compare them in Inkscape?
This is what the current icon looks like rendered by Github:
icon

I was using this tutorial: https://www.youtube.com/watch?v=_01Jdvr7eXI

danieleds:
I compared them with a desktop application, and on GitHub I was seeing them exactly like in the desktop application.

I suspect your're missing some font on your system, that's why it falls back to that similar but different font.

I compared with the PNG to make sure that *I* wasn't missing any font ;)

danieleds:
P.S. A path conversion would be great exactly for this reason. We just need the right font.

danieleds:
By the way, it's probably Verdana

JuBan1:
You're right- the PNG looks different. The two SVGs are the same on my system as well which is what I compared to look for differences.

darealshinji:
The left font seems to be Verdana. The right one isn't. @danieleds can you try to figure out which font it is? Or, if the font is displayed the intended way on your computer, can you do the path conversion yourself?

30027427-3fe48f78-9181-11e7-984b-5152117ce296

danieleds:
The left one is how it is supposed to look: the font is Verdana.

I'll do the conversion myself later in the day 👍

darealshinji:
Here's a changed SVG using Verdana:
[notepadqq.svg](https://github.com/notepadqq/notepadqq/files/1277615/notepadqq.svg.txt)
[notepadqq.svg.diff](https://github.com/notepadqq/notepadqq/files/1277614/notepadqq.svg.diff.txt)

danieleds:
@darealshinji how did you manage, in the old PR, to preserve the original text together the path?

danieleds:
oh ok, you just moved it, nevermind

PR #477, opened 2017-09-05
Convert text to path in svg logo

danieleds:
@darealshinji @JuBan1 can you check that you're seeing them correctly?


danieleds:
fyi, I have reorganized a little bit the stuff in the svg, by adding layers. The original text is stored in a hidden layer.

JuBan1:
Yes, this version looks identical to the .PNGs to me.

darealshinji:
Looks good!

PR #480, opened 2017-09-09
Added a "Retry as Root" to Nqq's save dialog.

JuBan1:
It's mildly infuriating to not be able to save changes to config files using Nqq (unless you start it with sudo) so here's a patch that adds a "Retry as Root" button to the dialog that pops up when a file could not be saved.

The patch queries the system for the existence of either `kdesu` or `gksu` and enables a few bits of code at compile time. If neither is found the dialog looks like it used to. 

Qmake provides a very nice facility to do the querying so I put it into the `ui.pro` file. We can move it to `configure` if you think that isn't a good place.

The new dialog:
image

The password prompt with `kdesu`:
image

The prompt with `gksu`:
image


danieleds:
I like it! But why a compile-time check instead of a runtime check?

JuBan1:
You're right- that's not needed. I did that for when I was tinkering with polkit (hence the name of the branch). After changing to using a shell command I didn't think about it again.

Changes are made. It now queries the return values of `which kdesu` and `which gksu` once at run-time instead.

src/ui/docengine.cpp
```diff
@@ -5,8 +5,11 @@
 #include <QTextCodec>
 #include <QTextStream>
 #include <QCoreApplication>
+#include <QPushButton>
```
danieleds: ?

src/ui/docengine.cpp
```diff
@@ -5,8 +5,11 @@
...
+
 #include "include/mainwindow.h"
 #include "include/nqqsettings.h"
+#include "include/Sessions/persistentcache.h"
```
danieleds: is this used?

src/ui/docengine.cpp
```diff
@@ -389,6 +392,58 @@ void DocEngine::unmonitorDocument(const QString &fileName)
...
+    QProcess p;
+
+    if (sudoProgram == "kdesu")
+        p.start("kdesu", QStringList()
+                << "--noignorebutton"
+                << "-n"
+                << "-c" << "cp" << filePath << outFileName.toLocalFile());
+    else if (sudoProgram == "gksu")
+        p.start("gksu", QStringList()
+                << "-S" << tr("-m Notepadqq asks permission to overwrite the following file:\n\n%1")
```
danieleds: -m should be outside of tr().

Also, it's better if you quote the string, so that translation won't mess up with it. For example, you could use `'` as the quoting character and then replace any occurrence of `'` in the string with `'"'"'`

danieleds: or well, instead of manually quoting, just add a separate entry in the QStringList

src/ui/docengine.cpp
JuBan1: Necessary because `msgBox.addButton(..)` returns a `QPushButton*`. Wasn't needed with the previous `setStandardButtons(..)` but using custom buttons means using `QPushButtons`/`QAbstractButtons` directly.

src/ui/docengine.cpp
JuBan1: Yes, to get a temporary path to store the file to:

```c++
QString filePath = PersistentCache::createValidCacheName(
                PersistentCache::cacheDirPath(),
                outFileName.fileName() )
            .toLocalFile();
```

Returns something like `/home/user/.config/Notepaddqq/tabCache/filename.12345`. Same method as sessions use.

danieleds: Oh ok I see it now, but still I'm not sure why you store it in the cache instead of in the system tmp dir?

danieleds: Also, does the file get deleted after a successful copy? Or we still need it?

src/ui/docengine.cpp
JuBan1: Changed the `-m` part. 

I don't think quoting the string is needed. Pretty sure that `Process::start()` automatically quotes arguments in the string list that contains whitespace because I initially had the whole string quoted manually and the quotes would actually show in the GUI.

src/ui/docengine.cpp
JuBan1: I just chose the cache because the code for it already exists and works fine. I can use /tmp/ instead if you want but I don't see much of a reason.

The file gets deleted after the save attempt regardless of whether it was successful or not since it has either done its job or the attempt was cancelled. That happens at the end of `trySudoSave`.

danieleds: Perfect, sorry I'm in a quite distracting environment right now :/ you do the squashmerge?

JuBan1: No worries. I'm glad you can find some time here and there at all. Slow progress is better than no progress ;)

PR #481, opened 2017-09-11
Add:Chinese language support

realasking:
I created an translation file for Chinese language of China. Could you merge it to notepadqq? Btw, I like your application very much.


JuBan1:
Thanks!

Issue #248, opened 2016-08-28
Opening a recent document does not check for the file's existence


JuBan1:
You can open a recent file that does not exist anymore.

Steps to reproduce:
- Create any new file and open it in Notepadqq.
- Close Notepadqq and delete the file.
- Open it again and select the deleted file from the list of recent documents.
- The file is opened as if it still exists, no indication is given that it's gone.
- Even worse, when pressing 'Save' the file will be stealthily recreated on disk.
##

danieleds:
True, we should at least give a warning.
I mean, I see nothing wrong with this behavior*, but we should inform the user that the file is actually gone, e.g. with a dialog that pops out right after he clicks on the recent filename from the list.


*it mimics the command line behavior so opening a unexisting file it's fine per se

JuBan1:
It makes sense for command line, but opening a *recent document* when it doesn't exist could be deceiving. I think most programs remove non-existing files from the recent list. That could be an option too.

danieleds:
Agreed

Teklad:
Another method would be to gray it out as a visual indicator.  Its a bit
cleaner than a pop out box. Or just removing works too. I'm fine with
either.

danieleds:
@JuBan1 @Teklad *how* and *when* would you detect that the file is not there anymore, so that it can be removed from the list? 
(I like this approach more than the dialog, btw)

JuBan1:
When thinking about it, the only surefire way would be to check it every time the Menu is opened (And even then a file could be deleted in the time between opening the menu and clicking an entry- but that's an obscure case.) or to create file monitors for all files on the recent list in order to be notified if they are deleted. That's probably unreasonably much effort for a feature like this.

N++ actually just shows you a dialog and removes it from the recent list if No is chosen.

![](http://pix.toile-libre.org/upload/original/1472504121.png)

Maybe a compromise: 

* Check all recent files for existence on program start-up and remove nonexistent files from the list. 

* Then, when the user actually selects an entry from the recent list, do another check for existence in case the file was deleted between start-up and this point in time. If it was, show a dialog like N++ does.

danieleds:
> When thinking about it, the only surefire way would be to check it every time the Menu is opened (And even then a file could be deleted in the time between opening the menu and clicking an entry- but that's an obscure case.) or to create file monitors for all files on the recent list in order to be notified if they are deleted. That's probably unreasonably much effort for a feature like this.

[...]

PR #486, opened 2017-09-12
Prompt user when opening a file from recent history that does not exist anymore

JuBan1:
We talked about that looong ago in #248.

This implements it like in N++:

* When clicking on an entry in recent history, the file existence is checked.
* * If it exists, it'll be loaded as usual.
* * If it does not exist, the user is asked whether he wants to re-create it.
* * * If user answers yes, the file is tentatively opened (this is what always happens at the moment).
* * * If user answers no, the file entry is removed from recent history.

The "Open all recent files" action works with this as well.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -1753,8 +1753,8 @@ void MainWindow::updateRecentDocsInMenu()
     for (QVariant recentDoc : recentDocs) {
         QUrl url = recentDoc.toUrl();
         QAction *action = new QAction(Notepadqq::fileNameFromUrl(url), this);
-        connect(action, &QAction::triggered, this, [=]() {
-            m_docEngine->loadDocument(url, m_topEditorContainer->currentTabWidget());
+        connect(action, &QAction::triggered, this, [this, url]() {
+            openRecentFileEntry(url);
         });
 
         actions.append(action);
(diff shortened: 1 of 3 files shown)


src/ui/mainwindow.cpp
```diff
@@ -1841,14 +1841,8 @@ void MainWindow::on_actionEmpty_Recent_Files_List_triggered()
...
-    QList<QVariant> recentDocs = m_settings.General.getRecentDocuments();
-
-    QList<QUrl> convertedList;
-    for (QVariant doc : recentDocs) {
-        convertedList.append(doc.toUrl());
-    }
-
-    m_docEngine->loadDocuments(convertedList, m_topEditorContainer->currentTabWidget());
+    for (const auto& doc : m_settings.General.getRecentDocuments())
+        openRecentFileEntry(doc.toUrl());
```
danieleds: The call to `m_docEngine->loadDocuments` was useful because, in case of problems (like missing permissions for a lot of files) it offers to abort the whole operation.

Could you instead warn the user in advance if *any* of the files doesn't exist, then offer to create them anyway (all or none)? After that, you would call `m_docEngine->loadDocuments` for all the files or for only those that already existed, depending on the user choice.

I don't know if it's clear what I'm saying

JuBan1: So if I understand you correctly adding a 'Yes to All' and 'No to All' option to the dialog would do what you want?

However, there's no permissions checking going on at the moment. Nqq doesn't actually create the file on the HDD until the user saves the document manually. It just opens it up as a tab and marks it as dirty. That is the same as it works right now in master. I haven't changed that part of the functionality; I added only the extra check for file existence.

danieleds: Take the master branch and follow these steps:

```bash
cd /tmp
echo a > a.txt; echo b > b.txt; echo c > c.txt
```
Now open a.txt, b.txt and c.txt in Notepadqq in order to put them in the recents list. Then close all the tabs and close Notepadqq. Now run:
```bash
chmod -r a.txt b.txt c.txt
```
and then open Notepadqq again. Choose "Open All Recent Files". You get a dialog complaining about not being able to open a.txt. You now have the option to Abort, which will abort the whole operation and will not even try to open b.txt and c.txt.

The Abort functionality is lost in this branch, because `loadDocuments` is called multiple times instead of one for all.

What I suggested was

 1. First check which files do exist, and which doesn't. Collect a list of the files which doesn't.
 2. Ask the user if he wants to open all the missing files anyway (just one dialog with a Yes/No reply).
 3. If user chose Yes, then pass the whole "recents" list to `loadDocuments`. Otherwise, pass the "recents" list minus the list of the missing files.

JuBan1: I see, yes that's sensible. I implemented it that way now.

Issue #411, opened 2017-03-27
Cancel auto indent


sonichy:
When press enter to start a new line, the line will auto indent, mostly I do not need this, and if I want to indent, I can press tab.

JuBan1:
You can turn that off: `Edit -> Identation -> Enable Smart Indent`

Teklad:
%s/Identation/Indentation/

sonichy:
The menu [Edit -> Identation -> Enable Smart Indent] does not exist !

danieleds:
What version of Notepadqq do you have?

JuBan1:
Closing this due to inactivity (and because it was most likely just a version issue).

PR #492, opened 2017-10-13
Fixes #488

### LINKED ISSUE #488: Please remember "Enable Smart Indent" setting
I was fed up with "Smart Indent" when write HTML.
When I disable "Smart Indent", the trouble gone.
But the option can not save, when restart notepadqq, the option enable again!

qcerasmus:
Added a setting to remember if a user wants smart indentation.

### DIFF
--- a/src/ui/include/nqqsettings.h
+++ b/src/ui/include/nqqsettings.h
@@ -81,6 +81,7 @@ public:
         NQQ_SETTING(RecentDocuments,                QList<QVariant>, QList<QVariant>())
 
         NQQ_SETTING(NotepadqqVersion,               QString,    QString())
+        NQQ_SETTING(SetSmartIndentation,            bool,       true)
     END_CATEGORY(General)
 
     BEGIN_CATEGORY(Appearance)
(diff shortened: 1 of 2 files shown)


qcerasmus:
This should also clear up #411 .

JuBan1:
I guess it's a good idea to remember the setting since it does no harm for users who want to disable it only temporarily.

src/ui/include/nqqsettings.h
```diff
@@ -81,6 +81,7 @@ class NqqSettings {
         NQQ_SETTING(RecentDocuments,                QList<QVariant>, QList<QVariant>())
 
         NQQ_SETTING(NotepadqqVersion,               QString,    QString())
+        NQQ_SETTING(SetSmartIndentation,            bool,       true)
```
danieleds: @qcerasmus @JuBan1  wouldn't "SmartIndentation" be better instead of "SetSmartIndentation"?

JuBan1: That came to my mind as well but after looking at the other settings it turns out we actually have a lot of those "verb-style" names. "RememberTabsOnExit", "ShowSpaces", "OverrideFontFamily", "CheckVersionAtStartUp". So it's actually not out of the ordinary.

danieleds: true, however this particular case creates weird/confusing things down in the code:

`m_settings.General.getSetSmartIndentation()`, `m_settings.General.setSetSmartIndentation()`

I think it's better to change it now before it goes into the users settings file, I can do it

PR #494, opened 2017-10-16
Use correct URL

JoniJnm:


danieleds:
Thanks!
Actually, could you make it point to the following address instead?

https://github.com/notepadqq/notepadqq/graphs/contributors

JoniJnm:
done

PR #500, opened 2017-10-30
Fixed always true condition

lunixoid:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. Warnings, found using PVS-Studio:
notepadqq/src/ui/Search/searchinfilesworker.cpp     119     warn    V547 Expression 'hasResult' is always true.


danieleds:
Thanks!

PR #504, opened 2017-11-09
Made use of .gitattributes to clarify linguist percentages

JayTurnr:


JayTurnr:
A lot better now! Doesn't claim your project is 33% HTML!

PR #506, opened 2017-11-09
Revert "Making use of GitHub Markdown Task Lists"

danieleds:
Reverts notepadqq/notepadqq#503


JayTurnr:
Was there something wrong with it?

danieleds:
I found out that github UI was implying something like "1/3 cheched => 33% completed".
Not quite our case :)

JayTurnr:
Oh, oops! My bad! Strange it says it reverts #504 yet it doesn't.

danieleds:
Yep, I don't know what happened

PR #525, opened 2018-01-10
Ability to disable saving search history and updated README

robertapengelly:
Added option to disable saving find and replace history and updated README with extra step required to build on Ubuntu.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Build
 
 #### Build it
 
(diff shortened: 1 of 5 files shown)


JuBan1:
Hey, sorry for the long wait.
I think that's a feature worth having! I went ahead and modified it a tad. Hope it'll still fit your needs this way.

I also moved the change to README to its own commit to keep things separated but it's also a good addition.

Much appreciated!

robertapengelly:
Yeah the changes are perfect it's what I originally intended.

PR #551, opened 2018-02-15
Update of Hungarian translation

SandorPetrovics:


JuBan1:
You're doing it right with making Pull Requests. Only danieleds and myself have push access to the main repository, so all changes go through either one of us.

PR #562, opened 2018-03-02
Update README

### LINKED ISSUE #522: Add screenshot to README.md
**I'm submitting a feature request**
Hi, I think it's a good idea if you add one or more screenshot of the program to README.md

JuBan1:
Now includes a short description and advertises some features. Also includes images. I took them from [Notepadqq.com](https://notepadqq.com/s/) since they're still relatively up-to-date for what they show.

Since the readme grows in size I included a small table of contents as well.

@danieleds could you give this a quick glance if its to your satisfaction?

(Fixes #522

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,39 @@
-Notepadqq
-=========
+# <img src="https://user-images.githubusercontent.com/4319621/36906314-e3f99680-1e35-11e8-90fd-f959c9641f36.png" alt="Notepadqq" width="36" height="36" /> Notepadqq [![Build Status](https://travis-ci.org/notepadqq/notepadqq.svg?branch=master)](https://travis-ci.org/notepadqq/notepadqq) [![Snap Status](https://build.snapcraft.io/badge/notepadqq/notepadqq.svg)](https://build.snapcraft.io/user/notepadqq/notepadqq)
 
-[![Build Status](https://travis-ci.org/notepadqq/notepadqq.svg?branch=master)](https://travis-ci.org/notepadqq/notepadqq)
-[![Snap Status](https://build.snapcraft.io/badge/notepadqq/notepadqq.svg)](https://build.snapcraft.io/user/notepadqq/notepadqq)
+### Index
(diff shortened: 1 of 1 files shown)


JuBan1:
FYI: You can see the changes directly [on my Nqq fork](https://github.com/JuBan1/notepadqq).

danieleds:
Some quick notes:

  1. the new header is awesome
  2. in "what is it", I'd keep only the first line of description and the screenshot; I'd remove the whole sections "How do you like it?" and "Speed is not a compromise". I don't want to overload too much the readme when there already is a full website for that. The first screenshot is really important though, keep it.
 3. with the "★" symbol in the packages section I was trying to mark the officially maintained packages, which would be those that we're sure we can always update (basically AppImage, Snap, Ubuntu and Debian). We're not really in control of the other packages. Now I realize the meaning of "★" is not clear at all

danieleds:
(I've also just bought the notepadqq.com domain ;-) )

danieleds:
If you want to add additional screenshots other than that one, a link to a wiki page would be great

JuBan1:
Yep, I thought the stars were just visual aesthetics :)

How's [this](https://github.com/JuBan1/notepadqq/blob/master/README.md) then? The Wiki mainpage (which is currently as good as empty) could perhaps contain the (now deleted) "How do you like it?" and "Speed is not a compromise" sections.

danieleds:
I liked the table of contents like it was before; instead of the link to the wiki in the ToC, you could just add a link under the screenshot saying something like "Click here for more screenshots".

As for the link to the website; I don't think we need it in the ToC; it's already at the top of the GitHub page.

> Yep, I thought the stars were just visual aesthetics :)

Okay, so right now they're useless :) I propose to remove them completely

JuBan1:
Okay, last revision: https://github.com/JuBan1/notepadqq/blob/master/README.md

danieleds:
Perfect!

PR #566, opened 2018-03-05
Add Appstream metainfo

JuBan1:
Add a metadata file for AppStream and install it to /usr/share/metainfo on `make install`.

As far as I can gather, this is all we've got to do. And the XML file includes all necessary info.


danieleds:
Did you forget the `$$INSTALL_ROOT$$PREFIX` ?

Other than that, looks good to me

JuBan1:
Ah, that was actually something I wanted to mention but it slipped my mind: as far as I understand AppStream, the file needs to land in `/usr/share/metainfo/`, not anywhere else. That means prefixes other than `/usr` won't have the intended result. Should we do anything about that?

PR #578, opened 2018-03-15
Prevent appmenu-qt5 from removing our menubar

JuBan1:
I searched for a better alternative but I've found none. It seems the best thing to do is to prevent Unity from turning Nqq's menubar into a global bar entirely. It messes with the overall look-n-feel of the platform and introduces a small bug (the window jumps upwards a little everytime it's reopened- likely a side-effect of unsetting `QT_QPA_PLATFORMTHEME`).

This isn't an issue with newer Ubuntu versions since they've ditched Unity and appmenu-qt5 in favor of GNOME which shows our menu bar as it should.


danieleds:
Looks good

Issue #577, opened 2018-03-15
Error saving session


vertigo220:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting (done)
```

**Current behavior**

I'm not even sure how it happened. I had multiple instances of Nqq open, and the only thing I can think of is that some were open as root (to edit files owned by root) and some were open as me, the user. But when I tried to close a Nqq window, I got the error "Error trying to write to "/home/keith/.config/Notepadqq/tabCache/monitrc.1006151010 -- Permission denied." Considering this is a directory under my (user) home directory, I wouldn't expect myself or root to have any problem writing to it. I did the "Retry as root" but that didn't work. I went to the directory and (IIRC) couldn't write anything in it except as root. I did ls -la and this was the result:

total 20
drwxr-xr-x 2 root  root  4096 Mar  7 12:04 .
drwxrwxr-x 3 keith keith 4096 Mar  6 15:24 ..
-rw-r--r-- 1 root  root   970 Mar  6 15:24 monitrc.794864509
-rw-r--r-- 1 root  root   727 Mar  6 15:24 system-status.1302088411
-rw-r--r-- 1 root  root     8 Mar  7 12:04 test.test

I then tried "sudo chown keith. *" but that didn't seem to change anything.

**Expected behavior**

I'm not really sure if this is a bug or just an unfortunate case of getting things all twisted up with permissions, but it seems that a) it shouldn't be possible to get into that position and b) "Retry as root" should have worked. In any event, I was unable to save my session. Not exactly a huge deal, but irritating, and if I'd had a bunch of files open it could have been problematic.

**Minimal reproduction of the problem with instructions**

As mentioned, I'm not even sure how I got to that point. I just know I had around 4 different Nqq windows open, some ran from the terminal, some (I think) run by launching from the panel and/or opening a file, some as root, some as user/me. I'll admit that being new to Linux I don't fully understand the intricacies of permissions and the limitations imposed by them. I just know that in many, many years of using N++ in Windows I've _never_ had a problem, and in just a few weeks of use in Linux this happened.

**Please tell us about your environment:**

Mint 18.3
Qt 5.5.1
Nqq 1.2.0

JuBan1:
I'm impressed, to be honest. I didn't even know you could run multiple instances of Nqq that way! But it turns out you can.

[...]

PR #580, opened 2018-03-15
Implement --allow-root and prohibit running as root by default.

JuBan1:


danieleds:
Before merging, I want to see if it works/compiles on macOS

JuBan1:
Sure, give it a go.

It should work since `unistd.h` is part of POSIX which macOS also implements and a uid of 0 universally indicates root as far as I know.

danieleds:
Perfect!

Nadav-Eyesight:
So I just started using Notepadqq, and the one thing that bothers me is having to use `-allow-root`. Why was this added? Is there anything stopping me from simply aliasing `notepadqq --allow-root`?

danieleds:
In a few words: because you probably don't really want to do it.

Read this discussion, which led to this change: https://github.com/notepadqq/notepadqq/issues/577

Nadav-Eyesight:
Thanks for the quick reply. I've made a feature request.

PR #586, opened 2018-03-28
Fix document printing.

JuBan1:
Turns out printing is rather easy to get into a reasonable state.

Doesn't use specific color themes to print. "Print Now" doesn't work for some reason so it stays disabled.

### DIFF
--- a/src/editor/styles/app-print.css
+++ b/src/editor/styles/app-print.css
@@ -9,3 +9,20 @@
 .editor.print .CodeMirror-activeline-background {
     background: none;
 }
+
+.editor.print .CodeMirror-linenumber {
+   visibility: hidden;
+}
+
+.editor.print .CodeMirror-gutters {
+    visibility: hidden;
+    width: 0px;
+}
+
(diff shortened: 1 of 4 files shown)


danieleds:
Nice!
For some reason, last time I tried, the lines at the bottom of the pages were cut vertically in half.

I see some issue with `m_webView->setStyleSheet()` though... I don't think it's supported on WebEngine. We're going to need to do that from Javascript. It would be cleaner anyway.

JuBan1:
I remember that too. Either I got lucky or the Qt dev fixed something about printing on their end. 

Perhaps WebEngine won't need `setStyleSheet`. It seems to me like a bug: less than at least one full page paints the *empty* area of the page in the QObject's background color. But that doesn't happen when printing more than one page. 

I don't know if that can be fixed through Javascript because it's not the HTML's background color that's at fault here.

danieleds:
Ah, got it, that makes sense. At first glance I thought it was css-related.

I'm testing it on WebEngine and it crashes, but at least all the methods seem to be supported.

PR #589, opened 2018-03-31
fix: Prefer GTK's icon theme if XDG_CURRENT_DESKTOP is not set

### LINKED ISSUE #552: Missing icon messages when launching from terminal
```
[ x] bug report => search github for a similar issue or PR before submitting
```

**Current behavior**

notepadqq

**Expected behavior**

"Missing icon..." lines shouldn't be present

See above

Mint 18.3 Cinnamon, Qt 5.5.1, Nqq 1.2.0, GNOME terminal 3.18.3

Teklad:
This fixes #552

Since quite a large margin of desktop environments tend to use GTK's icon theme preferences, this will probably be a better default.  Should fix any further complaints about  missing icons as well.  I use Openbox and it has always happened to me but I never even noticed.  

I tested the fix while running XFCE4 and i3 as well.  XFCE was just to make sure it wasn't interfering with desktops that do set that environment variable.

### DIFF
--- a/support_files/launch/notepadqq
+++ b/support_files/launch/notepadqq
@@ -33,6 +33,13 @@ if [ "$QT_QPA_PLATFORMTHEME" = "appmenu-qt5" ]; then
     export QT_QPA_PLATFORMTHEME=""
 fi
 
+# For less common desktop environments, this doesn't get set at all which will cause QT to default to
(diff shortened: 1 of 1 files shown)


JuBan1:
Nice! Seems to work on my end.

PR #595, opened 2018-04-05
add spec file for build rpm

andykimpe:
add spec file for build rpm

+ instrustion en readme

+ depot copr for fedora

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,27 @@ If you encounter errors make sure to have the necessary libraries installed. For
 
     notepadqq$ sudo apt-get install qt5-default qttools5-dev-tools libqt5webkit5 libqt5webkit5-dev libqt5webkit5-qmlwebkitplugin libqt5svg5 libqt5svg5-dev
 
-For CentOS:
+For Fedora(if <20 replace yum by dnf), CentOS RedHat(if <6 replace yum by dnf):
 
-    notepadqq$ sudo yum install -y qt5-qtbase-devel qt5-qttools-devel qt5-qtwebkit-devel qt5-qtsvg-devel
+    notepadqq$ sudo yum install -y qt5 qt5-devel qt5-qtbase-devel qt5-qttools-devel qt5-qtwebkit-devel qt5-qtsvg-devel
+
+For build rpm
+
+for Centos Readhat require epel-release
+
+    notepadqq$ sudo yum install -y epel-release
+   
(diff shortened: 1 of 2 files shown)


danieleds:
I don't know about the rpm build procedure, but the .spec file should probably reside in the notepadqq-packaging repo?

danieleds:
In addition, the changes to the prompt are a bit confusing: `notepadqq$ sudo whatever` should be reverted to `sudo whatever`. The "notepadqq$" prefix is used as an indication of the working directory for then the user is building nqq by source: there is no need for it when installing from a package.

Similarly, package build instructions do not belong here: you should move them to the notepadqq-packaging repo.

I believe the "-y" flag to the apt-get command is bad practice: we're asking a human to run these commands, so it is better if he has the chance to review the changes to their system.

About the "Fedora (Only Stable Version actual 03/2018 version 26 and 27)" paragraph: it's not really clear when it can be used. Also, does it have to be that complicated?

Basically, what I'd keep is the suggestion to replace yum by dnf. Actually, I'd just replace it inside the command and assume the user is using the latest version.

JuBan1:
> `wget https://github.com/notepadqq/notepadqq/archive/v1.3.1.tar.gz - O notepadqq-1.3.1.tar.gz`

Any reason not to just use `git clone https://github.com/notepadqq/notepadqq.git` instead?

We should keep mentioning specific versions (like 1.3.1) to a minimum because I guarantee we'll forget to update some of them when we change the version number.

jdeluyck:
I'd like to comment that I'm working on getting Notepadqq included with fedora. Please see the notepadqq-packaging repo.

scx:
> I'd like to comment that I'm working on getting Notepadqq included with fedora. Please see the notepadqq-packaging repo.

I have already provided packages for EPEL7 and Fedora 26-29.
https://github.com/notepadqq/notepadqq/issues/668#issuecomment-401185470

It would be nice if you could use my patches for EPEL7 in your RPM SPEC.

jdeluyck:
My package has gotten accepted, and I've just gotten sponsored into the packager group.  I'll have a look at your patches this weekend :)

scx:
See also:
https://github.com/notepadqq/notepadqq/issues/901

danieleds:
Closing since this has been stale for a long time. Feel free to open another PR in case of new contributions. Thank you

PR #598, opened 2018-04-07
Refactored DocEngine::loadDocuments.

JuBan1:
Here's what's done:

* Removed all the different loading functions from DocEngine (`loadDocument`, `loadDocuments`, `loadDocumentSilent`, `reloadDocument`, etc)
* Added a single new `loadDocuments(DocEngine::DocumentLoader&)` function.
* `DocumentLoader` is a simple wrapper class that contains all the possible arguments important for loading documents. This makes it much clearer what function invokation does what sort of work.

I was inspired to do this refactoring when I had to add yet another parameter to `loadDocuments` to handle:

* ~~Adding option to warn about loading large text files. (Can be set in Preferences)~~

An example of what this means for document loading. Instead of:

    m_docEngine->reloadDocument(tabWidget, tab, codec, bom); // What does this even do?!

It's now:

    m_docEngine->getDocumentLoader()
            .setUrl(editor->filePath())
            .setTabWidget(tabWidget)
            .setTextCodec(codec) // All those are optional
            .setBOM(bom)
            .setIsReload(true)
            .execute(); // Run the operation with the given params

I think this is a change for the better, especially if `loadDocuments` becomes even more complex in the future. But since it's a pretty significant change and might also affect the WebEngine branch I want to get your okay first, @danieleds.

### DIFF
--- a/src/ui/Sessions/sessions.cpp
+++ b/src/ui/Sessions/sessions.cpp
@@ -374,7 +374,11 @@ void loadSession(DocEngine* docEngine, TopEditorContainer* editorContainer, QStr
             if (!fileExists && !cacheFileExists)
                 continue;
 
-            const bool success = docEngine->loadDocumentSilent(loadUrl, tabW);
+            const bool success = docEngine->getDocumentLoader()
+                                 .setUrl(loadUrl)
+                                 .setTabWidget(tabW)
+                                 .setRememberLastDir(false)
+                                 .execute();
 
             if (!success)
                 continue;
--- a/src/ui/docengine.cpp
+++ b/src/ui/docengine.cpp
@@ -94,175 +94,146 @@ bool DocEngine::read(QFile *file, Editor* editor, QTextCodec *codec, bool bom)
     return true;
 }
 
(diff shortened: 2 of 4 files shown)


danieleds:
So nice!
I'd like to test it also on the webengine branch but could you first split the PR in more self-contained parts? I see there are changes about preferences that don't seem to belong here

JuBan1:
Yea, this shouldn't really have been there. 

The DocEngine refac is now self-contained in this PR. I'll submit the other PR once this is in.

danieleds:
There seem to be a conflict due to the new command line options.

Anyway, the conflicts with the WebEngine branch should be minimal and not really a problem.

Before merging, as this is a pretty delicate change, I want to be sure there are no bugs, and in particular I'd like to test every function that this change affects (reloading from the menu, reloading from the banner, etc). The old code was pretty messy.

JuBan1:
Resolved.

I tested the affected functionality on my side and didn't notice any issues. The changes done to the load function itself are mostly cosmetic to tidy things up a bit.

PR #600, opened 2018-04-08
Fix CodeMirror startup errors (Applies to WebEngine mostly)

Teklad:
Finally bothered to look into this and turns out it was a CodeMirror addon we forgot to load up front. This little change fixes the startup errors regarding defineSimpleMode missing as a function.

Decided to push this to the master branch because its something we overlooked in both the webengine branch as well as master.


danieleds:
Great! :)

JuBan1:
This should fix #378 then, shouldn't it?

Teklad:
Yea... it fixes that issue... should be able to close it... the other half of that issue is related to QT itself.

PR #604, opened 2018-04-08
feat: Add '--line' and '--column' commandline arguments.

JuBan1:
Adds `--line` and `--column` (and `-l` and `-c` in short) to open the file at the specified line and/or column. 

Does nothing when opening several files at once since `QCommandLineParser` doesn't allow arguments chains for specifying `-l` and `-c` for each positional argument.

I initially thought about appending optional parameters to the file path (`notepadqq path/to/file:10:1` for line 10, column 1) but `:` is a valid character in ext & friends and the extra processing to handle all cases would just set us up for headaches later down the road.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -34,6 +34,7 @@
 #include <QtPrintSupport/QPrintPreviewDialog>
 #include <QDesktopServices>
 #include <QJsonArray>
+#include <QTimer>
 
 QList<MainWindow*> MainWindow::m_instances = QList<MainWindow*>();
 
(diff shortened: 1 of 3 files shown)


src/ui/mainwindow.cpp
```diff
@@ -615,19 +616,57 @@ void MainWindow::openCommandLineProvidedUrls(const QString &workingDirectory, co
...
+
+    // This needs to sit inside a timer because CodeMirror apparently chokes on receiving a setCursorPosition()
+    // right after construction of the Editor.
+    Editor* ed = tabW->currentEditor();
+    QTimer* t = new QTimer();
+    connect(t, &QTimer::timeout, [t, l, c, ed](){
+        ed->setCursorPosition(l-1, c-1);
+        t->deleteLater();
+    });
+    t->start(0);
```
danieleds: I think you can do something like QTimer::singleShot(0, ...) which is cleaner

JuBan1: The needed overloads for `singleShot` have only been introduced in Qt 5.4. Since we're still stipulating a minimum of 5.3 I ended up having to use a timer object.

PR #606, opened 2018-04-08
Added custom indentation settings to sessions.

### LINKED ISSUE #602: can't save indentation settings for plain-text/literature writing
**I'm submitting a
[x] bug report => search github for a similar issue or PR before submitting
[x] feature request

**Current behavior**
If you set indentation setting to spaces and quit, after restarting, they're set to tabs, and can't be saved as spaces for next time.

**Expected behavior**
You should be able to save the indentation settings to be the same after quitting and starting.  When people edit plain-text/literature notes, papers, books, they probably don't want that set to tabs, at least for .TXT files that aren't code.  Notepadqq is now my favourite text editor (since my DE's main one has broken tab bars) but I can't actually set it to work more like an actual text editor that's a notepad, rather than an IDE, but mostly I edit .TXT files of notes, articles/papers, books, so would really like to be able to.  I'm a computer programmer/scientist too, so of course will eventually use tab indentation, but I'd like to set that of my own free will, or have it work for programming files but not .TXT, not just every session (plus getting warnings for two or three of my files I can only use the mouse to ignore.. but I love command-lines & keyboard shortcuts.)

'Edit -> Indentation -> Custom' should be saveable as permanent configuration between sessions.

Slackware GNU/Linux 14.2 (64-bit, up-to-date stable) w/KDE4... eventually will probably use with KDE3.5/TDE/Qt3 and/or classic *BSD Unix
Qt4.8.7, Qt5.7.1
notepadqq-1.2.0-x86_64-1_slonly (unofficial repository package compiled from un-/quasi-official build script repository) (first installed earlier version myself last year, then used SlackBuilds.org (SBo) build script, then used installer that SlackOnly.com updated SBo)

Teklad:
This will restore custom indentation modes between sessions when set via edit->IndentationMode->Custom...

Very small contained changes here.  This at least partially resolves #602

### DIFF
--- a/src/ui/Sessions/sessions.cpp
+++ b/src/ui/Sessions/sessions.cpp
@@ -47,6 +47,9 @@ struct TabData {
     bool active = false;
     QString language;
     qint64 lastModified = 0;
+    bool customIndent = false;
+    bool useTabs = false;
+    int tabSize = 0; 
 };
 
 struct ViewData {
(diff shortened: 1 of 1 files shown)


Teklad:
@JuBan1 you're more familiar with the sessions code than anyone else.  Do the changes here seem fine to you?

Issue #301, opened 2016-10-27
[Discussion] Hidden Preferences


danieleds:
Some of the newest source code editors have, as the only way to change settings, a json file which users can manually alter (see e.g. vscode). These programs do not contain a "settings" window.

I believe that an user interface is fundamental, especially for things like fonts or themes where we can display a real time preview of the result.

At the same time, maintaining a preferences window is time consuming and can probabily get messy when the number of settings increases (see Eclipse or similar). Our user base is composed by technical people, so it's not crazy to ask them to set their preferences in a text file.

### The idea
I was thinking that a good compromise could be in the middle: we could keep a preferences window, just like the one we already have. This window will provide access to the most important settings (I'd keep all the current settings).
For a lot of other less common settings, instead, users will be able to tweak them from within a text file (we'll need to switch to json or similar to have more flexibility and user-friendliness).

This should allow us to implement settings faster, to keep the UI clean for basic users, and to offer flexibility to advanced users.

I'd like to hear some feedback from @Teklad , @JuBan1 and @darealshinji

Teklad:
I don't have a problem with it. The code becomes more cluttered with each additional setting so this would be a good way to keep it cleaner.

Something along the lines of tweaks.json in the config directory would suffice.

Somebody could easily write a plugin later which works as a json editor on the config file if someone wants an interface to it.

JuBan1:
That's how I prefer it too. I dislike the lack of a proper settings menu because it forces users to read documentation on the syntax and possible settings of the json/xml/etc file just to make simple changes.

A possibility would be to allow modification of the settings file through a menu like Firefox's about:config. It'd allow us to show all available settings to a user and even give some hints or description but remove the need for having to create UI elements for each settings.

darealshinji:
I guess that's okay for some obscure and less common settings. I'd suggest to also add comments to the config file as a documentation and it should be mentioned in the manpage too.

danieleds:
Ok, so looks like we all agree!

Thanks to JuBan1's settings refactoring, replacing QSettings while keeping the same NqqSettings interface should be quite easy.

There's just one thing I want to better investigate, that is:
Suppose we switch to JSON, which has more flexibility than the current format. What changes (additions) to NqqSettings' interface could we need? Is it something that we can add later, or do we have to do it now?

So, for example, we might need to support nested categories. Or complex object values for more advanced settings.

JuBan1:
[...]

PR #610, opened 2018-04-10
feat: Add warning when opening large text files.

JuBan1:
Displays a warning dialog when opening files larger than X MiB.

img

By default this option is disabled and can be turned on in Preferences.

When testing file sizes I noticed that even just a 1MiB file makes Nqq seize up for seconds. Anything above 2MiB and you begin to wonder if it just crashed. So perhaps it's a good idea to enable this warning by default for files larger than 1MiB. Thoughts?


danieleds:
I like it, I'd make that the default.

I wouldn't expose it in the settings UI though. I mean, we should leave that setting in the settings file so that advanced users can tweak it, but I don't think most users will need to change it, so cluttering the UI isn't worth it.

JuBan1:
Sounds good. I guess that's what #301 is for.

PR #615, opened 2018-04-12
feat: Add option to backup documents periodically

JuBan1:
As discussed, the refresh of #340.

* Added an option to Preferences to enable/disable auto backup.
* Periodically saves changes to documents, tabs, and windows into the backup directory.
* Uses the Session mechanic to store those backups into `(NqqConfigDir)/autosaveCache`.
* Asks whether to restore from backup when a crash happened before.

The code's still got debug messages for debugging and I'm not done testing it. But the code is mostly ready for critique. I've been able to clean the old code up a bit so this PR is only a little larger while doing more.


danieleds:
Just a quick note: "autosave" feels to me like something that automatically saves *over* your files. What about "backup" or "fail-safe" or "crash protection"?

JuBan1:
Yes, I've been thinking about how to phrase it. The pref-ui says "Backup your documents every X seconds" but that seems misleading too, since the user can't interact with these backups directly.

danieleds:
There is a possible typo in autosave.cpp:55

JuBan1:
Updated. It's referred to as "Backup" now.

src/ui/Sessions/backupservice.cpp
```diff
@@ -0,0 +1,149 @@
...
+            continue;
+        }
+
+        if (isInOld && oldIter->isFullyEqual(*newIter)) {
+            // These windows are unchanged. Don't save them
+            savedData.push_back(*newIter);
+            continue;
+        }
+
+        // At this point the window is either new (!isInOld && isInNew) or changed (isInOld && !isFullyEqual(new,old)).
```
danieleds: Could you put a Q_ASSERT here?

src/ui/Sessions/backupservice.cpp
```diff
@@ -0,0 +1,149 @@
...
+    QDir autosaveDir(backupPath);
+    autosaveDir.setFilter(QDir::Dirs | QDir::NoDotAndDotDot);
+    const auto& dirs = autosaveDir.entryInfoList();
+
+    if (dirs.isEmpty())
+        return false;
+
+    auto ret = QMessageBox::question(nullptr,
+                          "",
+                          QObject::tr("It appears Notepadqq crashed. Do you want to restore the latest backup?"),
```
danieleds: "Notepadqq crashed" -> "Notepadqq was not closed properly"?
"restore the latest backup" -> "restore your unsaved changes"?

src/ui/Sessions/backupservice.cpp
JuBan1: You mean for the conditions inside the comments? They're only meant for people to follow the code logic. The code itself doesn't care which one is true.

The key point is that `isInNew==true` at this point, meaning the current item can be saved. This is always the case since the check in line 40 skips over items where `isInNew==false`.

JuBan1:
Had a nice idea when I was away. I didn't like the logic of `executeBackup`. It wasn't easy to follow. 

I already made use of `set_union` to simplify the logic but I didn't follow this approach consequently. Now I've replaced most of the messy code with `set_intersection` and `set_difference`. IMO the code's intent is much clearer now.

danieleds:
Hmm, there is a strange behaviour I'm seeing. When you open nqq, any tab that wasn't previously present within "tabCache" is never saved in "backupCache".

E.g., basic reproduction procedure, open nqq and open a new tab, then write something in it. That new tab is never saved in "backupCache".


Also, this should surely be enabled by default. We'll need to discuss about the default time interval though.

JuBan1:
Silly bug. I even wrote in the comment how `oldItem` always comes from the first range given to `set_intersection` yet I managed to grab `newItem` from the *wrong* range.

It didn't show up for me when testing because I didn't wait with making a new tab until *after* the first autosave.

danieleds:
Another minor idea that I had is to disable the timer whenever nqq goes out of focus. Nobody's editing anything while all the windows are out of focus, so that way we can reduce cpu wakeups.

For the same reason, we could use a VeryCoarseTimer like I just did for the Stats code: we don't care about millisecond precision, so a VeryCoarseTimer can cluster together multiple timer activations and, again, reduce power usage.

danieleds:
Just tested it with your fix and it works great!

danieleds:
Oh and another very important thing:

When a crash happened and Nqq asks to restore the backups, the backup folder **should not** be deleted. This prevents data loss when, for example, nqq starts crashing because of some malformation of one of the files.

JuBan1:
That's basically what happens. The backup directory is only cleared at `main.cpp:162` when `BackupService::enableAutosave` is called. At this point `BackupService::restoreFromBackup()` is already finished. If the program crashed on restoration (or when showing the MainWindow) the backup directory is kept.

danieleds:
Looks good to me. Do we need anything else?

What do we set as the default? I was thinking that 5 secs is probably a nice choice, but I'm not sure about the overhead when working with a lot of (large) files.

danieleds:
Yeah, Ubuntu is the kingdom of the crashes. Especially 17.10, for which we can't even easily update nqq because with the latest versions it segfaults on start like there's no tomorrow.

> I've been thinking about 20-30 seconds

Perhaps 15 is a good in-between? 30 feels to me like a lot :/
Anyway, I'm curious about @Teklad 's idea.

Teklad:
Most editors have the timer adjustable via settings... leaves the extra I/O caused by backups up to the user rather than hard-coded. This is especially useful for SSDs where excessive writes isn't desired in large amounts.

Another more advanced option would be to allow backups to be saved to wherever the user selects as the directory.  

This gives power users the option to use things such as RAM drive as an alternative means for backups to reduce disk I/O down to nothing if they so choose. I'm sure any user that chooses such an option is aware of the caveats of such a thing as well so it would be a useful feature for those types.

Teklad:
On that note... I'm okay with anywhere between 10-20 seconds.... If someone's WPM is so high that they can crank out THAT much in such a short period of time I'd applaud them.

JuBan1:
@Teklad Yes, the setting is adjustable. This is only about a sensible default. 

So I guess we'll go with 15 seconds. I've made the change.

@danieleds I've tested the system on a a window of ~3MiB of text opened and there was no lag when executing the autosave. So we should be good on that front.

danieleds:
I think it's everything! Let's finally merge it

PR #623, opened 2018-04-20
Update to adm-zip upstream code, minimal changes

jdeluyck:


danieleds:
Hi, thank you for the PR. Could you elaborate about any issue with the current version of adm-zip?

jdeluyck:
Real issues, none. But since I'm doing the work to get notepadqq included with Fedora, I noticed that it was 'old-ish'. Looking upstream there were some fixes that look useful (like doing path normalisation before using it, which was reversed in the code that was in your repo).

The other thing is that the latest code also comes with the license, which is practical. Not needed, but practical.

danieleds:
Good. Can you however update it using npm? I think some updates are missing in your PR (e.g. I don't see a changelog for package.json in node-modules/adm-zip)

jdeluyck:
Hm. Seems the code in your repo is the last one on npm. So I guess it hasn't been pushed out yet (tho it's also already a few months old).

danieleds:
It seems the last version is from 2015. It seems a good idea to update it, in order to deal with deprecation from node. You should be able to still update it via npm, just modifying our package.json to let adm-zip point to their github repo.

jdeluyck:
I admit to being a bit puzzled to what I'd need to change in it, though.

danieleds:
Basically, in extension_tools/package.json, you should replace the adm-zip version string with "cthackers/adm-zip#e116bc18df51".

Then you try updating via npm and it should work

jdeluyck:
Pushed the tag.

danieleds:
Reverted. Unfortunately it broke extensions.

PR #624, opened 2018-04-21
Keygrabber cleanup

Teklad:
This is my first pull request in a long list of pull requests that will be focusing on quality of life modifications and/or cleanup.

This one is pretty basic with just a few coding style adjustments and removal of some unused includes.  I'm going to try to keep these contained to just a few files at a time so reviews can be simple as well.

### DIFF
--- a/src/ui/frmpreferences.cpp
+++ b/src/ui/frmpreferences.cpp
@@ -289,7 +289,7 @@ void frmPreferences::loadShortcuts()
     QVBoxLayout *layout = new QVBoxLayout();
     QHBoxLayout *btnLayout = new QHBoxLayout();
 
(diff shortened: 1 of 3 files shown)


danieleds:
Can you keep the braces for single-line ifs? They may look bloated but they really help prevent scoping errors.

danieleds:
> I plead guilty to using them- especially in early exit scenarios (e.g. if (condition) return;)

That's exactly why a linter would be a PITA. Cases like these (`if (condition) return;`) are perfectly fine to me, as the probability of future mistakes is basically zero (single line which is a return... it's impossible to get it wrong). But at the same time, we want to enforce braces for all the other types of if statements.

I honestly don't know if any linter exists that allows us to say "ok, enforce braces but only if the body is not a single return"

PR #625, opened 2018-04-21
Experimental math rendering

danieleds:
Basic TeX rendering, very useful for those working in fields like machine learning.

example2

I still need to clean up the code and fix a few usability things before it's ready.


gamer01:
Bugs:
- [ ] does not work in enumerations
```
- jaccard similarity could be used $\sigma(A,B)= {|A\cap B|\over |A\cup B|}$
```
doesnt work, when removing the "-" it will automatically work
- [ ] Equations are rerendered when something is added to the line, even when the equation is not edited.

PR #629, opened 2018-04-24
Feature: Language Cache

Teklad:
Just pushing this in here so we can review over the changes and get everything tidied up.  About to head to work so looking for anything I might need to polish up for when I get back.

From what I can tell it seems to work just fine as-is, but it wouldn't hurt if someone could check for any regressions as well.  This is just the basic cache with no extra changes added as to keep the initial pull request smaller.

The end goal here is to have a user-based configurable language json so they can add additional languages and/or override internal ones without risking making nqq unusable in the event of an error.... Also it keeps people from having to poke at nqq's system files so that's always a bonus.

Another bonus:  all the languages are sorted automatically when you retrieve the list from the cache so the LanguagesGreater function is no longer required.  On top of this we no longer have to depend on having an editor instance in order to pull this information.


src/ui/include/EditorNS/languagecache.h
```diff
@@ -0,0 +1,102 @@
...
+    QString name;
+    QString mime;
+    QString mode;
+    QStringList fileNames;
+    QStringList fileExtensions;
+    QStringList firstNonBlankLine;
+    Language() {};
+    Language(const Language& o);
+    Language(Language&& o) noexcept;
+    Language& operator=(const Language& o);
```
JuBan1: For those assignments you could use `Language& operator=(const Language& o); = default` to save yourself having to define it manually.

src/ui/EditorNS/languagecache.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+const LanguageList& LanguageCache::languages()
+{
+    return m_languages;
+}
+
+int LanguageCache::lookupById(const QString& id)
+{
+    auto end = m_languages.constEnd();
+    for (auto it = m_languages.constBegin(); it != end; ++ it) {
```
JuBan1: This loops (and the following) should probably be using `std::find_if`.

src/ui/frmpreferences.cpp
```diff
@@ -141,27 +141,26 @@ void frmPreferences::on_buttonBox_accepted()
...
+    //"Default" language
+    ui->cmbLanguages->addItem("Default", "default");
+    LanguageSettings lang = {
+        "default",
+        ls.getTabSize("default"),
+        ls.getIndentWithSpaces("default"),
+        ls.getUseDefaultSettings("default")
+    };
+
+    for(const auto& l : LanguageCache::getInstance().languages()) {
```
JuBan1: Missign space in `for(`.

src/ui/include/EditorNS/languagecache.h
```diff
@@ -0,0 +1,102 @@
...
+    Language& operator=(Language&& o);
+};
+
+typedef QVector<Language> LanguageList;
+
+class LanguageCache
+{
+    public:
+        LanguageCache();
```
JuBan1: Constructor should be private.

src/ui/include/EditorNS/editor.h
```diff
@@ -332,7 +330,7 @@ namespace EditorNS
         QTextCodec *m_codec = QTextCodec::codecForName("UTF-8");
         bool m_bom = false;
         bool m_customIndentationMode = false;
-
+        Language m_language;
```
JuBan1: `m_currentLanguage` perhaps?

src/ui/include/EditorNS/languagecache.h
```diff
@@ -0,0 +1,102 @@
...
+        LanguageCache(LanguageCache const&) = delete;
+        void operator=(LanguageCache const&) = delete;
+
+        /**
+         * @brief Returns a Language struct at the given position by value
+         *        No bounds checking is performed in this function!
+         * @param i
+         * @return Language struct
+         */
+        Language operator [](int i) const {return m_languages[i];}
```
JuBan1: I think this can return `const Language&`.

src/ui/include/EditorNS/languagecache.h
```diff
@@ -0,0 +1,102 @@
...
+        
+        /**
+         * @brief Returns a reference to the Language struct at the given
+         *        position. No bounds checking is performed in this function!
+         * @param i
+         * @return Language struct reference.
+         */
+        Language& operator [](int i) {return m_languages[i];}
```
JuBan1: This should return `const Language&` or otherwise someone could modify the object. (`m_languages[i]` returns by reference too.)

src/ui/include/EditorNS/languagecache.h
```diff
@@ -0,0 +1,102 @@
...
+
+        /**
+         * @brief Returns a const reference to the Language struct at
+         *        the given position 'i'
+         * @param i
+         * @return const Language struct reference
+         */
+        const Language& at(const int& i) {return m_languages.at(i);}
```
JuBan1: With `operator[]` returning a const reference I think this function is redundant.

src/ui/ui.pro
```diff
@@ -74,6 +74,7 @@ SOURCES += main.cpp\
     EditorNS/bannerbasicmessage.cpp \
     EditorNS/bannerfileremoved.cpp \
     EditorNS/customqwebview.cpp \
+	EditorNS/languagecache.cpp \
```
JuBan1: Seems like indentation is wrong here. Tabs vs spaces. Same a little farther below.

JuBan1:
So much for syntax. Can't compile this stuff at the moment so I might be wrong with some comments.

Teklad:
Let me take a peek-e-poo.

src/ui/EditorNS/languagecache.cpp
Teklad:
```
int LanguageCache::lookupById(const QString& id)
{
    auto it = std::find_if (m_languages.constBegin(), m_languages.constEnd(),
        [&id] (const Language& l) {
            return (l.id == id);
        });
    if (it != m_languages.constEnd())
        return it - m_languages.constEnd();
    return -1;
}
```

Something like this?

src/ui/include/EditorNS/languagecache.h
Teklad: This was ported over rather directly when I was doing stranger things in webengine. I doubt we need it with how its setup now.

Teklad:
@JuBan1 feel free to double check and make sure I followed up on those changes properly.  I may have missed something.

src/ui/mainwindow.cpp
```diff
@@ -522,37 +522,23 @@ QList<QAction*> MainWindow::getActions() const
...
-        QChar letter = name.at(0).toUpper();
-
+    std::map<QChar, QMenu*> menuInitials;
+    for (const auto& l : LanguageCache::getInstance().languages()) {
+        QString id = l.id;
+        QChar letter = l.name.isEmpty() ? '?' : l.name.at(0).toUpper();
         QMenu *letterMenu;
-        if (menuInitials.contains(letter)) {
-            letterMenu = menuInitials.value(letter, 0);
+        if (menuInitials.count(letter)) {
```
JuBan1: Don't use implicit boolean conversion here please, since `count` returns `int`.

JuBan1:
Looks good. I'll see if I can actually test it this evening but from code inspection I don't expect any regressions.

Since one of my previous messages appears to have been lost in the void: did you earlier say that load times improved or worsened by ~100ms?

src/ui/mainwindow.cpp
```diff
@@ -522,13 +522,12 @@ QList<QAction*> MainWindow::getActions() const
...
 void MainWindow::setupLanguagesMenu()
 {
-    //ui->menu_Language->setStyleSheet("* { menu-scrollable: 1 }");
     std::map<QChar, QMenu*> menuInitials;
     for (const auto& l : LanguageCache::getInstance().languages()) {
         QString id = l.id;
         QChar letter = l.name.isEmpty() ? '?' : l.name.at(0).toUpper();
         QMenu *letterMenu;
-        if (menuInitials.count(letter)) {
+        if (menuInitials.contains(letter)) {
```
JuBan1: `std::map` doesn't have `contains` (an oversight if you ask me). `count` is fine but explicitly written as `menuInitials.count(letter) > 0` so C++ is not implicitly casting 0/1 to true/false.

Teklad:
@JuBan1, the load times are about the same right now... but once we load the javascript for the languages on the fly it appears it'll have about a 100ms improvement.

As for the last thing you commented on, its been a long day and I've been super sick for about a week so I didn't even look to see what kind of container I was using. xD

danieleds:
Is this ready? Also, did you by any chance already tested compatibility with the webengine2 branch?

Teklad:
@danieleds by all rights it should be compatible with it..... though there might need to be a few small changes to merge it in (not sure).  As far as I know it should be ready.  There doesn't appear to be any regressions.

danieleds:
Okay, so for now this is just a refactoring, right? Users don't yet have a file for custom langs, correct?

Teklad:
Nope... nothing over the top... Just put the cache in place so it can be added at a later date.  Wanted to keep this PR focused on getting the base system changed over first.

danieleds:
Perfect!

By the way, I think the name "LanguageCache" is a bit misleading now that we don't actually need to load them by webkit black magic anymore. What about "LanguageService"?

Teklad:
I know there was a reason in the old webengine branch but it evades me at
the moment.  I'll make the change today sometime.

Im fine with the name change as well.

With that said we should really do something about the file structure of
nqq at some point.  We've got widgets and all crammed into the toplevel
directory. Not too fond of uppercase letters in directory names either.

Might be better off to create an issue to discuss it after this webengine
madness is over.

danieleds:
Yeah, also the js editor would use a good refactoring. It's going to be easier after the changes I recently introduced, because now we can use requirejs to dynamically load stuff and declare encapsulated dependencies between files.

Teklad:
Oddly enough that's going to save me some work in regards to dynamically
loading language files later

JuBan1:
Some possible alternatives:

* Using `std::optional` for returning. It's a C++17 feature though, and I don't know if we want to require a C++17-compliant compiler at this time.

* Add an "invalid" language object that we can return.

* Return Plaintext language and add an optinoal `bool& found` parameter.

Teklad:
If I had to go a different route with it I'd prefer to use std::optional, but considering how new the feature is the LTS releases of some distros might not provide it.  We could use a pointer for now and switch over to std::optional when we feel C++17 is widespread enough.

Considering the only thing that really makes use of those functions is setLanguage, I don't think it'll be an issue for now.  Just finished testing using pointers and everything seems to work out.

src/ui/include/EditorNS/languageservice.h
```diff
@@ -30,36 +30,28 @@ struct Language {
...
-        /**
-         * @brief Returns a Language struct at the given position by value
-         *        No bounds checking is performed in this function!
-         * @param i
-         * @return Language struct
-         */
-        const Language& operator [](int i) const {return m_languages[i];}
-        
         /**
          * @brief Look up a language by its given Id and return an integer
```
JuBan1: The comments aren't updated yet.

JuBan1:
Perhaps, for the sake of const-correctness, we ought to use `const Language*` instead of `Langugae*`. I feel a bit queezyif someone can `lang->name = "foo"` to mess up the whole database.

Teklad:
That seems sane to me.  Done.  This should more or less complete this pull request.  Just gotta wait for @danieleds to crawl out of his cave and accept the pull. :p

src/ui/include/EditorNS/languageservice.h
```diff
@@ -41,26 +41,26 @@ class LanguageService
...
          *        containing the position of the language in the cache.
          * @param id
-         * @return int
+         * @return const Language pointer
          */
-        Language* lookupById(const QString& id);
+        const Language* lookupById(const QString& id);
 
         /**
          * @brief Look up a language by its file name and return an integer
```
JuBan1: Descriptions still wrong :P

src/ui/include/EditorNS/editor.h
```diff
@@ -227,7 +222,10 @@ namespace EditorNS
         Q_INVOKABLE void setZoomFactor(const qreal &factor);
         Q_INVOKABLE void setSelectionsText(const QStringList &texts, selectMode mode);
         Q_INVOKABLE void setSelectionsText(const QStringList &texts);
-        Q_INVOKABLE QString language();
+        Q_INVOKABLE QString getLanguageId() {return ((m_currentLanguage != nullptr) ? m_currentLanguage->id : "");}
+        Q_INVOKABLE QString getLanguageName() {return ((m_currentLanguage != nullptr) ? m_currentLanguage->name : "");}
+        Q_INVOKABLE QString getLanguageMime() {return ((m_currentLanguage != nullptr) ? m_currentLanguage->mime : "");}
+        Q_INVOKABLE QString getLanguageMode() {return ((m_currentLanguage != nullptr) ? m_currentLanguage->mode : "");}
```
JuBan1: In order to avoid all the nullptr checks perhaps we should just set m_currentLanguage to the plaintext language initially and ensure that calls to setLanguage() always either assign the correct new language to m_currentLanguage or plaintext.

Logically we'd never want an Editor that has no assigned language, so we could treat m_currentLanguage as a "non-null ptr" by ensuring we never set to it nullptr.

Teklad:
Good suggestion.  This is why I don't usually program right after work, lol.  I tend to overlook egregious things.

Teklad:
I could just make the LanguageService return plaintext as a default in the event it can't find a given language... That would remove basically all the checks in editor... though it might come with some behavior that users might not expect, lol.

JuBan1:
Can you make defaultLanguage be the actual plaintext object retrieved from the json? (Perhaps just ASSERT the existence of plaintext. If we can't find it we crash & burn)

For #630 I was planning to grab the file extensions for the tab's language to get a good default name for saving it. In the JSON file Plaintext has the "txt" file extension associated with it (which is great) but the defaultLanguage wouldn't have.

Teklad:
I believe I can... let me modify the initializer a bit.

Teklad:
Well actually I just made it where it uses lookupById to pull that value into the editor in the last commit.  I should definitely assert that its there during initialization though... its kinda mandatory.

Teklad:
So.... we can either assert that plaintext is there... or add it via code so its guaranteed to always be there.  Which would you prefer?

JuBan1:
I think its nice to keep all language modes in one place. So I'd assert if its not found in the json (or if the json couldn't be found).

Teklad:
Done... any other tidbits you need?

src/ui/EditorNS/editor.cpp
```diff
@@ -247,56 +242,60 @@ namespace EditorNS
...
-            newMode.insert("mime", mode.value("mime").toString());
-            newMode.insert("mode", mode.value("mode").toString());
-
-            out.append(newMode);
+        if (!m_customIndentationMode) {
+            setIndentationMode(lang->id);
         }
-
-        return out;
+        m_currentLanguage = lang;
```
JuBan1: Return early if `(m_currentLanguage == lang)` so we don't emit signals/etc.

JuBan1: And there's still the chance of `lang == nullptr`. In that case we should probably revert back to plaintext (or defaultLanauge).

Teklad:
Look good?

JuBan1:
Phew, I think we're done.

I should *actually* have some time to give it a test run today. Will do that when I get home. But unless there are any corner case issues left I am mollified.

Teklad:
Welp, from what I can tell there aren't any "known" regressions.  I've never been all that great at testing things thoroughly either, though. 

Also, I started testing your ote branch which is fairly impressive considering you just started not all that long ago.  It looks like you didn't use ksyntaxhighlight as a base directly, but rather rolled it up from scratch based on what you saw within it.

src/ui/EditorNS/editor.cpp
```diff
@@ -34,6 +34,7 @@ namespace EditorNS
 
     void Editor::fullConstructor(const Theme &theme)
     {
+        m_currentLanguage = LanguageService::getInstance().lookupById("plaintext");
```
JuBan1: Last issue: use setLanguage(nullptr) or setLanguage(LanguageService:: ...) here. Needed because setLanguage() does a few things like setting indentation.

You'll also have to move setLanguage(..) to the bottom of the page or we get stuck in the JS event loop because the page isn't loaded yet to handle the C_CMD_SET_LANGUAGE.

JuBan1:
With that that fix it's got my Okay but I assume danieleds will want to take another look too.

(I'm thinking about implementing the KSyntaxHighlighting definitions into OTE but I didn't want to base it directly on KSyntaxHighlighter because it wasn't precicely what we're looking for.)

danieleds:
I'm still outside and GitHub web UI sucks, however a couple questions:

- any reason why the Language detection by text content is implemented in the editor instead that in the LanguageService? Feels it's LanguageService's business.
- probably doesn't make any difference for performance, bit seems a good idea to make "const &" some of the QString arguments (e.g. when you pass filePath or rawTxt to the Editor)

Teklad:
I didn't feel the LanguageService having any concept of the actual contents of the editor seemed very logical.  That one made more sense when it was part of the javascript blob (at least in my opinion).  I'll still change it if you want though, just confirm with me on that one.

For the last thing... I can change the one for filePath, but not rawTxt due to how QTextStream is implemented.

danieleds:
Yeah move it to LanguageService... a signature like `const Language* lookupByContent(QString content);` should work

Teklad:
All done.

JuBan1:
Since that was the last issue I'll merge it in. Well done everyone.

Issue #632, opened 2018-04-26
Don't ask to save changes when closing an empty untitled tab


skalbers:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/notepadqq/notepadqq/blob/master/CONTRIBUTING.md#question
```

When opening a new tab in NP++ and typing something like 'aaa' and then removing it, one can close the tab without an additional request by the program.
In Nqq one always gets a Request like "Do you want to save changes to «new X» before closing?".

I couldn't find the spot in the code where this behaviour is checked, something like a check of the "fileSize" or length of the content in a new tab (with applying a trim or removing only whitespace-characters before this check) would suffice I think.

**Please tell us about your environment:**
OpenSuse Leap 42.1
Qt 5.5.1
Notepad 1.2.0 (the latest version with a compiled binary "notepadqq-v1.2.0-linux-x64.tar.gz")

danieleds:
I agree it is interesting behaviour, but I'm against being "too smart" in cases like this.

Just as a stupid example, consider if I want to save a small chunk of text (e.g. a random password, a small hash, whatever). I could type or paste it, then close the tab expecting nqq to ask me if I want to save it. When it doesn't, I've lost my text forever.

skleinealber:
Maybe I wasn't clear enough.
I expect this behaviour only, if the content size is zero (at the bottom 0 chars, 1lines) and it's in a new tab.

Teklad:
Should be as simple as marking as clean if content size is 0

JuBan1:
Can't do that because a named document (e.g. an actual file loaded from disk) would still be dirty when its contents are deleted.

This is a special case where the document is an orphan *and* it's empty.

PR #634, opened 2018-04-26
Refactored tab close logic. Don't ask to save empty orphan tabs anymore.

JuBan1:
See  #632


danieleds:
Avoid using goto here. Prefer if/elseif structure.

JuBan1:
The code used to use if-blocks and it ended up being several layers of nested code. I can use a lambda or function instead if you don't want to goto but I think this is (one of) the only legitimate use cases where goto makes sense.

danieleds:
Okay, let's keep it

PR #636, opened 2018-04-28
feat: Append file extension to suggested file names.

JuBan1:
When saving a "new X" tab we can suggest a file extension based on the highlighting/language mode selected. 

E.g. for a bash file the suggested name will now be `"new X.sh"`. For plain text files (the default) the suggested extension is `".txt"`.


danieleds:
I see you added a getLanguage method, could you remove the other redundant methods (e.g. getLanguageName, getLanguageMode, etc.)?

JuBan1:
Let me do that with another PR to keep things clean.

PR #638, opened 2018-04-28
feat: Unify document reload behavior. (fixes #603)

### LINKED ISSUE #603: Reopening a text file doesn't automatically reload its contents
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
```

**Current behavior**
Opening a text file that is already opened won't reload its contents. When the file was modified from outside Nqq you only get the ribbon telling you about it.

**Expected behavior**
Since the user explicitly re-opened the file it makes sense to reload its contents. If there are unsaved changes, Nqq should ask what to do with them. Basically, treat reopening a file the same as clicking the "Reload from Disk" action.

JuBan1:
Last one for today!

This slightly changes the default behavior of DocEngine:

* Opening a document that isn't already open will happen as normal.
* Opening a document that is already open (and in a clean state) will quietly reload it.
* Opening a document that is already open (and in a dirty state) will now ask the user if he wants to reload.
* Only exception is the `FileChangedBanner`. Clicking `Reload` there won't ask but simply reload. That's how it currently works.

That means it's now DocEngine's responsibility to ask for reloading. Previously this happened in Mainwindow.

### DIFF
--- a/src/ui/docengine.cpp
+++ b/src/ui/docengine.cpp
@@ -169,9 +169,9 @@ void DocEngine::loadDocuments(const DocEngine::DocumentLoader& docLoader)
         QFileInfo fi(localFileName);
 
         const QPair<int, int> openPos = findOpenEditorByUrl(url);
-        const bool isReloading = openPos.first > -1; //'true' when we're reloading a tab
+        const bool isAlreadyOpen = openPos.first > -1; //'true' when we're reloading a tab
 
-        if(isReloading && reloadAction == ReloadActionDont) {
+        if(isAlreadyOpen && reloadAction == ReloadActionDont) {
             EditorTabWidget *tabW = static_cast<EditorTabWidget *>
                                     (m_topEditorContainer->widget(openPos.first));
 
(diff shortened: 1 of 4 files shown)


Teklad:
I've not had a chance to test it out yet but it definitely makes more sense to handle it here.  Over time MainWindow has turned into a pretty large  conglomerate of code that we've not bothered to port to proper systems, unfortunately.

src/ui/docengine.cpp
```diff
@@ -130,8 +168,10 @@ void DocEngine::loadDocuments(const DocEngine::DocumentLoader& docLoader)
         QString localFileName = url.toLocalFile();
         QFileInfo fi(localFileName);
 
-        QPair<int, int> openPos = findOpenEditorByUrl(url);
-        if(!reload && openPos.first > -1) {
+        const QPair<int, int> openPos = findOpenEditorByUrl(url);
+        const bool isReloading = openPos.first > -1; //'true' when we're reloading a tab
```
danieleds: Just to use a more meaningful naming, this could be "isAlreadyOpen" instead of "isReloading"

PR #639, opened 2018-04-28
refactor: Update editor to use more C++11-isms

Teklad:
This just updates some of the older code in editor to use C++11 constructs.  Also fixes a small regression from the LanguageService PR where frmpreferences doesn't select the correct highlighting mode for editor preview.

### DIFF
--- a/src/ui/EditorNS/editor.cpp
+++ b/src/ui/EditorNS/editor.cpp
@@ -344,7 +344,10 @@ namespace EditorNS
 
     void Editor::setValue(const QString &value)
     {
-        setLanguage(LanguageService::getInstance().lookupByContent(value));
+        auto lang = LanguageService::getInstance().lookupByContent(value);
+        if (lang != nullptr) {
+            setLanguage(lang);
+        }
         sendMessage("C_CMD_SET_VALUE", value);
     }
 
--- a/src/ui/EditorNS/editor.cpp
+++ b/src/ui/EditorNS/editor.cpp
@@ -590,9 +590,7 @@ namespace EditorNS
     QList<Editor::Theme> Editor::themes()
     {
         auto editorPath = QFileInfo(Notepadqq::editorPath());
-        auto bundledThemesDir = QDir(editorPath.absolutePath() + "/libs/codemirror/theme/");
-
-        bundledThemesDir.setNameFilters({"*.css"});
+        QDir bundledThemesDir(editorPath.absolutePath() + "/libs/codemirror/theme/", "*.css");
 
         QList<Theme> out;
         for (auto&& theme : bundledThemesDir.entryInfoList()) {
--- a/src/ui/EditorNS/editor.cpp
+++ b/src/ui/EditorNS/editor.cpp
@@ -161,6 +161,7 @@ namespace EditorNS
                 }
             }
 
+
         } else if(msg == "J_EVT_READY") {
             m_loaded = true;
             emit editorReady();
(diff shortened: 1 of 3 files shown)


Teklad:
I've made a few extra changes since this PR was started.  If there's any you don't like I can revert each of them individually.  This was done intentionally so we can get the most out of this PR.

I'll continue to experiment with refactoring and make small changes here and there for a bit.

src/ui/EditorNS/editor.cpp
```diff
@@ -141,37 +141,33 @@ namespace EditorNS
...
             unsigned int id = rgx.capturedTexts()[1].toInt();
 
             // Look into the list of callbacks
-            for (auto it = this->asyncReplies.begin(); it != this->asyncReplies.end(); ++it) {
-                if (it->id == id) {
-                    auto cb = it->callback;
-                    it->value->set_value(data);
-                    this->asyncReplies.erase(it);
-
-                    if (cb != 0) {
```
JuBan1: Change you change that to `cb != nullptr`? It quenches clang warnings about using zero as nullpointer constant.

src/ui/EditorNS/editor.cpp
```diff
@@ -141,37 +141,33 @@ namespace EditorNS
...
+            auto it = std::find_if(asyncReplies.begin(), asyncReplies.end(), [&id] (const AsyncReply& reply) {
+                return (reply.id == id);
+            });
+            if (it == asyncReplies.end())
+                return;
+            auto cb = it->callback;
+            it->value->set_value(data);
+            this->asyncReplies.erase(it);
 
+            if (cb != 0) {
```
JuBan1: Change you change that to `cb != nullptr`? It quenches clang warnings about using zero as nullpointer constant.

danieleds:
We have to check if it compiles on Ubuntu 14.04. I've had a few bad experiences in the past where I introduced some nice new syntax and everything failed compiling when I was shipping the deb.

danieleds:
Also because of that, can you fix the regression with another PR?

Teklad:
So is Ubuntu 14.04 the bare minimum we'll support?  I'd figure we'd go for 16.04 considering the looming WebEngine conversion and other what nots.

Also, sure thing... let me cherry-pick that out of there real quick.

danieleds:
Unfortunately, yes. Ubuntu 14.04 is still alive and a lot of users (you'd be surprised) are using it.

I still need to prepare something real quick to show you the stats we collect.

We'll stop sending updates after webengine2 is ready, but for now we have this weight

Teklad:
I look forward to that day.  We've got a lot of spaghetti thrown around everywhere that could be greatly improved using proper C++11 syntax.

src/ui/EditorNS/editor.cpp
```diff
@@ -141,37 +139,33 @@ namespace EditorNS
         if (msg.startsWith("[ASYNC_REPLY]")) {
             QRegExp rgx("\\[ID=(\\d+)\\]$");
 
-            if(rgx.indexIn(msg) == -1)
-                return;
-
-            if (rgx.captureCount() != 1)
+            if (rgx.indexIn(msg) == -1 || rgx.captureCount() != 1)
```
danieleds: Messing with on_proxyMessageReceived is going to create a lot of merge conflicts in the webengine2 and webengine2-promise branches I'm working on... It's getting hard to fix conflicts at every PR xD.

I think it's not worth the work, especially because this code is going to disappear anyway. Could you revert these changes, or at least do them on those branches?

Teklad:
Done.  Even reverts tend to get a little hairy after enough commits have been made, lol.

danieleds:
In Editor::print(), can you correct the theme name from "Default" to "default"?

Teklad:
Done.

Issue #648, opened 2018-05-01
Regression: wrong window title on startup


danieleds:
Some of the latest commits introduced this regression:

 - when opening nqq with some unnamed file in the session, it displays its path (as shown in the picture)
 - when opening nqq without any file in the session, the title is `- Notepadqq` instead of `new 1 - Notepadqq`.

The problem disappears as soon as you write something.

![screenshot from 2018-05-01 17-54-45](https://user-images.githubusercontent.com/402652/39480694-48c2d4f2-4d69-11e8-928b-72bb00f44402.png)

Teklad:
I did a few tests.  If I have two new tabs open in a session and I swap back and forth between them after startup it fixes itself.  Something is running right after a session load(or maybe right before), that's causing this to happen.  Possibly due to the performance increase in language lookups, but I'm unsure.

Considering the nature of the bug it's gotta be something in mainwindow.cpp or sessions.cpp.  @JuBan1 if you wanna look around too it would be appreciated.  I think this bug may have been there prior, patiently waiting for us to poke at it to reveal itself.  LanguageCache doesn't touch hardly anything related to this bug, yet this commit somehow caused it.

JuBan1:
Will take a look later today. 

And when the webengine branch is merged the whole tabs system is gonna get some serious reworks.

Teklad:
Yea that tab system is a real mess.  There's quite a bit of indirection going on in there so tracking down even the smallest bug is becoming quite the burden.

JuBan1:
> since ce1d67a

Yeaaah... that was another fun one.

Basically, before the patch this happens in sessionRestore for every document that gets loaded:

* Create new tab (calls `refreshEditorUi` but with wrong title)
* Set tab title to correct name
* Call `setLanguage(..)`  (calls `refreshEditorUi` with correct title)

After the patch it's 

* Create new tab (calls `refreshEditorUi` but with wrong title)
* Set tab title to correct name
* Call `setLanguage(..)`  (which does nothing because Editor is alrady set to plaintext)

So for all loaded documents that had plaintext as their language mode we never call `refreshEditorUi` after setting the tab title.

#650 fixes this is the most stupid, yet most convenient way. Turns on `Editor->markDirty()` ends up triggering `refreshEditorUi` too, and there's a mandatory call to `markDirty()` for every tab that initially has the wrong tab title. So I moved that bit of code to come after setting the tab title.

danieleds:
[...]

PR #650, opened 2018-05-01
fix: Show correct file name in title bar at startup.

JuBan1:
See comment in #648 

PS: It's meant as a temporary fix. Once webengine is merged and we can make larger changes without having to worry about rebasing branches all the time we should make some thorough cleanup effort.


JuBan1:
I told you it was the most stupid fix! (That's what I get for writing patches tired)

Will change it later today.

JuBan1:
Reverted the commit and used a different way.

The issue is that a direct call to `refreshEditorUiInfo` isn't all that simple. Session doesn't know about what MainWindow it loads the tabs into, and there's the possibility of having multiple open windows.

Instead, I made the active editor emit a final `cleanChanged` signal which will cause function to run only on the MainWindow instance it is attached to.

danieleds:
> spontaneous combustion

We could turn that into a cool feature!

PR #651, opened 2018-05-03
Add new program parameters to open file:

moses-nebogipfel:
- 'read-only' to open (one) file in read-only mode
- parameters to find (in one file) passed text:
  - 'search'
  - 'matchCase'
  - 'matchWholeWord'

Motivation:
In my daily work I often need review some text files especially to find some text inside them.  I'd like to be sure doing this work I don't modify accidentally files. Moreover I want to bind noepadqq with my file manager so these new parameters are necessary to do it.

### DIFF
--- a/src/ui/Search/frmsearchreplace.cpp
+++ b/src/ui/Search/frmsearchreplace.cpp
@@ -187,6 +187,32 @@ int frmSearchReplace::selectAll(QString string, SearchHelpers::SearchMode search
     return count.toInt();
 }
 
+void frmSearchReplace::setSearchMode(SearchHelpers::SearchMode searchMode)
+{
+    switch(searchMode)
+    {
+    case SearchHelpers::SearchMode::SpecialChars:
+        ui->radSearchWithSpecialChars->setChecked(true);
+        break;
+    case SearchHelpers::SearchMode::Regex:
+        ui->radSearchWithRegex->setChecked(true);
+        break;
+    default:
+        ui->radSearchPlainText->setChecked(true);
+    }
+}
+
+void frmSearchReplace::setSearchOptions(SearchHelpers::SearchOptions searchOptions, bool toggleAdvancedSearchChk)
+{
+    ui->chkMatchCase->setChecked(searchOptions.MatchCase);
+    ui->chkMatchWholeWord->setChecked(searchOptions.MatchWholeWord);
+
+    if(toggleAdvancedSearchChk) {
+        const bool checkedState = searchOptions.MatchCase || searchOptions.MatchWholeWord;
+        ui->chkShowAdvanced->setChecked(checkedState);
+    }
+}
+
 SearchHelpers::SearchMode frmSearchReplace::searchModeFromUI()
 {
     if (ui->radSearchPlainText->isChecked())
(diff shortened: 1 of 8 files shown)


moses-nebogipfel:
Hi @danieleds, 
I'd like to invite you to familiar with my new feature proposition. I'm new on GitHub and I don't know yet how to invite participant to review so I'm doing it by this message.

danieleds:
Hi, thank you for your patch.

This adds quite a bit of complexity though, and I'm not sure how many users could benefit from this.
Are you sure that a text *editor* is the right choice for what you're trying to accomplish? Isn't opening the file in a web browser enough?

I'm just trying to better understand the use case and complexity/benefits balance. @JuBan1 , @Teklad what do you think?

<sup>P.S. in any case, I would apply the "--read-only" flag even when multiple files are present</sup>

JuBan1:
I'm not happy about the search, matchCase, matchWholeWord switches. I think read-only is nice but I'm not sure how useful a cmd-line flag for it is. In N++ each individual tab can be set to read-only.

moses-nebogipfel:
Regarding use cases:
I can imagine totalcommander-like tool which provides possibility to change default text file viewer. Such kind applications usually have two modes of view file. First one is standard modification mode (F4 is common shortcut for this). Second one is read-only view file (F3) so this is the use case for `read-only` option. 

Let's talk about `search` and other parameters. In the same totalcommander-like application that I mentioned above I've got option to search files through file system by content. The search mechanisms often have similar advanced options like matching whole word or case-sensitive. Inside searchs output I can review result files to search phrase which I interested in. I think that it'd be nice to propagate advanced search options from file manager app to notepadqq so that getting consistent search inside file with file manager find mechanism.

Teklad:
This is definitely one of those better as an extension kind of case.  We've
already got quite a bit of feature creep so this seems like unnecessary
complexity for little gain.

I'll be working on a new plugin system soon which should allow out of
source changes like this

moses-nebogipfel:
Ok, I can agree that `search` parameter's use cases seem quite limited  and it would introduce kind of complexity. Thanks guys for these opinions. 

I suggest therefore remove that implementation but I would propose to keep `read-only` parameter. Moreover I offer extend `read-only` behaviour so that it apply to more files (tabs). I would add short version param `ro` as well to keep consistency with notepad++ and also add read-only option to tab's context menu.

src/ui/Search/frmsearchreplace.cpp
```diff
@@ -187,6 +187,32 @@ int frmSearchReplace::selectAll(QString string, SearchHelpers::SearchMode search
     return count.toInt();
 }
 
+void frmSearchReplace::setSearchMode(SearchHelpers::SearchMode searchMode)
```
JuBan1: These two functions are left-overs from the other command-line functions, aren't they? Since they're not used, please remove them again.

src/ui/include/EditorNS/editor.h
```diff
@@ -230,6 +230,7 @@ namespace EditorNS
         Q_INVOKABLE void setEOLVisible(const bool showeol);
         Q_INVOKABLE void setWhitespaceVisible(const bool showspace);
         Q_INVOKABLE void setMathEnabled(const bool enabled);
+        Q_INVOKABLE void setReadOnly(const bool readOnly);
```
JuBan1: Could you also add a `bool isReadOnly() const`?

src/ui/Search/frmsearchreplace.cpp
moses-nebogipfel: Actually I left these methods intentionally. I assumed that such methods might come in handy in the future.

src/ui/include/EditorNS/editor.h
moses-nebogipfel: I'll do. I'm working on improvements this feature and I already had need to use `isReadOnly` :)

PR #652, opened 2018-05-03
fix: avoid unnecessary isClean call on start

danieleds:
@JuBan1 does it have any negative side effect?


danieleds:
the smallest PR ever

PR #675, opened 2018-05-12
fix: Remove white flash when opening new tab

JuBan1:
During page load the page just shows a white canvas which is very unpleasing to the eye. It's better to use the current color palette's background color. 

Even better would be to query the current theme's background color but we don't have an easy way of doing that.


danieleds:
Nice!
I noticed it only happens the first time we switch to that tab... maybe there is a way to pre-render it in the background.

PR #678, opened 2018-05-14
feat: Checks API

Teklad:
This enables Checks API integration on Github.  This will allow you to click on the "Checks" tab to view whether the formatting job failed or not without having to look deep in Travis itself.

This is closer to what we were looking for in the first place.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,9 @@
 language: cpp
 sudo: false
 
-matrix:
-  fast_finish: true
-  allow_failures:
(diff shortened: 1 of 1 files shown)


Teklad:
For some reason this works on my repository but fails here.  Are we using the Github App for Travis CI currently?  Support doesn't exist in the old Travis API.

danieleds:
It's because we're on the legacy travis-ci.org, it won't work until we move to travis-ci.com.

The migration is not automatic unfortunately, so I sent an email yesterday and I'm waiting for their action.

PR #680, opened 2018-05-14
refac: Use clang-format to sort all of our include files.

Teklad:
So this is the result of the regex I came up with to sort our include files.  Everything is categorized based on the header type.  First is the direct header file for definitions, then other local headers, Qt, then standard std.

This can be re-ran whenever is necessary, so keeping it maintained at a later date isn't really too big an issue if it comes up again.


danieleds:
We can't enforce this with the linter, right?

Teklad:
Actually I do believe the linter might pick up on this.  There's no way to really tell until someone adds an include though.  Hopefully yes, but I won't make any promises.

PR #681, opened 2018-05-14
refac: mainwindow initialization refactor

Teklad:
Refactored a lot of the code we've been just tossing into the constructor.  Moved most of the constructor code into their own contained functions.

Also managed to cut down on a lot of repetitive code using a lamba inside setupStatusBar().  All in all I think this is a good change.

### DIFF
--- a/src/ui/frmpreferences.cpp
+++ b/src/ui/frmpreferences.cpp
@@ -357,7 +357,7 @@ void frmPreferences::saveToolbar()
     m_settings.MainWindow.setToolBarItems(string);
 
     for (auto* wnd : MainWindow::instances())
-        wnd->loadToolBar();
+        wnd->setupToolBar();
 }
 
 bool frmPreferences::applySettings()
--- a/src/ui/include/mainwindow.h
+++ b/src/ui/include/mainwindow.h
@@ -71,11 +71,10 @@ public:
     QList<QAction*> getActions() const;
     QList<const QMenu*> getMenus() const;
 
-    // Creates or re-creates the window's main tool bar.
-    void loadToolBar();
+    void setupRunMenu();
+    void setupToolBar();
 
     DocEngine*  getDocEngine() const;
-    void generateRunMenu();
 public slots:
     void refreshEditorUiInfo(Editor *editor);
 
@@ -234,7 +233,6 @@ private:
      */
     bool                finalizeAllTabs();
 
-    void                createStatusBar();
     int                 askIfWantToSave(EditorTabWidget *tabWidget, int tab, int reason);
 
     /**
@@ -266,10 +264,17 @@ private:
     int                 save(EditorTabWidget *tabWidget, int tab);
     int                 saveAs(EditorTabWidget *tabWidget, int tab, bool copy);
     QUrl                getSaveDialogDefaultFileName(EditorTabWidget *tabWidget, int tab);
+
+    void                setupDocEngine();
+    void                setupIcons();
+    void                setupKeyboardShortcuts();
     void                setupLanguagesMenu();
+    void                setupStatusBar();
+    void                setupTabContextMenu();
+    void                setupUserInterface();
+
     void                transformSelectedText(std::function<QString (const QString &)> func);
     void                restoreWindowSettings();
-    void                loadIcons();
     void                updateRecentDocsInMenu();
     void                convertEditorEncoding(Editor *editor, QTextCodec *codec, bool bom);
     void                toggleOverwrite();
@@ -288,14 +293,8 @@ private:
     /**
      * @brief Workaround for this bug: https://bugs.launchpad.net/ubuntu/+source/appmenu-qt5/+bug/1313248
      */
-    void                fixKeyboardShortcuts();
     void                instantiateFrmSearchReplace();
     QUrl                stringToUrl(QString fileName, QString workingDirectory = QString());
-
-    /**
-     * @brief Initialize UI from settings
-     */
-    void initUI();
 
     /**
      * @brief Update symbol options using parameter `on` and Show_All_Characters toggle status.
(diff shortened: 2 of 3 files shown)


Teklad:
May I get a bit of love for this PR? Lol.

danieleds:
I need to review this, it is very difficult to check for regressions in the current state.

If possible, splitting this PR into smaller "setup*()" would be better. The current PR says "455 lines changed"; an average of about 100/110 per PR would be great.

Teklad:
I'm pretty sure you know how I am at this point, lol.  I get to working on an issue and go "ooo this could be better, and that could be better"....  before I even realize it I've changed a few hundred lines of code and my PR exploded into something much larger than I had initially anticipated.

At least during refactoring, trying to keep the PRs cut into smaller change sets isn't such a bad idea.  I'll see if I can't split this into 3 or 4 smaller ones.  I see quite a few glaring issues in mainwindow that could be easily remedied (did you know we still use foreach() in places there?).

There's also the abysmal startup performance issue I need to address since switching over to QWebEngine (it takes a full 2 seconds for NQQ to open on my machine with no session saved).  I think I know the cause of that, though.

danieleds:
> I get to working on an issue and go "ooo this could be better, and that could be better".... before I even realize it I've changed a few hundred lines of code and my PR exploded into something much larger than I had initially anticipated

Ahah, I know the feeling :D

> There's also the abysmal startup performance issue I need to address since switching over to QWebEngine (it takes a full 2 seconds for NQQ to open on my machine with no session saved). I think I know the cause of that, though.

That *should* be because of a couple messages going to the editor that still need to be converted to promises. Or did you spot something different?

JuBan1:
Ah, this flew below my radar. I can test for regressions too later today. But from looking over it I can't see any potential issues.

JuBan1:
Runs well for me. I can't see any issues.

(P.S. the startup time is indeed not great at the moment.)

danieleds:
Looks fine in general to me too, but since there are a lot of reorderings, and since timing is very important especially for our new async setup, it's better to stay on the safe side and evaluate each change separately :)

I don't want to risk adding any regression at this point, above all in the app loading phase.

danieleds:
Mostly yes, however for example setupIcons() is now after setupTabContextMenu() (it was before it), setupKeyboardShortcuts() is now at the end (it was after setupTabContextMenu()), and a few others.

I'm not saying that the previous order made more sense (actually, it's pretty messy and probably doesn't make any sense at all). In fact, this refactoring is waaaay better. But exactly since the previous code was messy, I want to be careful with regressions that may go unnoticed in this delicate part of nqq.

Teklad:
The only problem with the previous code being messy is larger PRs to fix the mess are almost inevitable.  Even just unifying the user interface initialization went well over 120 line changes (If you believe all that crud that git spits out anyways)... In actuality not much was changed, but shuffled around.

Do we even still support AppMenu, btw?  That's all fixKeyboardShortcuts was originally for, which is why it ended up at the bottom unified with connecting the actual keyboard shortcuts in the code.

Teklad:
In commit 67e5df83fc9e5ac097965576274a582b8d8a0b7e you modified a line in main.cpp which stated "Improve startup time" in the commit message.  The issue, however... is that this line actually negatively impacts startup performance heavily.

Since we have openCommandLineProvidedURLs in MainWindow's constructor we're already guaranteed an editor anyways.  addEditorToBuffer was initially in main.cpp as a workaround for some QWebKit bug (I don't recall which one).

If I remove this line completely my startup performance is actually on par with the old QWebKit branch.... possibly faster even.  As an added bonus the memory usage of NQQ drops considerably (about 50MB for fresh NQQ with no session).

I'm not sure how it affects performance for everybody, however.

danieleds:
> The issue, however... is that this line actually negatively impacts startup performance heavily.

Interesting. How many tabs did you have in the session?

That line is just a heuristic I added when one has many tabs. This is the rationale:

 * each editor, when initialized, has to load the HTML page with the scripts and emit the J_EVT_READY event (call this "initTime").
 * since we've still got some synchronous messages in the startup process, each editor is initialized *sequentially*
 * this means that if we've got *n* tabs to open, we have to wait *n* times the loading of the html page and the scripts and J_EVT_READY (which are independent from c++ so they could run in parallel)... so *n \* initTime*
 * with a buffer of 1, like before, startup time for 1 tab in the session is maximum, but startup time for *n* tabs in the session is still high: *n \* initTime - 1*
 * with a buffer of k, the startup time becomes worse for 1 tab in the session, but when we have *n* tabs in the session the time becomes *n/k \* initTime* which is a nice improvement.

So we have to play with the value of k, I think.
An interesting thing to do would be to set "k = number of tabs to open", but we have to somehow do this as early as possible, because if done in the session loading code, it is already too late (I had tried this).

Teklad:
The actual construction of the QWebEngineView is not asynchronous from the looks of it.  You can actually see this by doing something similar in a basic project with nothing but QWebEngine initialized 15-20 times.

The startup performance degrades rapidly due to just how enormous it is.

danieleds:
Hmm, on Qt 5.10 they go in parallel.

Anyway, this is surely something to be optimized.

Teklad:
If the whole single thread for User interface thing still applies to QT like it has historically then the actual creation of the widget would be in linear time. Im on QT 5.10 as well so its definitely not a version issue.  My desktop is old but by no means ancient.  Lol 

I had an idea many moons ago about swapping out qwebenginepage objects on a single view but its a fairly large change that I will need to make time to do without causing any regressions.

danieleds:
> I had an idea many moons ago about swapping out qwebenginepage objects on a single view

that would be an awesome change but must be thought carefully (also with respect to "split view" etc)

PR #689, opened 2018-05-20
refac: Allow the status bar to show tooltip messages.

Teklad:
This change modifies how the status bar currently operates.  As it stands we were using quite a few hacky methods to get the status bar to display properly.

I've made some slight modifications which cleans up a lot of the status bar code and allows some of its UI elements to be updated via signals/slots instead of a request.  Namely content, selections, and current char/line.

In order for this to work properly I had to move the current language status label to the right side, which isn't necessarily a huge change but could take a little getting used to.  This will enable us to add status tips to all of our actions which display when you hover over one of them.

Here's how it looks now:

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -667,6 +667,17 @@ UiDriver.registerEventHandler("C_CMD_EOL_TO_SPACE", function(msg, data, prevRetu
     editor.setValue(text.replace(/\n/gm," "));
 });
 
+function getActivityInfo()
+{
+    var map = new Object();
+    var selections = editor.getSelection("\n");
+    var cursor = editor.getCursor("head");
+    map["cursor"] = [cursor.line, cursor.ch];
+    map["selections"] = [selections.split(/\r\n|\r|\n/).length, selections.length];
+    map["content"] = [editor.lineCount(), editor.getValue().length];
+    return map;
+}
+
 $(document).ready(function () {
     editor = CodeMirror($(".editor")[0], {
         lineNumbers: true,
@@ -712,12 +723,17 @@ $(document).ready(function () {
         UiDriver.sendMessage("J_EVT_CLEAN_CHANGED", isCleanOrForced(changeGeneration));
     });
 
-    editor.on("cursorActivity", function(instance, changeObj) {
-        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY");
+    editor.on("cursorActivity", function(instance) {
+        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY", getActivityInfo());
     });
 
     editor.on("focus", function() {
         UiDriver.sendMessage("J_EVT_GOT_FOCUS");
+        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY", getActivityInfo());
+    });
+
+    $(document).mouseenter(function() {
+        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY", getActivityInfo());
     });
 
     UiDriver.sendMessage("J_EVT_READY", null);
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -667,7 +667,7 @@ UiDriver.registerEventHandler("C_CMD_EOL_TO_SPACE", function(msg, data, prevRetu
     editor.setValue(text.replace(/\n/gm," "));
 });
 
-function getActivityInfo()
+function getDocumentInfo()
 {
     var map = new Object();
     var selections = editor.getSelection("\n");
@@ -677,6 +677,13 @@ function getActivityInfo()
     map["content"] = [editor.lineCount(), editor.getValue().length];
     return map;
 }
+
+/**
+* @brief Replies to the request for document information. 
+*/
+UiDriver.registerEventHandler("C_CMD_GET_DOCUMENT_INFO", function(msg, data, prevReturn) {
+    UiDriver.sendMessage("J_EVT_DOCUMENT_INFO", getDocumentInfo());
+});
 
 $(document).ready(function () {
     editor = CodeMirror($(".editor")[0], {
@@ -724,16 +731,11 @@ $(document).ready(function () {
     });
 
     editor.on("cursorActivity", function(instance) {
-        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY", getActivityInfo());
+        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY", getDocumentInfo());
     });
 
     editor.on("focus", function() {
         UiDriver.sendMessage("J_EVT_GOT_FOCUS");
-        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY", getActivityInfo());
-    });
-
-    $(document).mouseenter(function() {
-        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY", getActivityInfo());
     });
 
     UiDriver.sendMessage("J_EVT_READY", null);
--- a/src/ui/include/mainwindow.h
+++ b/src/ui/include/mainwindow.h
@@ -80,6 +80,7 @@ public:
     void generateRunMenu();
 public slots:
     void refreshEditorUiInfo(Editor *editor);
+    void refreshEditorUiCursorInfo(QMap<QString, QVariant> data);
 
 protected:
     void closeEvent(QCloseEvent *event);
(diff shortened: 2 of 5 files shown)


Teklad:
And here's what we can use it for:
2018-05-20-014937_768x535_scrot

JuBan1:
Having language selection on the right-hand side is going to be weird, but I do like the temporary messages!

One suggestion though- I like the changed frame style around the labels but it does give them a button-like look. That's great for the two labels that can be clicked on, but a bit misleading for the other two. 

I suggest either removing the frame from non-clickable labels or making all labels clickable (which I think is a better idea). Clicking on the overtype label could change between INS and OVR modes, and the line-endings label can show the "end of line" menu from the Edit menu.

Teklad:
Made them both clickable.  I think this is more in line with what we want anyways.  Code is extremely simplified compared to what it was before.

danieleds:
Emitting "J_EVT_CURSOR_ACTIVITY" when there has not been any cursor activity seems like a harmful hack. It seems a better idea to create a new dedicated signal for that.

+1 for the clickable labels

Teklad:
Well the way the entire signal is handled is different from what it was before... If I'm not mistaken it's actually being transmitted as-needed now vs. whenever we feel like requesting it(how it was before)... the name of the signal is the same but in all reality it operates quite a bit differently.

The final outcome ends up the same, however.

src/ui/include/mainwindow.h
```diff
@@ -207,9 +206,6 @@ private slots:
     QMenu*                m_tabContextMenu;
     QList<QAction *>      m_tabContextMenuActions;
     QLabel*               m_statusBar_fileFormat;
```
JuBan1: These remaining `QLabel*`s should probably all just be `ClickableLabel*`s instead. I see no reason to have to up/down-cast them.

src/editor/app.js
```diff
@@ -712,12 +723,17 @@ $(document).ready(function () {
...
+        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY", getActivityInfo());
     });
 
     editor.on("focus", function() {
         UiDriver.sendMessage("J_EVT_GOT_FOCUS");
+        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY", getActivityInfo());
+    });
+
+    $(document).mouseenter(function() {
+        UiDriver.sendMessage("J_EVT_CURSOR_ACTIVITY", getActivityInfo());
```
JuBan1: Why's this necessary? Isn't this basically the "on hover"-event?

Teklad: CodeMirror doesn't have a native onHover event so this is the next best thing available, unfortunately.

Otherwise the status text would remain blank until the user moves the cursor to a different postiion.

src/ui/include/mainwindow.h
Teklad: I agree with this.

src/editor/app.js
JuBan1: I mean, why is it necessary at all to have this? The `"J_EVT_CURSOR_ACTIVITY"` signal is also emitted on focus; I'd assume that ought to be enough.

JuBan1: Hm is it possible to use `CustomWebView::gotFocus()` instead? That one should work fine.

Teklad: That's not a bad idea... there's a few things I might have to change to get it to work but its likely the "proper" way to do it assuming it works.

I've noticed QWebEngine loves to eat events so the typical standard "it should work" might not apply here.  We'll see, however.

JuBan1: I was thinking bring back `refreshEditorUiCursorInfo` and call it through `on_cursorActivity` and `on_currentTabChanged` or whatever the proper signal is. Pretty sure we're already handling CustomWebView's focus event somewhere in MainWindow.

JuBan1:
Like the buttons. I suggest to also add a `btn->setFocusPolicy(Qt::NoFocus);` to the button creation so the selection outline won't stay behind.

There's also an issue with the status bar: as soon as I mouse over any item in the menu bar the cursor information disappears and won't show up again unless I move the cursor or switch tabs.

What I can find is that according to the [Qt Docs](https://doc.qt.io/qt-5/qstatusbar.html) we should use a QLabel to display our cursor information (see *Normal* status indicators).

Doing that- after some struggle -the status bar seems to work fine. It shows cursor information unless I hover over an action that has `statusTip` set to anything non-empty. And as I move away again the cursor info shows up as well.

JuBan1:
P.S. If we keep the PushButtons I believe that makes ClickableLabel unused?

src/ui/include/mainwindow.h
```diff
@@ -206,10 +206,10 @@ private slots:
...
     QMenu*                m_tabContextMenu;
     QList<QAction *>      m_tabContextMenuActions;
-    QPushButton*          m_sbFileFormatBtn;
-    QPushButton*          m_sbEOLFormatBtn;
-    QPushButton*          m_sbTextFormatBtn;
-    QPushButton*          m_sbOvertypeBtn;
+    QPushButton* m_sbFileFormatBtn;
+    QPushButton* m_sbEOLFormatBtn;
+    QPushButton* m_sbTextFormatBtn;
+    QPushButton* m_sbOvertypeBtn;
```
JuBan1: This one's another annoying bit about clang-format. It's not too bad with several variables of the same type, but if we were to remove all the extra indentation it would definitely become less readable.

Teklad: Yea.... I'm fairly fond of how we do header files so it might not be a bad idea to restrict that.

Teklad:
So basically store the cursor data in a single QLabel rather than putting it in the message window so when the message is done being displayed we get our data back?

JuBan1:
Basically. The status bar knows permanent, normal, and temporary indicators. Permanent indicators won't ever be hidden. Normal indicators can be hidden by temporary ones but will show up again once the temporary message disappears. So that's exactly what we want.

danieleds:
Just a useless stylistic note: instead of that border, I would highly prefer a completely flat statusbar (like before), but with a different shade on mouse-over in order to show that it can be clicked.

Teklad:
The buttons I replaced the clickable label with make it flat again.  Also has a little arrow indicating clicking it will display a menu, which is a nice cosmetic addition (imo):
2018-05-21-043723_739x577_scrot

I made the changes @JuBan1 suggested and it appears to be working a lot better this way.

src/ui/mainwindow.cpp
```diff
@@ -324,95 +323,24 @@ void MainWindow::loadIcons()
...
-    scrollArea->setFixedHeight(frame->height());
+    m_sbDocumentInfoLabel = new QLabel;
+    m_sbDocumentInfoLabel->setMinimumWidth(1);
+    statusBar()->addWidget(m_sbDocumentInfoLabel);
+    auto createStatusButton = [&](const QString& txt, int minWidth, QMenu* mnu = nullptr) {
+        auto* btn = new QPushButton(txt);
+        btn->setFlat(true);
+        btn->setMenu(mnu);
+        btn->setMinimumWidth(minWidth);
+        btn->setMaximumWidth(minWidth);
```
JuBan1: I don't think we should set maximum width. Without it, the buttons will resize to fit their labels. [The way it is right now text gets cut off frequently](https://i.imgur.com/iW748gk.png).

src/ui/mainwindow.cpp
```diff
@@ -324,95 +323,24 @@ void MainWindow::loadIcons()
...
-    label->setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Fixed);
-    label->setMinimumWidth(40);
-    layout->addWidget(label);
-    m_statusBar_overtypeNotify = label;
-
-
-    status->addWidget(scrollArea, 1);
-    scrollArea->setFixedHeight(frame->height());
+    m_sbDocumentInfoLabel = new QLabel;
+    m_sbDocumentInfoLabel->setMinimumWidth(1);
```
JuBan1: Why set minimum width? It doesn't seem to make a difference.

Teklad: I don't mind removing it.  There's a couple of oddities related to not having a minimum size and working with tiny tiny windows.... but its really an edge case and I'm probably being paranoid.

Teklad: Actually... this one needs to stay.   The minimum size of the entire window is limited to the text in the widget there.  Assuming they start off with a small window and open a large document the window would automatically resize to fit the obviously larger text.

I'd rather leave it up to the user on cutting off informational text themselves.

Teklad:
@danieleds if you see fit you should be able to remove clickablelabel.cpp/h from the repository after this PR is done.  I'm pretty sure the only place we actually used that was the status bar.

JuBan1:
It's mergin' time!

danieleds:
Yeah sorry I'm quite busy again but I'm not dead (for now).
As @Teklad suggested, let's remember to remove clickablelabel.cpp/h if it's not used anymore.

Also:
the status bar had quite a few modifications for showing up properly on macos... I didn't test it yet. I'll let you know when I can.

danieleds:
In addition, this is not actually a refactor. A more appropriate tag should be "feat:".

It looks stupid but it is very important for when there is a bug and we have to find out where it was introduced. "Refactor" commits should only contain code changes that *do not* affect functionality (variable changes, method extractions, and stuff like that).

Not a big deal for now, but saying so that you know it. I'll try to make a quick recap in the wiki for the correct prefix to use in each situation.

JuBan1:
Now make it a bright red flashing icon above the Merge button!

PR #692, opened 2018-05-21
refac: Add clang-formatted files to commit before completion.

Teklad:
Changes a bit of the output of the git hook and automatically adds the changed files to the commit tree prior to commit.  This will prevent having to commit twice when this hook is being used.  It also only adds files that already have staged changes so we don't accidentally pull anything extra in.

### DIFF
--- a/build-tools/githooks/pre-commit
+++ b/build-tools/githooks/pre-commit
@@ -1,10 +1,13 @@
 #!/bin/sh
 
-# Run the format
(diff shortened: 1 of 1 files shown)


danieleds:
Very often I use the switch "-p" to git add when I only want to partially stage a file. Does this behave well in such case, or is it going to commit the whole file anyway?

PR #696, opened 2018-05-31
update chinese translation

wevsty:
I updated the Simplified Chinese translation file. Make the translation file correspond to the current version.
In addition, I observed that the display of the about interface may not be fully displayed, so I modified the interface size to ensure full display.


JuBan1:
Looks good, thank you!

danieleds:
@JuBan1 remember the "l10n:" prefix when merging language stuff

Issue #702, opened 2018-06-08
Crash on Mac after launching


liangliangpan:
**I'm submitting a bug report** 

**Current behavior**
Double click the notepadqq app on Mac,  wait 20 seconds, the appliction crash.

**Expected behavior**
do not crash after launching the application.

**Minimal reproduction of the problem with instructions**
Double click the notepadqq app on Mac,  wait 20 seconds, the appliction crash.

**Please tell us about your environment:**
Source code version:    20180608, master branch( V1.4.8) 
OS:   Mac OSX 10.13.4
QT:   5.11 (installed by homebrew)
Compiler:  Apple LLVM version 9.1.0 (clang-902.0.39.2)
Machine: macbook pro 15.4 with retina display

PR #703, opened 2018-06-08
Fix crash on Mac after launching #702

liangliangpan:
This fix has low risk, just fix the crash bug on Mac( which I have already open a bug #702).
Now it works well on Mac.
However I see some other crash related to preference dialog, which I may submit a new fix later.

### DIFF
--- a/src/ui/EditorNS/editor.cpp
+++ b/src/ui/EditorNS/editor.cpp
@@ -239,7 +239,7 @@ namespace EditorNS
     bool Editor::isClean()
     {
         QVariant data(0);
-        return asyncSendMessageWithResult("C_FUN_IS_CLEAN",data).get().toBool();
+        return asyncSendMessageWithResult("C_FUN_IS_CLEAN", data).get().toBool();
     }
 
     QPromise<void> Editor::markClean()
--- a/src/ui/EditorNS/editor.cpp
+++ b/src/ui/EditorNS/editor.cpp
@@ -238,7 +238,7 @@ namespace EditorNS
 
     bool Editor::isClean()
     {
-        QVariant data(0); // avoid crash on Mac OS X
+        QVariant data(0); // avoid crash on Mac OS X, see issue #702
         return asyncSendMessageWithResult("C_FUN_IS_CLEAN", data).get().toBool();
     }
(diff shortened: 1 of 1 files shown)


liangliangpan:
@danieleds could you please help to review the patch?

liangliangpan:
@danieleds could you please help to review the patch?

liangliangpan:
@danieleds could you please help to review the patch? The build system says it checked failed. I see the logs, the source diff in travis logs is different from what I commit. It is a little bit weird,  like something wrong in travis scripts?

Teklad:
Looks like a formatting issue.  We check against clang-format for cpp files so we can try to keep the same formatting across all the new code (this is kinda new).

Anyways, here's the build log:
```
0.18s$ .travis/script.sh
Running clang-format against branch master, with hash 2434e38e84ff1082c9183707481dbb2f22bf0674
clang-format failed:
diff --git a/src/ui/EditorNS/editor.cpp b/src/ui/EditorNS/editor.cpp
index b84af80..78c2f0a 100644
--- a/src/ui/EditorNS/editor.cpp
+++ b/src/ui/EditorNS/editor.cpp
@@ -239,7 +239,7 @@ namespace EditorNS
     bool Editor::isClean()
     {
         QVariant data(0);
-        return asyncSendMessageWithResult("C_FUN_IS_CLEAN",data).get().toBool();
+        return asyncSendMessageWithResult("C_FUN_IS_CLEAN", data).get().toBool();
     }
```

Looks like your spacing is just a bit off.  If you can correct that, it should allow the build to pass.

danieleds:
Thanks. Could you also add a comment specifying that this fixes a crash on macOS? So that we aren't tempted to refactor it in the future.

JuBan1:
Looking at this, it's probably not just a fix for macOS. We pass a temporary into the `asyncSendMessageWithResult` function call that takes its argument as a references (and internally passes it along). Somewhere along that path it probably keeps a reference to that QVariant that lives past the end of the async call. That's undefined behavior.

danieleds:
@JuBan1 nice catch. We'd have to see if we do the same for other calls.

PR #712, opened 2018-06-13
Removed Icon line from appdata.xml

jdeluyck:
After adding the notepadqq.appdata.xml to my Fedora SPEC file (and validating it), I came across the fact that the Icon line is not allowed.

appdata-util validate-relax gave an error that the line was not allowed.

### DIFF
--- a/support_files/notepadqq.appdata.xml
+++ b/support_files/notepadqq.appdata.xml
@@ -4,7 +4,6 @@
   <metadata_license>MIT</metadata_license>
   <project_license>GPL-3.0</project_license>
   <name>Notepadqq</name>
-  <icon width="96" height="96">https://user-images.githubusercontent.com/4319621/36906314-e3f99680-1e35-11e8-90fd-f959c9641f36.png</icon>
   <summary>An advanced text editor</summary>
   <description>
     <p>


JuBan1:
That's a bit surprising considering it's [part of the official AppStream spec](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html).

On the other hand, I can't find an AppData file on my system that uses `<icon>` - other than Nqq, that is. But that begs the question: how to software-centers that use the AppStream metainfo retrieve the application's icon?

jdeluyck:
Reading through the spec for the type desktop (which perhaps should be updated to type="desktop-application"), I don't find any reference to Icons, but it is stating:
> All tags defined in the generic component specification are valid for desktop-application components as well. 
so this _should_ be valid. I've come across more github issues with appdata.xml files and icons.

I'm going to take it up upstream.

jdeluyck:
https://github.com/hughsie/appstream-glib/issues/243

JuBan1:
Seems like the best idea is to remove the icon-tag. We've got an icon set in .desktop and other projects (like [keepassxc](https://github.com/keepassxreboot/keepassxc/issues/1131)) have gone through the same change.

Will merge. Thanks!

Issue #393, opened 2017-03-13
Begin/End select Menu


ellicefix:
There is a 'Begin/End select Menu' in the notepad++, it is very useful,Would you consider adding this feature to the new edition?

JuBan1:
I see why this would be useful- our current GoTo Line implementation moves cursor selection as well.

It's something to keep in mind although it likely won't come anytime soon since it's fairly low priority.

maravento:
After a year, I ask the same question as ellicefix. When will they implement 'Begin/End select Menu'... or or it is still "low priority"

Teklad:
We're currently revamping the editor component to (hopefully) be a native widget.  This feature request got buried somewhere in the mounds of work we've managed to make for ourselves hacking together different parts of NQQ.  If all goes well we should be able to implement this feature soon-ish.  I can't give an exact ETA on it but basically whenever the editor component gets a very needed upgrade.

JuBan1:
To add to that, Notepadqq is being worked on by three semi-active devs and we've about 100 outstanding issues of varying importance and difficulty. We have to triage because there's just too much to do.

PR #727, opened 2018-07-15
Added Begin/End Selection Functionality

Daniel-McCarthy:
Greetings,

I have implemented the Notepad++ Begin/End Select feature that has been requested ( #393 ).
This should allow users to mark a beginning cursor location for a selection, then use go-to-line and set an end cursor location for the selection without losing the initial cursor location.

I would be happy to modify it's implementation or make other changes to allow it to match the project structure better. Any feedback would be great.

Dan

### DIFF
--- a/src/ui/include/mainwindow.h
+++ b/src/ui/include/mainwindow.h
@@ -105,6 +105,7 @@ private slots:
     void on_actionCopy_triggered();
     void on_actionPaste_triggered();
     void on_actionCut_triggered();
+    void on_actionBegin_End_Select_triggered();
     void on_currentEditorChanged(EditorTabWidget* tabWidget, int tab);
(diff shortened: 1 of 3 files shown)


danieleds:
Hi Daniel, thanks for the PR.

First of all I suggest you fix the small formatting issues that are causing the tests to fail: https://travis-ci.com/notepadqq/notepadqq/jobs/134552397
(I think you're using tabs instead of spaces)

Other than that, I didn't try it yet but your code looks fine to be merged

Daniel-McCarthy:
Thank you for the great software. I am just happy to be able to contribute something back and would be glad try some more contributions in the future.

I have re-submitted and it seems the changes have passed the formatting and compilation tests.

PR #731, opened 2018-07-21
fix: Use custom event filter for QWebEngineView.

JuBan1:
QWebEngineView doesn't propagate a number of events, inclusing FocusIn and KeyPress which we need. There's something of a hack where an event filter installed in the right child object can be used to sniff out those events.

It fixes switching between INS and OVR, and proper tab change when clicking on editors.

Would be nice if someone could test it too since it's kind of a big change in event handling. I don't see any ill effects though.


danieleds:
@JuBan1 sorry for the delay. Looks fine to me, I just want to test how it behaves on macOS just to be sure. I'll let you know asap

Issue #469, opened 2017-08-14
Add "toggle last viewed tab" feature


s101d1:
Feature request here, I would like to have functionality to toggle last viewed tab, which in Windows  people usually would assign it with Ctrl + Tab key shortcut.

Daniel-McCarthy:
I would be interested in implementing this feature.

I can submit a pull request for this feature, though I feel it may be worth discussing first. 

(1. Notepad++ has a similar feature to this, but it's functionality allows it to also select other tabs than the most recent one. Would a short-cut that purely switches between the current and most recent tab be sufficient?

(2. On the version I have tested (v1.4.8), Ctrl + Tab is set to switch to the adjacent tab to the right. There seems to already be short cuts for switching to adjacent left/right tabs with Ctrl + Page Up/Page Down. But since it is already taken, the toggle feature could potentially use Ctrl + Q. I think this would make practical sense since it is immediately next to Tab.

I would happy to add this if it's something that fits the vision for the project and the details of it are agreed upon.

danieleds:
<kbd>(Shift +) Ctrl + Tab</kbd> is commonly used to switch to adjacent tabs also in web browsers, so I wouldn't change that.

<kbd>Ctrl + Q</kbd> sounds totally fine to me for switching to the last viewed tab. What is your suggestion with respect to the behaviour when multiple views are in use?

JuBan1:
Isn't CTRL+Q often used as a shorthand for quitting? We're not using it as such but there's an argument to be made for swimming with the stream.

Daniel-McCarthy:
Great points. It doesn't seem that Ctrl + T is being used and would potentially be more idiomatic.
I would be happy to set it to Ctrl + T if there are no other problems with that.

As for the functionality when there are multiple views, this is a good question. The functionality in Notepad++ appears to be a combination of the adjacent tab and most recent tab functionality. The Ctrl+Tab shortcut switches to the most recent tab. Ctrl + (Tab(2x)) will instead switch to the adjacent tab.

If the most recent tab is not tracked between multiple views (therefore reset each time you switch views) then it may seem inconsistent and less useful. This feature seems useful to me, though may be a bit more complicated when considering multiple views and windows.

JuBan1:
Fine with me. Shortcuts are easy enough to change in case we get a better idea.

> This feature seems useful to me, though may be a bit more complicated when considering multiple views and windows.

[...]

PR #733, opened 2018-07-22
Implemented Toggling to Most Recent Tab

Daniel-McCarthy:
I have added a shortcut (Ctrl + T) to toggle to the most recent tab the user was on prior to the current. It allows toggling back and forth, and for each view in a window. It implements the open Issue #469.  This feature depends on the functionality in the current pull request #731 in order to function as expected.

In my tests with that pull request's branch this feature works one/two views and has no problems with multiple windows.

I tried to keep it simple to avoid adding too many functions and changing too many. If any changes are requested I am more than happy to update this and will make sure to update it for TravisCI.

### DIFF
--- a/src/ui/editortabwidget.cpp
+++ b/src/ui/editortabwidget.cpp
@@ -21,6 +21,8 @@ EditorTabWidget::EditorTabWidget(QWidget *parent) :
     this->setTabBarHidden(false);
     this->setTabBarHighlight(false);
 
+    connect(this, &EditorTabWidget::currentChanged, this, &EditorTabWidget::on_currentTabChanged);
+
 #ifdef Q_OS_MACX
     this->tabBar()->setExpanding(true);
     this->setUsesScrollButtons(true);
@@ -354,4 +356,18 @@ void EditorTabWidget::on_fileNameChanged(const QUrl & /*oldFileName*/, const QUr
 
     setTabText(index, fileName);
     setTabToolTip(index, fullFileName);
+}
+
+int EditorTabWidget::formerTabIndex()
+{
+    return m_formerTabIndex;
+}
+
+void EditorTabWidget::on_currentTabChanged(int index)
+{
+    // Store current index to become former index on next tab change.
+    if (m_mostRecentTabIndex != index) {
+        m_formerTabIndex = m_mostRecentTabIndex;
+        m_mostRecentTabIndex = index;
+    }
 }
--- a/src/ui/include/editortabwidget.h
+++ b/src/ui/include/editortabwidget.h
@@ -66,12 +66,17 @@ public:
     void setTabText(Editor* editor, const QString& text);
     void setTabText(int index, const QString& text);
 
+    int formerTabIndex();
+
 private:
 
     // Smart pointers to the editors within this TabWidget
     QHash<Editor*, QSharedPointer<Editor>> m_editorPointers;
 
     qreal m_zoomFactor = 1;
+
+    int m_formerTabIndex = 0;
+    int m_mostRecentTabIndex = 0;
 
     void setTabBarHidden(bool yes);
     void setTabBarHighlight(bool yes);
@@ -91,6 +96,7 @@ private slots:
     void on_cleanChanged(bool isClean); 
     void on_editorMouseWheel(QWheelEvent *ev);
     void on_fileNameChanged(const QUrl &, const QUrl &newFileName);
+    void on_currentTabChanged(int index);
 signals:
     void gotFocus();
     void editorAdded(int index);
--- a/src/ui/include/mainwindow.h
+++ b/src/ui/include/mainwindow.h
@@ -197,6 +197,7 @@ private slots:
     void on_actionShow_Menubar_toggled(bool arg1);
     void on_actionShow_Toolbar_toggled(bool arg1);
     void on_actionMath_Rendering_toggled(bool on);
+    void on_actionToggle_To_Former_Tab_triggered();
 
 private:
     static QList<MainWindow*> m_instances;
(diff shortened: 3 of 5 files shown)


Daniel-McCarthy:
Certainly. I had already taken a look at that in case it was requested. I will try to get that implemented and update the pull request. As for asking in case I need help, thank you, I do greatly appreciate that.

Teklad:
Its nice seeing new people making PRs for this project.  It's been a while since I've seen any new faces around here. :)

danieleds:
I'm not super-happy with spreading this feature into different files (the fields in editortabwidget.h, part of the logic in topeditorcontainer.cpp).

Would it be possible to implement this without touching TopEditorContainer? Maybe with a method inside EditorTabWidget that returns the "formerTabIndex" so that in MainWindow we can just do

    curTabWidget->setCurrentIndex(curTabWidget->previousTabIndex());

without bothering about "bookkeeping" details.

In addition, let's not forget to test this for edge cases (e.g. "formerTabIndex" doesn't exist anymore).

Any thoughts?

Daniel-McCarthy:
I'm happy to make more modifications to my implementation. As for moving changes to the TopEditorContainer to the EditorTabWidget, I'm not sure how it would be best to do this.

I am not aware of an existing way to update the formerTabIndex from the EditorTabWidget class on tab change. My understanding is TopEditorContainer uses the connect function to call it's on_CurrentTabChanged function to a function that's inside a QT class that EditorTabWidget inherits from.

I am new to both QT and the code base so I'm not sure what would be the most natural.

Thanks for your time.

danieleds:
Don't worry, it's a lot of stuff.

Basically, instead of doing that from within the event handler, you can override the protected "EditorTabWidget::currentChanged" method and insert the logic in there, similarly to what happens with the "EditorTabWidget::mouseReleaseEvent" or "EditorTabWidget::tabRemoved" methods.

Feel free to ask if you need any clarification! The codebase is quite large.

JuBan1:
To be clear, there's no `EditorTabWidget::currentChanged` function. But there is a [currentChanged](https://doc.qt.io/qt-5/qtabwidget.html#currentChanged) signal in QTabWidget which is EditorTabWidget's base class.

[In Qt, signals](https://doc.qt.io/qt-5/signalsandslots.html) are a great way to hook your own code into certain events. If you want to run a function when a certain signal is emitted, you have to connect it to said signal.

An example is in our [Keygrabber](https://github.com/notepadqq/notepadqq/blob/master/src/ui/keygrabber.cpp#L12) class. Using the `connect` function in the constructor, the KeyGrabber's ` itemChanged` function will be run whenever the base class' (QTreeWidget) `itemChanged` signal is emitted. 

So you could do the same with EditorTabWidget's `currentChanged`. One thing to note though: `currentChanged` has an `int index` argument, so the function you connect to that signal should also take an int argument. The [Qt documentation](https://doc.qt.io/qt-5/qtabwidget.html#currentChanged) is love & life in these cases.

danieleds:
My bad, I was in a hurry and I was sure there was a protected `QTabWidget::currentChanged`, but it is actually a signal.

Ignore what I said, the solution proposed by @JuBan1 is totally correct.

Daniel-McCarthy:
I have implemented it as suggested by using connect to attach a function in EditorTabWidget to the CurrentChanged signal.

There is a bit of a side effect to this implementation that I had not considered. When using the on_currentTabChanged in the TopEditorContainer class then it was possible to only set the formerTabIndex for the currentTabWidget.

When using the on_currentTabChanged function I added to the EditorTabWidget all of the tab widgets detect a tab has changed and set the formerTabIndex (even if they are not the ones that have changed tabs). 

This issue can potentially be avoided if the EditorTabWidget has any way to find out what the TopEditorContainer's current tabWidget is, but I'm not sure that it can currently. Perhaps I have made a mistake with my implementation with the connect.

Daniel-McCarthy:
Okay, so I have implemented this in the EditorTabWidget class with the connect function (thank you for the information and documentation). 

I have hit an issue that I think may be a bug. In my testing noticed that if I run a build of my new branch and proceed to move a tab from the left view to the right view, then the on_currentTabChanged function will stop being called entirely for the left view. It remains like this until I create a new tab in that view.

So the left view will work entirely normally (including my added feature) until I move a tab to the other view. Then my feature will no longer work in the left view because on_currentTabChanged is no longer being called.

I have published my current implementation to a new branch. It is not fully tested yet. It seemed to be working expectedly save for this issue I have been running into.
https://github.com/Daniel-McCarthy/notepadqq/commit/157e48edabc3904e4d4e2a8b4c7db9564080c92f
(I had placed a breakpoint in the on_currentTabChanged function to confirm it was not running after the steps mentioned above, but I am not sure if this is something just on my end or not.)

JuBan1:
That's because how you added the signal connections. `connectEditorSignals` is called every time a new Editor (=tab) is added to the widget. Likewise, `disconnectEditorSignals` is called on each tab just before it is removed. We use those two functions to connect *Editor* signals to the EditorTabWidget when the Editors are created, and to disconnect them when they're removed.

Problem is, when a tab is removed, your `disconnect(this, &EditorTabWidget::currentChanged...` is called and you won't receive that signal anymore.

The solution should be to remove the disconnect call and move the connect call into the constructor so it's called once at the start.

Daniel-McCarthy:
I haved fixed this thanks to your explanation. That makes a lot of sense. I greatly appreciate the assistance from both you and Daniel; thanks again for your time.

I have updated the pull request with a new version. The fields in the EditorTabWidget.h are now private and not modified directly outside of the on_currentTabChanged function. The formerTabIndex now has a public get method, and the TopEditorContainer class is no longer modified.

This build appears to be working equivalently to my original implementation. Each view can toggle as expected, and multiple windows with views seem to be working properly.

danieleds:
Looks good!
Thanks for your patience, merging now :)

PR #738, opened 2018-07-30
Update Polish translations

einsteinsfool:
Updated Polish translation file after running `lupdate -no-obsolete src/ui/ui.pro -ts src/translations/notepadqq_pl.ts` in forked repo as suggested [here](https://github.com/notepadqq/notepadqq/wiki/Translate-Notepadqq).

Please do not merge yet.

I didn't know what to do with phrases with `&` (in MainWindow), that can be triggered with a shortcut `ALT+<letter>`. Should I make it so the shortcuts in translation are the same as in the original text? Or should I just pick whatever I think suiting (1st/2nd letters mostly) but so the shortcuts are not duplicated?
In case of $1, what should I do if my translation doesn't have the letter necessary? E.g. for "&File" the translation is "&Plik" which doesn't have 'f'.
In case of $2, can the letters duplicate but in different submenus? So can I make one shortcut with letter 'p' in "File" menu and also one in "Edit" menu?

I'll update this PR after I know what to do.

### DIFF
--- a/src/translations/notepadqq_pl.ts
+++ b/src/translations/notepadqq_pl.ts
@@ -2,19 +2,271 @@
 <!DOCTYPE TS>
 <TS version="2.1" language="pl_PL">
 <context>
+    <name>AdvancedSearchDock</name>
+    <message>
+        <location filename="../ui/Search/advancedsearchdock.cpp" line="161"/>
+        <location filename="../ui/Search/advancedsearchdock.cpp" line="651"/>
+        <source>Advanced Search</source>
+        <translation>Wyszukiwanie zaawansowane</translation>
+    </message>
+    <message>
+        <location filename="../ui/Search/advancedsearchdock.cpp" line="166"/>
+        <source>Clear Search History</source>
+        <translation>Wyczyść historię wyszukiwania</translation>
+    </message>
+    <message>
+        <location filename="../ui/Search/advancedsearchdock.cpp" line="175"/>
+        <source>Go To Previous Result (Shift+F4)</source>
+        <translation>Przejdź do poprzedniego wyniku (Shift+F4)</translation>
+    </message>
+    <message>
+        <location filename="../ui/Search/advancedsearchdock.cpp" line="180"/>
+        <source>Go To Next Result (F4)</source>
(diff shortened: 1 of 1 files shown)


JuBan1:
> Should I make it so the shortcuts in translation are the same as in the original text?

I believe the shortcut will be changed based on where you put the '&', so you don't have to pick the same keys. Give it a try if it works, and if it does you can pick whatever shortcuts you think are best. We haven't hardcoded any of the shortcuts so it shouldn't be an issue on our side.

I think it would be best to use the common Polish shortcuts for all the basic stuff (open/save/close file, copy/paste, etc). In cases where the Polish word doesn't have the right letter you can add it in parenthesis (`Plik (&F)`).

> In case of $2, can the letters duplicate but in different submenus?

Yes, since actions in menus are only triggered when that menu is open. So there's no conflict between actions in different menus.

einsteinsfool:
Thanks for the clarification. Seems rational.
Updated the translation. I also changes `&` in some places from previous translation as there were duplicates in some submenus. It might cause problems for users who remember shortcuts but imo it was worth changing as some entries could never be reached.

From my perspective, good to merge.

src/translations/notepadqq_pl.ts
```diff
@@ -1,155 +1,423 @@
...
+    </message>
+    <message>
+        <location filename="../ui/Search/advancedsearchdock.cpp" line="175"/>
+        <source>Go To Previous Result (Shift+F4)</source>
+        <translation>Przejdź do poprzedniego wyniku (Shift+F4)</translation>
+    </message>
+    <message>
+        <location filename="../ui/Search/advancedsearchdock.cpp" line="180"/>
+        <source>Go To Next Result (F4)</source>
+        <translation>Przejdź do następnego wyniku (Shift+F4)</translation>
```
JuBan1: I think this is supposed to be `(F4)`, since the previous one is already `(Shift+F4)`

einsteinsfool:
Thanks for spotting. And sorry for missing it. Fixed.

JuBan1:
Looks good. Thanks for your contribution.

Issue #509, opened 2017-11-19
Problem in Italian localization


marcoM32:
**Description**
On my Linux Mint after updating Notepadqq the program is not fully translated, but more than one menu is displayed in English.

![schermata a 2017-11-19 20-28-13](https://user-images.githubusercontent.com/16719549/32994494-69ee26bc-cd68-11e7-806c-37728730b87d.png)

**Current behavior**
Not all labels is in the system language

**Expected behavior**
All labels is in system language because the program resource for this language is complete

```
<message>
        <location filename="../ui/mainwindow.ui" line="883"/>
        <source>Interpret as UTF-8 without BOM</source>
        <translation>Interpreta come UTF-8 sans BOM</translation>
</message>
```

**Minimal reproduction of the problem with instructions**
Systematic

**Please tell us about your environment:**
Program version: 1.2.0

danieleds:
Not sure, but it could be because of the introduction of the keyboard shortcuts (e.g. "Interpret" is now actually "&Interpret").

If that's the case, we just need to update the translation file.

PR #742, opened 2018-08-05
fix: Italian translation accelerator issues

Daniel-McCarthy:
This pull request is in response to Issue #509. As the issue reporter mentioned, a number of lines from the Italian translation are no longer appearing in the current version of Notepadqq. And as Daniel mentioned, it is due to a number of lines being modified adding ampersands that were not there previously.

-I have gone ahead and fixed any text I could find that was no longer showing due to this. There are a number of lines that are not currently translated that did not have an existing translation, and I have left those as is.

-Lastly, in regard to the use of ampersands to underline a specific letter:

(1. In cases where words were similar enough I kept the ampersand as close as possible.

(2. In cases where the translation shared the letter that had an ampersand, I included it (even if it was not idiomatic). (This may be a point of contention that I would happily change.)
For example: &Open in a New Window -> Apri in una nu&ova finestra

(3. In cases where words did not have a matching letter I included no ampersand.
For example: E&xtensions -> Estensioni

I hope this is acceptable, as I have not updated a translation before. Let me know if there are any issues with this.

italiantest

### DIFF
--- a/src/translations/notepadqq_it.ts
+++ b/src/translations/notepadqq_it.ts
@@ -1,20 +1,271 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="1.0" language="it_IT">
+<TS version="2.1" language="it_IT">
+<context>
+    <name>AdvancedSearchDock</name>
+    <message>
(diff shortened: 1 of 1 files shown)


JuBan1:
I gave it a whirl and I can see the improvements. Looks good. Thanks!

Issue #753, opened 2018-08-28
Auto-detect .tsx files as TypeScript


Daniel-McCarthy:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[X] feature request
[ ] support request => Please do not submit support request here, instead see https://github.com/notepadqq/notepadqq/blob/master/CONTRIBUTING.md#question
```

**Current behavior**
Upon opening a .tsx file the language is set to Text.

**Expected behavior**
Opening a .tsx file should set the language to TypeScript.

**Minimal reproduction of the problem with instructions**
-Create file with .tsx extension, then open it in Notepadqq.

**Please tell us about your environment:**
Notepadqq v1.4.8, QT v5.5.1

When working on projects that contain lots of .ts and .tsx files it can feel very inconsistent having .ts files auto-detect as TypeScript whereas when opening a .tsx file it opens as plain text. It would be ideal to be able to auto-detect just as .ts files are.

[Here is an example](https://github.com/desktop/desktop/tree/master/app/src/ui/add-repository) of a project that contains numerous .ts and .tsx files mixed in their directories.

I would be happy to submit a pull request accompanying this Issue.

PR #757, opened 2018-09-03
Auto-detect TSX files as TypeScript

Daniel-McCarthy:
This pull request is tied to the Issue #753 which I created. The intent is for .tsx files to be auto-detected as TypeScript. 

Working with projects that have many .ts and .tsx it can be quite inconsistent to have the .tsx files open as plain text and no syntax highlighting while the .ts files do.

This change would seem especially fitting since the "jsx" extension is already associated with JavaScript, but the "tsx" is not associated with TypeScript.

### DIFF
--- a/src/editor/Languages.json
+++ b/src/editor/Languages.json
@@ -601,7 +601,8 @@
     "mode":"javascript",
     "mime":"application/typescript",
     "fileExtensions":[
-      "ts"
+      "ts",
+	    "tsx"
     ]
   },
   "jinja2":{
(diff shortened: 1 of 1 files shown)


JuBan1:
Thanks :) Just a minor issue: could you fix the whitespace indentation to be uniform? If you check the "Files changed" tab here in Github you know what I mean.

Daniel-McCarthy:
@JuBan1 Ah, of course. I was sure I had double checked for that. I have updated the pull request now to fix the formatting.

PR #762, opened 2018-09-19
Feat: Use uchardet to improve encoding detection

### LINKED ISSUE #730: Better encoding auto detection
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
```

**Current behavior**

Unable to quickly switch language of the file.

**Expected behavior**

The best solution: add language auto detection after reload from disk and reload with another encoding. And add encoding auto detection.
The other solution: add a button to menu that auto detects the language. Add a way to add buttons for switching directly to a language to the toolbar.

The use case: I have a project where files are windows-1251 encoded, so when I open a file it has syntax highlight, but wrong encoding. I reload the file with that encoding and I loose the syntax highlight. Now I have to go to languages/P/PHP for example. Too many things to do to start working with a file.

OpenSuse 15.0, Qt 5.9.4, Notepadqq 1.4.0

Teklad:
I tested this against uchardet's multiple test files and it appears to detect each and every one of them properly.  It also happens to greatly simplify the decoding process overall.  So this should fix #730 

The problem here is with distribution.  I'm sure uchardet is probably available on most distributions.  The issue I'm having here is qmake's poor dependency tracking with anything that's... well... not QT.

I know it didn't make sense before, but if we're going to start using other libraries alongside QT (which we should, it makes things a lot easier), we should probably rethink our current build system.  I tend to prefer CMake, but I'm okay with anything that has better library dependency handling.

### DIFF
--- a/.travis/macos-script.sh
+++ b/.travis/macos-script.sh
@@ -4,6 +4,7 @@ compile()
 {
     brew install qt
     brew install uchardet
+
     export PATH="/usr/local/opt/qt/bin:$PATH"
     export PKG_CONFIG_PATH="/usr/local/opt/qt/lib/pkgconfig"
 
--- a/.travis/macos-script.sh
+++ b/.travis/macos-script.sh
@@ -3,7 +3,7 @@
 compile()
 {
     brew install qt
-    uchardet
+    brew install uchardet
     export PATH="/usr/local/opt/qt/bin:$PATH"
     export PKG_CONFIG_PATH="/usr/local/opt/qt/lib/pkgconfig"
 
--- a/.travis/macos-script.sh
+++ b/.travis/macos-script.sh
@@ -3,6 +3,7 @@
 compile()
 {
     brew install qt
+    uchardet
     export PATH="/usr/local/opt/qt/bin:$PATH"
     export PKG_CONFIG_PATH="/usr/local/opt/qt/lib/pkgconfig"
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,8 @@ RUN apt-get -qq update && apt-get --no-install-recommends -y install \
     pkg-config \
     qtbase5-dev \
     qttools5-dev-tools \
-    qtwebengine5-dev
+    qtwebengine5-dev \
+    libuchardet-dev
 
 WORKDIR /build/
 CMD bash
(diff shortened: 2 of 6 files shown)


JuBan1:
The snapfile will have to include `libuchardet-dev` too.

danieleds:
Great, I'll take a look later today

danieleds:
Also interesting: did the macOS build automagically found the lib?

danieleds:
Okay, can you try these 2 things?

1) merge the changes from master, then
2) add "uchardet" at the end of [this line here](https://github.com/Teklad/notepadqq/blob/master/.travis/macos-script.sh#L5)

I expect an error but that's good.

danieleds:
In addition, the configure script should probably be updated to include checks for this lib

Teklad:
That reminds me, the configure script isn't failing properly when certain libraries are missing.  Qt5WebSockets doesn't cause a fail on my end (I'm not sure why)... also Qt5Svg is missing from the checks.

Teklad:
Fixed the configure script... was forgetting a "1" at the end of errorExit for the pkg-config checks.  Just need to ensure this build here passes and it should be good to go.

Teklad:
Everything appears to have passed, I might need to open up an issue, however.  The configure script appears to do odd things on MacOS:

```bash
checking for make... /usr/bin/make
checking for pkg-config... /usr/local/bin/pkg-config
checking for Qt5Core library... not found!
checking for Qt5Gui library... not found!
```

danieleds:
@darealshinji made most of the configure script, and I think that missing 1 was on purpose.

I think that's because not all libs could be detected by pgk-config (as just seen on macOS for example), so it is just intended as a warning for the user. @darealshinji can you confirm?

danieleds:
Just kidding. It should detect Qt. It does on my local machine. I'm not sure what's happening.

danieleds:
Can you add libuchardet-dev to the build-packages section in snapcraft.yaml?

danieleds:
Also, the instructions in the README should be updated

danieleds:
@Teklad something strange is happening... on my Ubuntu, if I open mainwindow.cpp with notepadqq it is detected as US-ASCII instead of UTF-8. If I run uchardet on it from the command line, however, UTF-8 is correctly returned. Are you able to reproduce it?

Teklad:
I can reproduce this issue here on Arch Linux.  Give me a bit of time to investigate.  I think I know what the issue is.

Teklad:
Just created a PR to fix it.  I didn't bother to see what the uchardet command line program used before implementing this.  It should work properly with that new PR.

PR #765, opened 2018-09-21
Fix: Buffer size for uchardet was too low

Teklad:
Increases the buffer size to match what uchardet's command line function uses.  This should ensure better accuracy in detecting the encoding.


danieleds:
Perfect! Thanks

PR #771, opened 2018-09-23
Update notepadqq_it.ts

Mek101:
Updated italian translation

### DIFF
--- a/src/translations/notepadqq_it.ts
+++ b/src/translations/notepadqq_it.ts
@@ -162,47 +162,47 @@
     <message>
         <location filename="../ui/Search/advancedsearchdock.cpp" line="358"/>
         <source>Use Regular Expressions</source>
-        <translation type="unfinished"></translation>
+        <translation>Usa le Espressioni Regolari</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


JuBan1:
Thanks! Merged.

Mek101:
@JuBan1 Sorry, I know it's the wrong place, but may I ask you how to implement characters like "ù", "à" and similar in the translation files? Or perhaps, what's the name of the format their written?

Thanks

JuBan1:
@Mek101 The translations files should be able to contain them from the get-to. Or do you mean how to type them? I'm not sure if it's standard but typing ` followed by an e,i,o,u,a produces the è, ì, etc for me.

Mek101:
I mean how to type those characters between the `<translation>` tags. I saw that `'` it's instead typed as `&apos;` 

ex:
`don&apos;t` instead of `don't`

JuBan1:
You should be able to just type them normally. `'` is one of the characters that gets escaped because they could interfere with how the translation file is read.

Looking at [other translation files](https://github.com/notepadqq/notepadqq/blob/master/src/translations/notepadqq_pt.ts) they just contain accented letters in text, without anything special.

Issue #768, opened 2018-09-22
Icons are low-res


danieleds:
Our icon theme is low-res and looks very bad on high DPI screens.

On Ubuntu, we see the problem for the icons not associated with a system theme:

<img width="529" alt="screen shot 2018-09-22 at 11 21 35 2" src="https://user-images.githubusercontent.com/402652/45915719-3af0e500-be5a-11e8-96fd-9641e49759b2.png">

on macOS, all the icons have this problem:
<img width="501" alt="screen shot 2018-09-22 at 11 25 02" src="https://user-images.githubusercontent.com/402652/45915723-404e2f80-be5a-11e8-909c-cb833714f66f.png">

A new icon theme would be nice to have.

(if you're not on a high DPI screen, click to see the unscaled images)

danieleds:
Some ideas that I'm considering:

https://material.io/tools/icons/?style=outline

https://feathericons.com/?ref=producthunt

https://icomoon.io/#preview-free

https://fontawesome.com/icons?d=gallery&m=free

JuBan1:
One issue with monochrome icons is that they'll be very hard to see on dark themed desktops.

I'm surprised some desktop themes comes without save-all, document-close, or document-close-all. Those aren't exactly uncommon.

danieleds:
> One issue with monochrome icons is that they'll be very hard to see on dark themed desktops.

Yes, that's something that we should definitely keep in mind. Actually, Google offers both white and dark versions, so we can decide to switch them automatically if there isn't enough contrast.

Or, given the fact that they're monochrome, we could even invert the colors by code. Either on user's choice, or by checking the contrast.

Teklad:
Should be able to do this easily by adding rgb values together, and some pseudo-code assuming we base it on widget background color:
```cpp
const int contrast = (r + g + b) / 3;
if (contrast > 160) { DarkTheme; }
else if (contrast < 96) { LightTheme; }
else { DefaultTheme; }
```

Since all the images you chose are gray scale this is probably one of the better ways to go about it.

JuBan1:
FYI, this code works:

```
    // Load the image and create a color mask (in this case, from black)
    QIcon icon("/path/to/image.svg");
    auto pm = icon.pixmap(QSize(size,size));
    auto mask = pm.createMaskFromColor(Qt::black, Qt::MaskOutColor);
 
    // Create the new pixmap and paint using the color mask.
    QPixmap pix = QPixmap(size, size);
[...]

Issue #659, opened 2018-05-09
Missing icon for Advanced Search Dock


danieleds:
It seems that at least one of the icons in the ASD is not included within resources.qrc.
It is important to include them as otherwise we might miss them on some distros and on MacOS.

danieleds:
Fixed in 25f98a896a86a1f7832db19dfd5e658cc3fbf0c0

PR #774, opened 2018-09-25
new icon theme (work in progress)

danieleds:
I'm adding a few Google icons.

- [x] still need to add a few missing icons
- [x] still need to add light icons
  - [x] turns out, the color isn't specified anywhere within the svg. In fact, on the website, the downloaded file is the same for both light and dark icons. Maybe we're able to assign a color from within c++.
- [x] License statements:
    - https://github.com/KDE/breeze-icons
    - https://material.io/resources/icons/
- [x] Light theme: better mousedown feedback


danieleds:
Probably useful:

https://stackoverflow.com/questions/43125339/creating-qicon-from-svg-contents-in-memory

Teklad:
This will fix both #768 and #659 yes?

danieleds:
> This will fix both #768 and #659 yes?

yep

JuBan1:
I found this as a possible solution: https://www.qtcentre.org/threads/53946-Is-it-possible-to-change-color-of-a-QGraphicsSvgItem

Don't know much about how SVG items work but that seems like an option.

Teklad:
Yea it would make a lot more sense just to invert the color since we're strictly targeting light/dark.

danieleds:
Yes, that's my plan. It can be done but it won't be pretty, I guess.

Teklad:
Couldn't you convert the QIcon to a QPixmap and adjust the color map?  I'm pretty sure that would work if you index the image.  I couldn't say for sure though.

Teklad:
Have a look at: http://doc.qt.io/qt-5/qpixmap.html#createMaskFromColor

Using a mask might be a little easier than fiddling with the color map.

danieleds:
The thing is, these icons are basically small text strings like this one:

    <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
        <path d="M0 0h24v24H0z" fill="none"/>
        <path d="M18.4 10.6C16.55 8.99 14.15 8 11.5 8c-4.65 0-8.58 3.03-9.96 7.22L3.9 16c1.05-3.19 4.05-5.5 7.6-5.5 1.95 0 3.73.72 5.12 1.88L13 16h9V7l-3.6 3.6z"/>
    </svg>

we'd just need to stick a "stroke=color" in the right place. But without a costly parser, we'd need to use some sort of hacky template.

danieleds:
@JuBan1 yes, any other option seems too complicated.

I just need to find some of the missing icons and we're basically done.

scx:
I'm sure you are aware of the missing icons problem.

```
Missing icon:  "edit-clear"
Missing icon:  "go-previous"
Missing icon:  "go-next"
Missing icon:  "go-bottom"
Missing icon:  "view-refresh"
Missing icon:  "document-open-recent"
Missing icon:  "application-exit"
Missing icon:  "edit-delete"
Missing icon:  "edit-select-all"
Missing icon:  "go-next"
Missing icon:  "go-previous"
Missing icon:  "go-jump"
Missing icon:  "view-fullscreen"
Missing icon:  "system-run"
Missing icon:  "window-new"
Missing icon:  "help-about"
```

Here you will find out about a similar problem with the proposed solution:
https://github.com/manisandro/gImageReader/issues/73#issuecomment-453829372

danieleds:
It also works nice by default with realtime theme changes on MacOS:

nqqtheming

danieleds:
#659 is fixed

PR #778, opened 2018-09-29
Fix: Properly restore focus & cursor on sessions

Teklad:
Exactly what it says on the tin.  Some very simple fixes to a few oversights that I came across while working on making the session code work with promises.

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -738,5 +738,7 @@ $(document).ready(function () {
         UiDriver.sendMessage("J_EVT_GOT_FOCUS");
     });
 
+    editor.focus();
+
     UiDriver.sendMessage("J_EVT_READY", null);
 });
--- a/src/ui/EditorNS/editor.cpp
+++ b/src/ui/EditorNS/editor.cpp
@@ -187,15 +187,11 @@ namespace EditorNS
     void Editor::setFocus()
     {
         m_webView->setFocus();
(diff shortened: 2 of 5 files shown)


Teklad:
I'll run a few tests after work.  I can't quite recall why we change the javascript focus in C++ but I'm sure at some point it was some kind of WebKit work-around.  Daniel might remember exactly why but I've not heard a peep from him in almost 3 weeks now.  

At the moment there's a lot of changes that need to be made, but I know if I open more than a couple of PRs without getting them merged I'll end up with overlapping work which translates to merge conflicts.  I'm starting to get a little worried we've been abandoned.  Productivity as it stands right now is nearly at a halt. :(

JuBan1:
> Productivity as it stands right now is nearly at a halt. :(

So much for the fear of stagnation :p

So while testing I didn't find any issues. I also didn't find any difference with lines 190 and 198 (= the two Codemirror focus calls) commented out or not. 

As far as I can see, the blur command is supposed to add `CodeMirror-focused` to the HTML. That class disables the blinking cursor and changes the (line?) highlight color. The blinking cursor thing obviously does; the color changes doesn't. But that doesn't even happen in the current master. So either it's been borked for a while now or it doesn't do what I think it should do.

JuBan1:
If this PR were a housefly it would already have died of old age. 

I'll merge since it seems to work on Teklad's and my end.

PR #784, opened 2018-10-11
Add missing include.

kensington:
In certain conditions it seems to [fail to build](https://665286.bugs.gentoo.org/attachment.cgi?id=547832) with Qt 5.11.


JuBan1:
Thanks :+1:

PR #794, opened 2018-10-28
chore:  Remove LRELEASE configuration option in favor of built-in function.

JuBan1:
We did the work in `configure` but not in QMake yet. With this, one can finally build entirely without qtchooser.


JuBan1:
Updated. This is certainly a lot cleaner. Especially without LRELEASE in ./configure either. 

Though, we've now lost the ability to set it manually. It was introduced because of #206 but this shouldn't be a problem when QMake provides lrelease on its own. If you're still interested in Nqq, @darealshinji, perhaps you could give it a try?

Teklad:
Well you could still provide the --lrelease flag as an override and check if it's set before running qtPrepareTool.

JuBan1:
Could, yes. But I'd rather get rid of it entirely if there's no need anymore.

Teklad:
Agreed.  Now just need a couple of people to test to make sure it works across the board, lol.

PR #807, opened 2018-11-29
fix: Wayland snap startup failure

### LINKED ISSUE #806: Notepadqq does not start
**Describe the bug**

Notepadqq does not start.

- System: Ubuntu 18.10

- Started with: notepadqq
- Notepadqq does not start, but gives the following 

**To Reproduce**
1. Install notepadqq with: sudo snap install --classic notepadqq
2. Start Notepadqq with: notepadqq
3. See error: This application failed to start because it could not find or load the Qt platform plugin "wayland-egl".
Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, xcb.
Reinstalling the application may fix this problem.
Abgebrochen (Speicherabzug geschrieben)

**Expected behavior**
Notepadqq starts.

**Screenshots**
bildschirmfoto von 2018-11-25 17-21-20

**Environment (please complete the following information):**
 - OS: Ubuntu 18.10
 - Installation type [sudo snap install --classic notepadqq]
 - Qt version [find it from the menu "? -> About Qt..."]
 - Notepadqq version [notepadqq             1.4.8           855   stable     danieleds     -]

**Additional context**

Teklad:
@JuBan1 this should fix the snap issue with Wayland (fixes #806).  Care to get this merged?

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -12,6 +12,8 @@ base: core18
 apps:
   notepadqq: 
     command: bin/desktop-launch $SNAP/usr/local/lib/notepadqq/notepadqq-bin
+    environment:
+        DISABLE_WAYLAND: 1
     desktop: usr/local/share/applications/notepadqq.desktop
 #   plugs:  # Plugs are not needed for classic confinement
 #     - browser-support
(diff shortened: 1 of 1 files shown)


JuBan1:
What exactly would that do? Run Nqq as an X11 application through XWayland?

Teklad:
@JuBan1 more or less yea.  We don't officially support Wayland yet so this is the best option I could think of.

Teklad:
We could try without DISABLE_WAYLAND, but I read around and there appears to be a bug in snap which causes some issues without the option.

Please see [this](https://forum.snapcraft.io/t/problem-launching-qt-snaps-in-wayland/7055/5) post for details on that.  Without the environment variable it looks like QT styles get borked.

PR #808, opened 2018-12-01
Enhance AppData file

scx:
I would like to add release history to the AppData file.

### DIFF
--- a/support_files/notepadqq.appdata.xml
+++ b/support_files/notepadqq.appdata.xml
@@ -33,49 +33,5 @@
   </provides>
   <releases>
     <release version="1.4.8" date="2018-05-10"/>
-    <release version="1.4.6" date="2018-05-10"/>
-    <release version="1.4.4" date="2018-05-01"/>
-    <release version="1.4.2" date="2018-05-01"/>
-    <release version="1.4.0" date="2018-04-29"/>
-    <release version="1.3.6" date="2018-04-23"/>
-    <release version="1.3.4" date="2018-04-10"/>
-    <release version="1.3.1" date="2018-04-03"/>
-    <release version="1.3.0" date="2018-04-03"/>
-    <release version="1.2.0" date="2017-10-15"/>
-    <release version="1.0.1" date="2017-02-05"/>
-    <release version="1.0.0" date="2017-02-05"/>
-    <release version="0.53.0" date="2016-08-29"/>
-    <release version="0.52.1" date="2016-08-12"/>
-    <release version="0.52.0" date="2016-08-11"/>
-    <release version="0.51.0" date="2016-03-05"/>
-    <release version="0.50.6" date="2015-12-24"/>
-    <release version="0.50.5" date="2015-12-24"/>
-    <release version="0.50.4" date="2015-09-18"/>
-    <release version="0.50.2" date="2015-08-27"/>
-    <release version="0.50.1" date="2015-06-27"/>
-    <release version="0.50.0" date="2015-06-27"/>
-    <release version="0.46.2" date="2015-04-26"/>
-    <release version="0.46.0" date="2015-04-05"/>
-    <release version="0.45.1" date="2015-02-21"/>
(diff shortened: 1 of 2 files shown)


support_files/notepadqq.appdata.xml
```diff
@@ -28,4 +28,54 @@
   <url type="homepage">https://github.com/notepadqq/notepadqq</url>
   <url type="bugtracker">https://github.com/notepadqq/notepadqq/issues</url>
   <url type="help">https://github.com/notepadqq/notepadqq/wiki</url>
+  <provides>
+    <binary>notepadqq</binary>
+  </provides>
+  <releases>
+    <release version="1.4.8" date="2018-05-10"/>
```
TingPing: There is no value at all of keeping every release unless they have release notes.

scx: @TingPing 

> There is no value at all of keeping every release unless they have release notes. If it is just a meaningless version number then only the most recent version is shown in any UI.

It's funny that you didn't say that before asking me for this pull request.

Anyway, the fact that Gnome Software currently doesn't display all this information doesn't mean that another program can't do it. What's more, we already have a lot of applications that use the AppData file in this way - just to store these "meaningless version numbers", e.g. Signal (an encrypted communications app).
https://github.com/flathub/org.signal.Signal/blob/280ed766a87b4eba3a8d3f321af2ca1fa5f03f82/org.signal.Signal.appdata.xml#L24-L58

You even asked for it last year. Don't you remember?
https://github.com/flathub/flathub/pull/144#discussion_r149519903
`Adding <releases> would be nice.`

And please don't say that `rules are fluid`, because the last change in this file was made just three days ago.
https://github.com/flathub/org.signal.Signal/commit/280ed766a87b4eba3a8d3f321af2ca1fa5f03f82

Teklad: @scx I don't see a problem with keeping it in there as-is.  We can always go back at a later date and add in the older release notes when we find time (literally all of it is on github).  I like the idea of having proper version history for notepadqq.

TingPing: @scx 	I didn't say remove `releases`. I said not keep 100 entries when only 1 is used. Anyway it is obviously harmless.

JuBan1: I tend to agree. There's no practical use to knowing Nqq's version release dates of the past five years. Even if we had release notes. I'd say we keep only the most recent release version unless there's a good reason to keep more.

Starting with the next release we can be more diligent and add release notes as well. I also wrote that into the [release checklist](https://github.com/notepadqq/notepadqq/wiki/Release-Checklist) but we haven't had a release since then.

JuBan1: I think without release notes the information isn't useful to anyone even if they are updating a couple of versions at a time. So let's stick to only the latest release, and from now on we'll keep the AppData file updated with release notes of future versions.

Teklad: @JuBan1 it probably wouldn't hurt to add release notes for at least a few prior to the latest (when we find time, of course).  This information isn't particularly useful to me because I just read the git logs, but considering how large the release notes list is for the current version already it might be a good idea to start from 1.4.2 or something (i.e. very recent).  

P.S. Speaking of which the release notes for the current beta is stupidly huge.

Teklad: You're probably right about that.  We didn't start using the "**prefix**: desc" format until fairly recently.  We could probably write something pretty simple that uses those tags to compile a proper detailed changelog between release tags.  Having to manually maintain a changelog sounds fairly tedious.

scx: @JuBan1 

> So let's stick to only the latest release

Done.
https://github.com/notepadqq/notepadqq/pull/808/commits/9ff490ecbfb24353462b0ed8d2c9768d85426be9

JuBan1:
Thanks, merged.

Issue #812, opened 2018-12-09
No option to retry as root, can't open with gksu


Yournished:
**Describe the bug**
I want to open system files in Notepadqq. When I try to do this normally, I get an error dialog with message "Permission denied" and options Ignore, Abort and Retry.

Supposedly if `gksu` is detected Notepadqq will offer the option to retry as root, but it doesn't for me even though though I have `gksu` installed.

Running `gksu notepadqq` in a terminal brings up the enter password dialog, and when I enter my password it disappears and nothing else happens. The terminal command ends with no output and Nqq doesn't launch. `gksudo notepadqq` has the same effect. `sudo notepadqq` nags about how you're not supposed to use sudo, but `sudo notepadqq --allow-root` succeeds in running it as root, with a different graphical theme (so it really is running as root, and even using root's settings, but I want to edit system files with my own settings).

`pkexec notepadqq` gives:
```
qt.qpa.screen: QXcbConnection: Could not connect to display
Could not connect to any X display.
```

**Environment (please complete the following information):**
 - OS: Manjaro
 - Installed from https://www.archlinux.org/packages/community/x86_64/notepadqq/
 - Qt version 5.11.2
 - Notepadqq version 1.4.8

JuBan1:
What does `which gksu` return for you? That's what Nqq uses to determine whether gksu is installed. If it properly returns a path, you at the very least ought to see the "Retry as Root" button.

Yournished:
@JuBan1 I get `/usr/bin/gksu'. So it seems like there really is a bug after all. If it helps, months ago I recall this working as expected on the same machine, so I wonder if a recent update broke it.

Teklad:
pkexec isn't meant to run graphical applications by default so that last one is intended behavior.  With that said, gksu has been deprecated for some time so it's no surprise that it randomly stopped working here.  In fact, it's not even available in Ubuntu or Arch Linux repositories either outside of the AUR.

As a side note, I was able to get notepadqq to use gksu by rebuilding from the AUR if you absolutely have to have it now, but due to the fact that it's deprecated I'd recommend against it.

I think the proper solution here would be to migrate over to using pkexec and polkit, which has been fairly standard for a good while now.

Yournished:
So what's the status of this? There's apparently a fix merged over a month ago, but I still have the issue.

Yournished:
[...]

PR #813, opened 2018-12-10
fix: Don't rely on `which` return values for verifying binary existence

Teklad:
This changes around a few things where we aren't relying on the return values of a 3rd party program in order to verify whether a binary exists.

The new method works by:
* Grabbing the **PATH** environment variable
* Creating a QDir object for each directory in the envvar
* Checking against a const list of known gui-sudo programs

It also returns the full path of the binary as well, just as a safety precaution.

It should (hopefully) fix the issue in #812, but there's really no way to tell without the user giving it a try.

I don't particularly think that utility functions like this belong in docengine, but there's not really a better place to put it currently.

### DIFF
--- a/src/ui/docengine.cpp
+++ b/src/ui/docengine.cpp
@@ -624,22 +624,28 @@ void DocEngine::unmonitorDocument(const QString &fileName)
 
 QString DocEngine::getAvailableSudoProgram() const
 {
+    //NOTE: Don't rely on `which` for this information.  Some slower hard
+    //drives were exceeding the 10ms execution limit and failing to
+    //find any sudo program at all.
+    static QString sudoProgram;
 #ifdef __linux__
-    const QStringList sudoPrograms{"gksu", "kdesu", "pkexec"};
-    QString envPath = QString::fromLocal8Bit(qgetenv("PATH"));
-    if (!envPath.isEmpty()) {
-        QStringList pathList = envPath.split(':', QString::SkipEmptyParts);
-        for (const auto& path : pathList) {
-            QDir dir(path);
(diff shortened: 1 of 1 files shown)


JuBan1:
Why can't we rely on `which`? It basically does what you just coded manually:

> For each of its arguments it prints to stdout the full path of the executables that would have been executed when this argument had been entered at the shell prompt. It does this by searching for an executable or script in the directories listed in the environment variable PATH [...]

Afaik `which` is preinstalled pretty much anywhere and we're also using it in our install scripts. So any system without `which` probably wouldn't get past installation anyways.

Teklad:
It's really a matter of not having to perform extra exec() calls spawning extra processes for the simple task of finding a file.  Outside of that I suppose it's neither better or worse.

JuBan1:
I suppose that's true. And it would fix what I think is the cause of #812: the process invokation took longer than 10ms (perhaps because it sits on a HDD) and failed because of that.

But while we're at it, could you change `getSudoProgram` to cache its result instead of fetching it anew? For example with a function-local static variable or a member variable in DocEngine? And also add a short comment since it's not immediately obvious (to me :P) what it did?

Teklad:
I think that should cover it.  I try my best not to needlessly change things (though the codebase is **really** showing its age at this point).

JuBan1:
Thanks. Will merge.

(Well, we would be able to modernize the codebase more freely if Daniel could be convinced to merge a certain thing *cough cough* :)

Teklad:
Ah... speaking of Daniel he isn't gonna be back until March so it's gonna
be a long wait.

PR #821, opened 2019-01-02
chore: fixed snap packaging

Bstiler:
Enter snap/ and run snapcraft to generate the .snap package.


danieleds:
Hi, thanks for the patch.

However, this is meant to be used by Canonical's automated build servers, so the directory from which snapcraft is ran must necessarily be the root one.

For this reason I'm closing this PR, but feel free to add further comments.

PR #910, opened 2019-10-14
Use 'command -v' and $TMPDIR in configure

noloader:
'command -v' is Posix, and more robust than 'which'. It also handles environments and aliases correctly. Also see https://stackoverflow.com/q/592620/608639.

### DIFF
--- a/configure
+++ b/configure
@@ -7,13 +7,13 @@ test -n "$CXX"       || CXX="c++"
 test -n "$QTCHOOSER" || QTCHOOSER="$(command -v qtchooser 2> /dev/null)"
 test -n "$TMPDIR"    || TMPDIR="$PWD"
 
-# Use qtchooser if we have it, otherwise use which
+# Use qtchooser if we have it, otherwise use Posix command -v
 if test -n "$QTCHOOSER"; then
     test -n "$QMAKE"    || QMAKE="$QTCHOOSER -run-tool=qmake -qt=5"
     test -n "$LRELEASE" || LRELEASE="$QTCHOOSER -run-tool=lrelease -qt=5"
 else
-    test -n "$QMAKE"    || QMAKE="$(command -v qmake-qt5 2>/dev/null || which qmake 2>/dev/null)"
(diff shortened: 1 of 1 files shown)


danieleds:
Nice, thanks!
I think there are a few more "which" to convert on lines 15/16?

scx:
To be honest, `command` is not available on all UNIX-like systems.
It is especially true for non-GNU Linux distributions.

`BusyBox`:
```
$ which ls
/bin/ls
$ echo $?
0
$ command -v ls
-sh: command: not found
$ echo $?
127
```

Although it matters mainly for embedded and mobile systems, some desktop distributions may be affected as well. For example, [Alpine Linux](https://wiki.alpinelinux.org/wiki/Main_Page) is based on `musl` and `BusyBox`, but some people actually use it on their [desktops](https://wiki.alpinelinux.org/wiki/Desktops).

Unfortunately, there is no a good alternative to `which` or `command`.
Some people suggest to use `type -P` (to find the path) or `hash` (to check), but these commands are bash-specific.
```
% type -P ls
zsh: bad option: -P
```

And because `which` is an external tool, it will be functional no matter what shell is in use. We can't say the same for `command`.
In the past, `which` was not such a good option because it was not consistent across various platforms. However, today it shouldn't be a problem anymore. In my opinion, you have a better chance to meet Alpine Linux 3.10 (released on 19 June 2019, will be supported until 1 May 2021) than HP-UX 10.20 (released in 1996, support by HP ended on 30 June 2003), OpenBSD 4.1 (released on 1 May 2007, support ended on 1 May 2008), Red Hat Enterprise Linux 2.1 (released on 26 March 2002, support ended on 31 May 2009) or Debian 5.0 Lenny (released on 14 February 2009, support ended on February 2012). To be honest, I doubt that you will find a Linux system with Qt5 and QtWebEngine, that was released before 2014.

Personally, I would stay with the current code. Just keep it in mind when someone with Alpine Linux or yet another distribution for hipster reports such an issue. ;)

See also:
http://mywiki.wooledge.org/BashFAQ/081

PR #911, opened 2019-10-14
Fix typo in original for it/ja translation files

xstasi:
The typo was corrected in the original and most translations, but these two missed the train


danieleds:
Nice catch, thanks!

PR #1080, opened 2023-01-26
czech translation

Amereyeu:
new czech translatin for notepadqq


danieleds:
thank you for the PR - this will need a bit of cleaning before it's ready to be merged

Issue #1135, opened 2025-04-28
Crash after some time after open editing


donarturo11:
**Describe the bug**
Program is crashing after open file.

**To Reproduce**
1. Open file
2. Wait some time
3. Result: SEGFAULT

**Expected behavior**
Stable running

**Diagnostic on GDB**
Load compiled binary to gdb.
Before run type to GDB:
```
b editor.cpp:233
commands
silent
printf "Calling Editor::isClean()\n"
continue
end

b editor.cpp:458
commands 
silent
printf "Calling std::shared_future<QVariant> Editor::asyncSendMessageWithResult\n"
continue
end
```
gdb output:
```
Starting program: /home/arch/art/github/notepadqq/out/debug/lib/notepadqq-bin 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
Start-time benchmark started.
WARNING: Notepadqq is running in DEBUG mode.
[New Thread 0x7fffd92456c0 (LWP 27685)]
[New Thread 0x7fffd8a446c0 (LWP 27686)]
[New Thread 0x7fffd0be86c0 (LWP 27687)]
[New Thread 0x7fffc37ff6c0 (LWP 27688)]
[New Thread 0x7fffc2ffe6c0 (LWP 27689)]
Extension support is not installed.
[New Thread 0x7fffc27fd6c0 (LWP 27690)]
[New Thread 0x7fffc1ffc6c0 (LWP 27691)]
[New Thread 0x7fffc17fb6c0 (LWP 27692)]
[New Thread 0x7fffabfff6c0 (LWP 27694)]
[Detaching after fork from child process 27695]
[Detaching after fork from child process 27696]
[Detaching after fork from child process 27697]
[New Thread 0x7fffab7fe6c0 (LWP 27700)]
[New Thread 0x7fffaaffd6c0 (LWP 27701)]
[New Thread 0x7fffaa7fc6c0 (LWP 27702)]
[New Thread 0x7fffa9ffb6c0 (LWP 27703)]
[New Thread 0x7fffa97fa6c0 (LWP 27704)]
[New Thread 0x7fffa8ff96c0 (LWP 27705)]
[New Thread 0x7fff93fff6c0 (LWP 27706)]
[New Thread 0x7fff937fe6c0 (LWP 27707)]
[New Thread 0x7fff92ffd6c0 (LWP 27708)]
[New Thread 0x7fff927fc6c0 (LWP 27709)]
[...]

PR #1140, opened 2025-05-13
Fix segfault on autosave

### LINKED ISSUE #1138: SEGFAULT after opening.
**Describe the bug**
The app crashes consistently after around 17-18 seconds after opening. Even if i do nothing and do not interact with the app at all it will still crash. The only difference i have found is that if the qt information screen is open it does  not count towards the time (if the information modal is closed it will then crash after a certain amount of time)

**To Reproduce**
1. Open app
2. Click on either yes or no to restore session
3. Wait for about 18 seconds
4. Should crash

**Expected behavior**
A clear and concise description of what you expected to happen. I expected it to not crash. 

**Screenshots**

**Environment (please complete the following information):**
 - OS: Arch (Wayland)
 - Installation type pacman
 - Qt version 5.15.16
 - Notepadqq version: 2.0.0 Beta

**Additional context**

donarturo11:
Fixes: #1138 , #1135 .

### DIFF
--- a/src/ui/Sessions/sessions.cpp
+++ b/src/ui/Sessions/sessions.cpp
@@ -289,7 +289,8 @@ bool saveSession(DocEngine* docEngine, TopEditorContainer* editorContainer, QStr
 
         for (int j = 0; j < tabCount; j++) {
             auto editor = tabWidget->editor(j);
-            bool isClean = editor->isClean();
+            bool isClean = true;
+            editor->isCleanP().wait().tap([&](bool _isClean){ isClean = _isClean; });
             bool isOrphan = editor->filePath().isEmpty();
             Editor::IndentationMode indentInfo = editor->indentationMode();
 


danieleds:
thanks for the PR @donarturo11. The project is currently unmaintained but since it looks like many users are experiencing this problem, I'm more than glad to merge a fix.

Could you add some notes about why this change fixes the issue? Also, won't this cause a race condition?

donarturo11:
> thanks for the PR @donarturo11. The project is currently unmaintained but since it looks like many users are experiencing this problem, I'm more than glad to merge a fix.
> 
> Could you add some notes about why this change fixes the issue? Also, won't this cause a race condition?

Well... before fix calling editor->isClean directly called Editor::asyncSendMessageWithResult (noticed as deprecated). Segfault was traced with gdb (I described at issue #1135 ). I noticed too, that another parts of program calls this function but did not cause SEGFAULT. In some conditions like autosave returned shared_future fut and it tried to get access to unallocated pointer. Disabling autosave fixed this problem too.

I analysed too sessions.cpp, where I looked at both Sessions::saveSession and Sessions::loadSession. I tried use similarly code to sessions.cpp::468: ``` editor->isCleanP().then([=](bool isClean){ emit editor->cleanChanged(isClean); }); ```. 
Because boolean isClean needs be read from editor, I decided to modify isClean variable with lambda.
After patch I tried to write something on new tabs, then I waited more than autosave time and closed without save and after reopen and confirmed loading unsaved session I got the same state, like before closing.

danieleds:
It looks like this introduces a race condition with the `isClean` variable though. The moment it will get updated by the lambda, it might be too late and the code on line 302 might have already been executed.

can you confirm?

donarturo11:
So do You propose wait for result before executing line 302? Can we leave lambda handling variable isClean, but look for waiting loop? Maybe leave sessions.cpp without changing, but look for solution inside isClean functions in editor.cpp? And at last do You recommend to use deprecated functions basing on std::future.

donarturo11:
@danieleds I corrected and squashed.

danieleds:
Thanks @donarturo11. 

This looks good to me. I'd like someone else to loop in and confirm this fix work for them too. @theendofline?

theendofline:
Hi @donarturo11!
I built and tried it on Ubuntu 24.04 (kernel 6.11.0) on Wayland (in VirtualBox), and it works. At least the `Segmentation fault (core dumped)` error does not occur, and persistence works for the case. As well as on my host Linux Mint 22 Cinnamon on x11 (kernel 6.8.0).

donarturo11:
Hi @theendofline . I built on Ubuntu in Podman (using Dockerfile from this repo) and tried on ArchLinux. Too works with no problem. Unluckly I have no Mac to try.

danieleds (approved):

danieleds:
Thanks guys! Much appreciated.

Merging

rsaxvc:
I think something is still failing. I've got this patch, but...

```
Thread 1 "notepadqq-bin" received signal SIGSEGV, Segmentation fault.
0x00007ffff6766a10 in QString::operator=(QString const&) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
(gdb) bt
#0  0x00007ffff6766a10 in QString::operator=(QString const&) () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#1  0x000055555569e719 in Sessions::saveSession (docEngine=0x555555d37fa0, editorContainer=0x555555bebc10, sessionPath=Python Exception <class 'gdb.MemoryError'>: Cannot access memory at address 0x2ffffd358

   #2  0x00005555556c6822 in BackupService::writeBackup (wnd=0x555555b0a420)
    at Sessions/backupservice.cpp:89
#3  0x00005555556c6356 in BackupService::executeBackup () at Sessions/backupservice.cpp:53
#4  0x000055555567cfae in QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (*)()>::call(void (*&)(), void**)
    (f=@0x7fffffffd5d8: 0x5555556c5ec2 <BackupService::executeBackup()>, arg=0x7fffffffd720) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:146
#5  0x000055555567c928 in QtPrivate::FunctionPointer<void (*)()>::call<QtPrivate::List<>, void>(void (*)(), void*, void**)
    (f=0x5555556c5ec2 <BackupService::executeBackup()>, arg=0x7fffffffd720) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:209
#6  0x000055555567c280 in QtPrivate::QStaticSlotObject<void (*)(), QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*)
    (which=1, this_=0x555556315b30, r=0x55555585c7e0 <BackupService::s_autosaveTimer>, a=0x7fffffffd720, ret=0x0)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobject_impl.h:90
#7  0x00007ffff691893e in ??? () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#8  0x00007ffff691ccfe in QTimer::timeout(QTimer::QPrivateSignal) () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#9  0x00007ffff690ba7b in QObject::event(QEvent*) () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#10 0x00007ffff7763052 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#11 0x00007ffff68dde98 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#12 0x00007ffff693a261 in QTimerInfoList::activateTimers() () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#13 0x00007ffff693ab6c in ??? () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#14 0x00007fffec7043c5 in ??? () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#15 0x00007fffec7065f7 in ??? () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#16 0x00007fffec706d60 in g_main_context_iteration () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#17 0x00007ffff693aeed in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#18 0x00007ffff68dc802 in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#19 0x00007ffff68e5222 in QCoreApplication::exec() () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#20 0x000055555558e9c3 in main (argc=1, argv=0x7fffffffdfa8) at main.cpp:183
(gdb) 
```

PR #1148, opened 2025-09-07
Use bash to setup git hooks

rsaxvc:
Fixes the following errors on Debian, where /bin/sh is not bash:

    ./setup-hooks.sh: 2: Bad substitution


build-tools/setup-hooks.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/bash
```
danieleds: have you considered using this instead?

```suggestion
#!/usr/bin/env bash
```

rsaxvc: I would be fine either way. If there's a preference - do it.

rsaxvc:
Do you prefer squashed? If not, this is ready.

PR #1150, opened 2025-09-07
Fix PRNG deprecation warnings

rsaxvc:
Move to the QT5.10+ way


rsaxvc:
Oops, just saw that qt5.6 is supported in the readme.

PR #1071, opened 2022-12-09
Qt6 Support

danieleds:
Initial support for Qt6 - builds on Linux and Apple Silicon.

Any help testing this would be welcome :)

---

## Notes

- In `configure`, qtchooser has been removed (it's not supported on Qt6).
- Started using GitHub Workflows instead of Travis (only build checks are supported for now, still missing auto deployments)

PR #1162, opened 2026-04-26
Qt6 Support, this time with visible close buttons in the advanced search

Danaozhong:
I've had a look at the [existing  Qt6 support PR](https://github.com/notepadqq/notepadqq/pull/1071), which seemed like 95% completed. I resolved the few remaining TODOs:
- fixed the missing close button on advanced search.
- implemented proper window name comparison.
- small performance improvement in the search by using `QStringView` instead of `QString`,

Tested on Ubuntu 24:
- Build compiles.
- Notepadqq opens.
- Text search tested across multiple documents.


src/ui/mainwindow.cpp
```diff
@@ -1668,7 +1668,7 @@ void MainWindow::on_editorMouseWheel(EditorTabWidget *tabWidget, int tab, QWheel
 {
     if (QApplication::keyboardModifiers() & Qt::ControlModifier) {
         qreal curZoom = tabWidget->editor(tab)->zoomFactor();
-        qreal diff = ev->delta() / 120;
+        qreal diff = ev->pixelDelta().y() / 120;
```
danieleds, line 1681: might `angleDelta` be more appropriate here?

danieleds:
@Danaozhong thanks! Would you be ok for me to merge this into the Qt6 branch first?

skjerns:
great effort! ps: the build instructions still mention qt6 build tools - I guess that needs to be adapted in the final version as well :)

PR #1165, opened 2026-06-17
added some null pointer checks to avoid crashes

MojtabaKD:
I wanted to start adding some line operation features, but engaged in solving the annoying crashes. It seems that these simple null pointer checks avoids the crashes. But maybe to completely solve the issue, we need to fix timer and backup service objects life cycles in the future.


danieleds (approved):
Thanks!

PR #1166, opened 2026-06-18
adding the transpose-line feature

MojtabaKD:
Notepad++ has a line-operation feature named "Transpose Line" which transposes the current line with the above one, I found the name of this feature from it's source code because at the time of this writing, it is not mentioned in the menus. I put this feature in the line operation menu with the shortcut key ctrl+shift+T, the shortcut for it in the notepad++ is ctrl+t but that shortcut in notepadqq is already taken up for toggling the tabs... 

I think later we should organize and group the line operations and commands of it in the source code by some standard, because right now it may looks a little unorganized 🤔.

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -529,6 +529,22 @@ UiDriver.registerEventHandler("C_CMD_MOVE_LINE_DOWN", function(msg, data, prevRe
     
 });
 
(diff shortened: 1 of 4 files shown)


danieleds (approved):
ty!

PR #1169, opened 2026-06-20
feature: a simple column editor to insert text

MojtabaKD:
Adding a simple initial Column Editor feature. I forced to change the Encoding menu alt key from C to E, because it seems that Alt+C is better suited for this feature.

### DIFF
--- a/src/editor/app.js
+++ b/src/editor/app.js
@@ -10,6 +10,27 @@ UiDriver.registerEventHandler("C_FUN_GET_VALUE", function(msg, data, prevReturn)
     return editor.getValue("\n");
 });
 
+UiDriver.registerEventHandler("C_CMD_INS_TXT_COL", function(msg, data, prevReturn) {
+
+    var doc = editor.getValue();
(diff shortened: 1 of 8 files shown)


danieleds:
Thanks! This feature sounds like it has a possible overlap with the "multicursor" feature. Is there any advantage that I may not be aware of, with building it this way?

MojtabaKD:
@danieleds If you want to add something at a specific position of lines in a file with for example hundreds of lines, it will be very helpful. I think multicursor is just for few lines, because you have to use mouse for it, if I'm right.

MojtabaKD:
I used column editor in notepad++ a lot, for example, I want to add a namespace in front of hundreds of titles in a mediawiki based website, or something like that..

MojtabaKD:
As you said and showed me, the multicursor can do the column editing perfectly. I didn't know that. So it will be better to keep notepadqq simple and let multicursor do that. 

i will close this pull request. 

thanx.

PR #1170, opened 2026-06-21
removing QRandomGenerator::global() seeding to avoid crash for Qt6

MojtabaKD:
QRandomGenerator::global() seeding caused a runtime crash for Qt6. For more info about why this kind of global seeding is deprecated and should not be used for newer Qts read this:

https://doc.qt.io/qt-6/qrandomgenerator.html


danieleds:
Great catch!

PR #1171, opened 2026-06-21
Fixing a stack-use-after-return crash in BackupService::executeBackup()

MojtabaKD:
Apparently, to avoid stack-use-after-return crash in BackupService::executeBackup() we need to change the order of  wait() and tap().


danieleds:
Nice catch.

There is a similar pattern in sessions.cpp:

```
editor->isCleanP().wait().tap([&](bool _isClean){ isClean = _isClean; });
```

should we fix it there as well?

MojtabaKD:
@danieleds 🙏✌️, it is highly probable that one needs to be corrected too. But if you want to make sure, it takes some time to test and study it...

danieleds:
makes sense! I will merge this one for now then

PR #1172, opened 2026-06-22
fixing dock button in Adv. Src. after Qt upg.

MojtabaKD:
After Qt upgrade, that _q_toggleTopLevel should not be used that way. I upgraded that line with a safer code.


danieleds (approved):
LGTM

PR #1188, opened 2026-07-21
Add second popular build system CMake

GermanAizek:
Added alternative build system

https://en.wikipedia.org/wiki/CMake


GermanAizek:
@danieleds You have already pull changes from Jan 10 2025 my fork with addition Cmake to notepadqq, thank you very much.

